PR #23, opened 2020-11-11
better semantics?

JohnRDOrazio:
I believe direct children of `ul` elements should only be `li` elements? Fixing this by wrapping all `div`s and `a`nchor tags in `li` tags doesn't seem to affect the bootstrap theming, I think it's more semantically correct?


mftruso (approved):
looks good! doesn't seem to change the appearance

PR #45, opened 2021-11-26
prepare v3.0

JohnRDOrazio:


JohnRDOrazio:
We have more solid data now that I have a copy of the 2008 Calendar from the Tertia Editio Typica emendata.

However it would also be nice to rewrite the main script as a Class, similarly to the work done in the BibleGet project.

Perhaps I should do this before publishing v3.0 ...

JohnRDOrazio:
> However it would also be nice to rewrite the main script as a Class, similarly to the work done in the BibleGet project.
> 
> Perhaps I should do this before publishing v3.0 ...

I'm going to merge this PR in the meantime. Since rewriting from the ground up is quite a task, it'll be better to have clean code merged in the meantime.

PR #225, opened 2024-09-05
implement resource paths

JohnRDOrazio:
* package the API source as a composer library for autoload functionality
* move the endpoints from PHP scripts to resource paths, and create a router
* add `YAML` as a response media type
* use JSON resource files for the Tests server in place of PHP Unit defs (remove dependency on PHP Unit)
* fix bug where liturgical commons were not getting localized correctly
* fix XML response, implement XML and ICS validation
* update the OpenAPI schema
* add timing to response output and to response headers


JohnRDOrazio:
CodeFactor score is high enough that I'm going to pull now. There might still be some things to iron out but we can sort those out in the dev branch. I have started refactoring the frontend code and the unit test code to match, but those will probably require some work also.

Issue #266, opened 2024-11-25
Canadian National Calendar


chrissherren:
Hello from Canada!
Thanks for starting this project!

I'd like to help by providing the Canadian national calendar. I've attached a json in English. I did not include the common or the readings yet but I will update it later to include those properties. I was focusing on getting the feasts to at least show up in the calendar on the correct days first.

[CANADA.json](https://github.com/user-attachments/files/17903841/CANADA.json)

Of course, in Canada we also have a French missal. It has all the same feasts on the same days so it is just a matter of including the translation. I have the French missal at my office so I can do that. Would I use the json file: nations/Americas/fr_CA.json to create this? I assume each object just needs to have a tag matching it to the Canadian English calendar and the name property with the French name for the fest. Is that correct? Or do I need to include the action property in metadata for each object in fr_CA.json ?

For reference, the Canadian Conference of Catholic Bishops have posted our calendar online in both English and French. The Links below will take you directly to the PDFs.

English: (https://www.cccb.ca/wp-content/uploads/2017/09/nationallitcalender2016_01_27.pdf)

French: (https://www.cecc.ca/wp-content/uploads/2017/11/cal_lit_approuve2016_01_27-3.pdf)

JohnRDOrazio:
Thanks a lot for this contribution! I'll try to get it integrated as soon as I can. The format has changed slightly in a newer version of the API that I've been working on, thanks to contributions from some fellow Catholic developers from the Open Source Catholic network, but I can take care of updating the format (basically all of the JSON keys are now lowercase and snake_case, to make it easier for developers).
I haven't yet implemented multiple languages for National Calendars, but I foresaw that this would come up, and I have created an issue for this #150 . So with this contribution, I guess it will be the perfect opportunity to get this worked out in the upcoming release of v4 of the API!

JohnRDOrazio:
The wider region celebration of the Immaculate Conception will be taken care of by the `wider_region` data that is already defined for the Americas, so it can be removed from this JSON file, and any translations taken care of in weblate (I'll have to create a translation component for this)

JohnRDOrazio:
If I understand correctly, comparing to the provided national calendars, Saints Joachim and Anne are celebrated as a Feast only in the dioceses of the Province of Quebec, and not all throughout Canada, so I'll put that aside for now and we can define it in a local diocesan calendar.
```json
        {
            "festivity": {
                "month": 7,
                "day": 26,
[...]

PR #267, opened 2024-11-26
Start implementing multiple language support for national calendars

### LINKED ISSUE #150: Multiple language support for national calendars
Issue #112 was taken care of by PRs #142 and #147 , and we now have full support for the different languages in a wider region.

However we may need to do further development for multiple languages within a single country.

For example in the United States, both English and Spanish are supported. So we should be able to define data for the United States both as "English USA" and "Spanish USA".

JohnRDOrazio:
Fixes #150

# Summary of changes
* move all calendar resources to a `/calendars` subfolder
* separate out diocesan calendars into their own subfolder `/dioceses`
* create `/i18n` subfolders for all national calendars
* reorder the `/wider_region` folder structure to more closely resemble `/nations`

TODO: update all code references to any of these folders (perhaps we should finally create globals, so we don't have to update scattered references, for example create a class with constants?)


JohnRDOrazio (commented):
not yet ready to merge until all of the code references have been updated.

Also, we should consider that diocesan calendars will most likely also need translation files, where there is more than one language withiin the same diocese. For example, French and English in Quebec. This might start to get a bit unwieldly when it comes to the nested folders, but I don't see how else we can do it?)

JohnRDOrazio:
This PR will also implement #266 as proof of concept.

JohnRDOrazio:
I'd say that we're at a point now that we can start merging to development, and pass to further testing on the development branch using the UnitTest interface (which might require a couple touch ups too, maybe?)

Issue #273, opened 2024-12-05
[Bug]: weekdays between Epiphany and Baptism of the Lord are not calculated correctly


JohnRDOrazio:
### What happened?

We are not calculating correctly the weekdays between Epiphany and Baptism of the Lord, when Epiphany is a Sunday AND falls on Jan 7 or Jan 8. Since we are calculating these weekdays _until the next Sunday_, it doesn't work in this case because Baptism of the Lord is NOT the next Sunday, it's the next Monday (= the day after in 2024).

Only way to fix this is to move the calculation of the weekdays between Epiphany and Baptism of the Lord to after we have calculated Baptism of the Lord.

Currently, we are calculating these weekdays in `Calendar::calculateEpiphanyJan6()` and in `Calendar::calculateEpiphanySunday()`, which is too early because we haven't yet calculated the Feasts of the Lord. So we're going to have to move this calculation to a proper place after the Feasts of the Lord are calculated. However, this merits an issue of it's own, so I'll close the current issue as fixed and open a new one.

### Version

v1

JohnRDOrazio:
check if the memorial of Saint Basil on Jan 2nd (or any other obligatory memorial that should take place during the Christmas weekdays before and after Epiphany) should suppress the Christmas weekday before Epiphany... (see for example Jan 2nd 2023)

PR #274, opened 2024-12-06
move calculation of christmas weekdays

JohnRDOrazio:
weekdays of Christmas before and after Epiphany need to be calculated after Baptism of the Lord


JohnRDOrazio:
fixes #273

JohnRDOrazio:
check if the memorial of Saint Basil on Jan 2nd should suppress the Christmas weekday before Epiphany...

Issue #288, opened 2025-01-25
[Bug]: ICS output not working since v4.0 release


JohnRDOrazio:
### What happened?

ICS output has not been re-aligned since the release of v4

### Version

dev (Edge)

### Relevant error messages

```shell

```

PR #289, opened 2025-01-25
fix iCalendar output

JohnRDOrazio:
iCalendar output hadn't been realigned since the v4.0 release. Current patch fixes that.


JohnRDOrazio:
fixes #288

PR #316, opened 2025-06-23
changed translation for latin day of week

nassaubv:
# Issue 310

### DIFF
--- a/src/LatinUtils.php
+++ b/src/LatinUtils.php
@@ -83,13 +83,13 @@ class LatinUtils
     ];
 
     public const LATIN_DAYOFTHEWEEK = [
-        "Feria I",     //0=Sunday
+        "Dominica",     //0=Sunday
         "Feria II",    //1=Monday
         "Feria III",   //2=Tuesday
         "Feria IV",    //3=Wednesday
         "Feria V",     //4=Thursday
         "Feria VI",    //5=Friday
-        "Feria VII"    //6=Saturday
+        "Sabbato"    //6=Saturday
     ];
 
     public const LATIN_MONTHS = [


nassaubv:
Hello, I will take some time to correct it later. This was the first pull request I ever made, that is why it might not look nice.

JohnRDOrazio:
ok I re-added your fix to a new PR #317 to fix the history. You still show as author of the fix, I'll just pull it from PR #317 and close this one. Thanks!

mkasberg:
> I suppose I could add a few more tools to automatically tidy up the code, perhaps @mkasberg has more experience with implementing such tools?

[PHP-CS-Fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer) is probably what you want. Laravel uses [Pint](https://laravel.com/docs/12.x/pint), built on top of php-cs-fixer, but you probably just want php-cs-fixer since you're not using Laravel. You'd configure the style rules you want (or just use the defaults), and run the `fix` command to bring the codebase inline with your style rules. You can then also run the `check` command alongside tests as part of PR validation.

JohnRDOrazio:
I have already implemented a `phpcs.xml` with the ruleset, and the `squizlabs/php_codesniffer` package in `composer.json` development dependencies. VSCode picks up automatically on the ruleset with the [phpcs extension](https://marketplace.visualstudio.com/items?itemName=shevaua.phpcs), however there is no trigger to auto-fix on push or PR. I believe you have some experience in implementing such triggers with husky? I would be more then happy if you could create a PR with a husky trigger that will at least run `./vendor/bin/phpcs src` on push or PR. The other command `./vendor/bin/phpcbf` will actually perform fixes that can be applied automatically.

PR #334, opened 2025-08-03
calculate readings for events that have easter season readings

### LINKED ISSUE #332: calculate readings for events that have easter season readings
Some liturgical events have specific lectionary readings depending on whether the celebration falls during Easter season or not. We need to check whether the event's date falls within Easter season, and associate the appropriate lectionary readings accordingly.

JohnRDOrazio:
Fixes #332

# Issue being addressed
For events that have seasonal readings, which change based on whether they fall during Easter season or not, we should produce the readings based on the season in which they fall.

# Summary of changes
 * add a method to ReadingsGeneralRoman `hasSancoraleReadings` to check whether there are readings in the Sanctorale (which includes even for national or diocesan calendar events) for the `event_key`
 * update the signature of the `ReadingsGeneralRoman::getSanctoraleReadings` method and add `ReadingsSeasonal` as a possible return type
 * add checks in `LiturgicalEventCollection::setCyclesVigilsSeasons` whether we have an instance of `ReadingsSeasonal`, and if so apply the appropriate lectionary readings based on the liturgical season


JohnRDOrazio (commented):
all tests are passing on the Unit Test interface

PR #381, opened 2025-10-08
refactor `date` to ISO 8601 formatted datetime

### LINKED ISSUE #333: refactor `date` property to `timestamp`, and add a `date` property with an ISO 8601 formatted datetime
Currently, the `liturgical_event` objects returned by the API have a `date` property which actually contains a timestamp, without milliseconds (which is the default way of dealing with Unix timestamps in PHP).

This can be a little bit confusing, people would generally expect `date` to contain an ISO 8601 formatted datetime.

So we should refactor `date` to `timestamp`, and format the timestamp **with** milliseconds (since most other web programming languages such as javascript take timestamps with millisecond resolution. This can be achieved in PHP with `(int) $date->format('Uv'); // Timestamp in milliseconds` (see [datetime.gettimestamp.php#example-2233](https://www.php.net/manual/en/datetime.gettimestamp.php#example-2233)).

We could optionally provide another more PHP friendly timestamp without millisecond resolution: what would be a good property name to avoid confusion? Are there any standards when it comes to timestamps?

We should then create a `date` property that contains an ISO 8601 formatted datetime such as `2004-02-12T00:00:00+00:00` (or perhaps with only the date part `2004-02-12`? seeing the time is not relevant...). This can be achieved in PHP with `$date->format('c')` (full ISO 8601 datetime) or `$date->format('Y-m-d');` (only year, month and day), see [datetime.format.php#refsect1-datetime.format-parameters](https://www.php.net/manual/en/datetime.format.php#refsect1-datetime.format-parameters).

Here's what ChatGPT says about 20 of the most used programming languages:

-----
<html>
<body>
<html><head></head><body><p>Here's how <strong>20 major programming languages</strong> handle <strong>timestamps</strong> by default, specifically focusing on their <strong>resolution</strong> (e.g. whether they include milliseconds or not) and the <strong>typical representation</strong> (e.g. Unix epoch, datetime object, etc.):</p>

Language | Default Timestamp Format | Millisecond Resolution | Notes
-- | -- | -- | --
C | time_t (seconds since epoch) | ❌ No | Milliseconds require gettimeofday() or clock_gettime()
C++ | std::chrono::system_clock::now() | ✅ Yes (nanoseconds) | High-resolution time since C++11, but epoch duration needs conversion
Java | System.currentTimeMillis() | ✅ Yes (milliseconds) | Nanosecond resolution via System.nanoTime() (not absolute time)
JavaScript | Date.now() or new Date().getTime() | ✅ Yes (milliseconds) | Based on epoch in ms
Python | time.time() | ✅ Yes (float seconds) | Float includes fractional seconds (microsecond resolution); datetime supports microseconds
Go | time.Now() | ✅ Yes (nanoseconds) | Full nanosecond resolution
Rust | SystemTime::now() | ✅ Yes (nanoseconds) | High precision via Duration
Ruby | Time.now.to_f | ✅ Yes (float seconds) | Microsecond resolution
PHP | time() | ❌ No (seconds only) | Milliseconds via microtime(true)
C# (.NET) | DateTime.UtcNow / DateTimeOffset.UtcNow | ✅ Yes (100-nanosecond ticks) | Resolution down to ticks (100 ns)
Swift | Date().timeIntervalSince1970 | ✅ Yes (float seconds) | Double-precision seconds since epoch
Kotlin | System.currentTimeMillis() | ✅ Yes (milliseconds) | Same as Java
R | Sys.time() | ✅ Yes (fractional seconds) | POSIXct as float
Perl | time() | ❌ No (seconds only) | Milliseconds via Time::HiRes
Lua | os.time() | ❌ No (seconds only) | Millisecond support via socket.gettime() (LuaSocket)
Haskell | getPOSIXTime (from Data.Time.Clock.POSIX) | ✅ Yes (fractional seconds) | Returns NominalDiffTime in seconds with picosecond precision
Scala | System.currentTimeMillis() | ✅ Yes (milliseconds) | Same as Java
Elixir | System.system_time(:millisecond) | ✅ Yes (configurable) | Nanosecond precision available
Erlang | os:system_time(millisecond) | ✅ Yes (configurable) | Nanosecond resolution with options
MATLAB | now or datetime('now') | ✅ Yes (fractional days) | Internally uses double precision (accurate to ~milliseconds)

<hr>
<h3>Summary</h3>
<ul>
<li>
<p><strong>Milliseconds or better supported</strong>: Most modern languages (Java, Python, JS, Go, Rust, etc.)</p>
</li>
<li>
<p><strong>Seconds only by default</strong>: PHP, Perl, Lua, C (without extra libs)</p>
</li>
<li>
<p><strong>High resolution available with effort</strong>: C, Perl (<code inline="">Time::HiRes</code>), Lua (with socket libs), PHP (<code inline="">microtime()</code>)</p>
</li>
</ul>
<p>Would you like this as a downloadable table or JSON for further processing?</p></body></html>
</body>
</html>

JohnRDOrazio:
Timestamps aren't super useful, seeing we're not using time (seconds or milliseconds) but simply using the year, month and day. And ISO 8601 formatted date strings with midnight time and UTC timezone can easily be instantiated as UTC date objects in most programming languages. So to avoid any confusion between PHP timestamps (in seconds) and javascript timestamps (in milliseconds), the cleanest and clearest approach is probably to just produce ISO 8601 (RFC 3339).

Fixes #333 

## Summary by CodeRabbit

- **Refactor**
  - Liturgical event date fields now use ISO 8601 / RFC 3339 date-time strings instead of numeric timestamps; outputs, messages, serialization, and test runner text reference "date".
- **Chores**
  - Schemas and metadata updated to accept/date-time strings (including pattern constraints); metadata.timestamp minimum lowered.
- **Tests**
  - Test fixtures and assertions converted to ISO 8601 date-time strings; assertion type renamed from timestamp → date.
- **Documentation**
  - Examples, comments, phpdoc and a badge line updated to reflect date-as-string usage.

### DIFF
--- a/src/Models/Calendar/LiturgicalEvent.php
+++ b/src/Models/Calendar/LiturgicalEvent.php
@@ -416,8 +416,13 @@ final class LiturgicalEvent implements \JsonSerializable
         // When we read data from a JSON file, $obj will be an instance of stdClass,
         // and we need to cast the values to types that will be accepted by the LiturgicalEvent constructor
         if ($obj instanceof \stdClass) {
+            // TODO: Will the date ever be a string? Will any JSON src file contain a date as a string?
             if (is_string($obj->date)) {
-                $obj->date = new DateTime($obj->date);
+                $parsed = \DateTime::createFromFormat(\DateTime::ATOM, $obj->date);
+                if (false === $parsed || $parsed->format(\DateTime::ATOM) !== $obj->date) {
+                    throw new \InvalidArgumentException('Property `date` must be a valid RFC 3339 (ISO 8601) date-time string');
+                }
+                $obj->date = $parsed;
             }
 
             if (property_exists($obj, 'color')) {
@@ -585,8 +590,13 @@ final class LiturgicalEvent implements \JsonSerializable
             throw new \InvalidArgumentException('Invalid grade provided to create LiturgicalEvent');
         }
 
-        if (is_int($arr['date'])) {
-            $arr['date'] = new DateTime($arr['date']);
+        // TODO: Will the date ever be a string? Will any JSON src file contain a date as a string?
+        if (is_string($arr['date'])) {
+            $parsed = \DateTime::createFromFormat(\DateTime::ATOM, $arr['date']);
+            if (false === $parsed || $parsed->format(\DateTime::ATOM) !== $arr['date']) {
+                throw new \InvalidArgumentException('Property `date` must be a valid RFC 3339 (ISO 8601) date-time string');
+            }
+            $arr['date'] = $parsed;
         }
 
         $colors = LitColor::GREEN;


JohnRDOrazio:
Actually the Unit Test interface currently checks against the timestamps, perhaps we should keep a `timestamp` property for a smoother transition? Or should we just fix the Unit Tests to check against date-time strings instead of against timestamps?

PR #462, opened 2026-01-03
fix: handle uninitialized Router::$apiFilePath in LoggerFactory

JohnRDOrazio:
## Summary

Fixes fatal error when running the WebSocket server:

```
Fatal error: Typed static property LiturgicalCalendar\Api\Router::$apiFilePath 
must not be accessed before initialization
```

## Root Cause

`LoggerFactory::resolveLogsFolder()` was accessing `Router::$apiFilePath` without checking if it was initialized. The WebSocket server (`LitCalTestServer.php`) creates a `Health` object which uses `LoggerFactory::create()`, but doesn't go through the main `Router` which initializes the static property.

## Solution

Added `getBasePath()` method that uses reflection to check if `Router::$apiFilePath` is initialized before accessing it. Falls back to deriving the path from the package directory structure.

## Test plan

- [x] PHPStan analysis passes (level 10)
- [x] All 168 PHPUnit tests pass
- [x] Code style check passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced log file storage reliability with improved directory resolution and better fallback handling for edge cases.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


JohnRDOrazio:
@coderabbitai generate docstrings

PR #533, opened 2026-04-18
Echo optional runToken in WebSocket responses

### LINKED ISSUE #532: Echo run token in WebSocket responses to isolate test runs
## Problem

The UnitTestInterface sends test requests over a single WebSocket connection. When a user stops a test run and starts a new one, stale responses from the previous run can arrive and be applied to the new run's UI, causing incorrect test result display.

The client-side guard (`currentState === TestState.Stopped`) helps, but cannot distinguish between responses belonging to the current run vs a previous run if the user restarts quickly.

## Proposed Solution

Add support for an optional `runToken` field in WebSocket messages:

1. **Incoming messages**: If a client message includes a `runToken` property, store it per-connection
2. **Outgoing responses**: Attach the stored `runToken` to every response sent back to that connection
3. **Backward compatible**: If no `runToken` is provided, responses behave exactly as before (no `runToken` field in response)

This allows the client to generate a unique token per test run and ignore responses with a mismatched token.

## Implementation Notes

- Store `runToken` per connection (keyed by `resourceId`), similar to the existing `cacheHitCounters` pattern
- Attach the token in `sendMessage()` so all response paths are covered without modifying individual action handlers
- The field is optional — existing clients without run token support are unaffected

JohnRDOrazio:
## Summary

- Store an optional `runToken` from incoming WebSocket messages per-connection (keyed by `resourceId`, like `cacheHitCounters`)
- Attach the stored `runToken` to every outgoing response in `sendMessage()`
- Clean up stored token on connection close
- Fully backward compatible — clients that don't send `runToken` are unaffected

This enables the UnitTestInterface to generate a unique token per test run and discard stale responses from previous runs.

## Test plan

- [ ] Verify existing tests pass (no behavior change for clients not sending `runToken`)
- [ ] Verify PHPStan level 10 passes (`composer analyse` — confirmed locally)
- [ ] Verify phpcs passes (`composer lint` — confirmed locally)
- [ ] Test with UnitTestInterface client sending `runToken` and confirm it is echoed in responses

Closes #532

## Summary by CodeRabbit

* **New Features**
  * Enhanced run token management for WebSocket connections with improved state tracking and persistence across message exchanges.

### DIFF
--- a/src/Health.php
+++ b/src/Health.php
@@ -330,7 +330,12 @@ class Health implements MessageComponentInterface
         /** @var ExecuteValidationSourceFolder|ExecuteValidationSourceFile|ExecuteValidationResource|ValidateCalendar|ExecuteUnitTest $messageReceived */
         $messageReceived = json_decode($msg);
         // Store optional run token for response correlation
-        if ($messageReceived instanceof \stdClass && property_exists($messageReceived, 'runToken') && is_string($messageReceived->runToken)) {
+        if (
+            $messageReceived instanceof \stdClass
+            && property_exists($messageReceived, 'runToken')
+            && is_string($messageReceived->runToken)
+            && preg_match('/^[A-Za-z0-9_\-]{1,64}$/', $messageReceived->runToken)
+        ) {
             $this->runTokens[$resourceId] = $messageReceived->runToken;
         }
         if (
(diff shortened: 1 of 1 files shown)


JohnRDOrazio:
@coderabbitai review

PR #584, opened 2026-05-11
fix(params): make EventsParams national_calendar=VA override order-independent (#576)

### LINKED ISSUE #576: EventsParams: VA branch's `$params` mutation is a no-op (foreach iterates a snapshot)
## Summary

`EventsParams::setParams()` mutates `$params` mid-`foreach` to enforce that requesting `national_calendar=VA` (the General Roman Calendar) forces `locale=la_VA` and `eternal_high_priest=false`. Because PHP `foreach` over an array iterates a copy, those writes are never observed by later iterations of the same loop, so the override only takes effect when the caller happens to pass `national_calendar` **last**.

Found while writing unit tests for #570 — the test in `phpunit_tests/Params/EventsParamsTest.php` had to be reworded to document the actual (order-dependent) behavior rather than the apparent intent.

## Where

`src/Params/EventsParams.php` lines 120-134:

```php
case 'national_calendar':
    if (false === $this->isValidNationalCalendar($value)) {
        $description = "Unknown value `$value` for nation parameter, ...";
        throw new ValidationException($description);
    }
    if ($value === 'VA') {
        $this->Locale                  = LitLocale::LATIN;
        $this->baseLocale              = LitLocale::LATIN_PRIMARY_LANGUAGE;
        $this->EternalHighPriest       = false;
        $params['eternal_high_priest'] = false;   // no-op
        $params['locale']              = LitLocale::LATIN; // no-op
    } else {
        $this->NationalCalendar = strtoupper($value);
    }
    break;
```

## Repro

```php
// Order matters — this *should* end up Latin / EHP=false but does not:
$p = new EventsParams([
    'national_calendar'   => 'VA',     // sets EHP=false, Locale=la_VA
    'eternal_high_priest' => true,     // overwrites with true
    'locale'              => 'en_US',  // overwrites with en_US
]);
// Observed: $p->EternalHighPriest === true, $p->Locale === 'en_US'
// Expected by the VA branch: EHP=false, Locale=la_VA

// Reverse order — works as the VA branch intends:
$p = new EventsParams([
    'eternal_high_priest' => true,
    'locale'              => 'en_US',
    'national_calendar'   => 'VA',
]);
// Observed: $p->EternalHighPriest === false, $p->Locale === 'la_VA'
```

## Why

`foreach ($params as $key => $value)` over a plain array operates on an internal copy of the array's iterator state. Writing to `$params[...]` inside the loop body does not change what the next iteration sees. (See [PHP manual on `foreach`](https://www.php.net/manual/en/control-structures.foreach.php) — "Unless the array is referenced".)

## Impact

- `national_calendar=VA` quietly fails to enforce its locale / EHP invariants when those keys appear after `national_calendar` in the input array.
- API request parsing typically iterates query string / JSON body keys in user-supplied order, so this is reachable from clients without any unusual usage.
- No corresponding bug in `CalendarParams` because that class doesn't try to override sibling params from inside the loop — it just sets its own state and lets the rest of the loop run.

## Suggested fix directions (not prescriptive)

1. **Two-pass parse:** scan once for `national_calendar=VA` and apply the overrides up-front, then iterate the params (skipping `locale` / `eternal_high_priest` if VA was set). Most explicit, no surprising precedence rules.
2. **Iterate by reference:** `foreach ($params as $key => &$value)` *plus* re-reading `$params[$key]` rather than the captured `$value`. Surfaces the intent of the existing code with the smallest diff, but ordering-of-keys still matters.
3. **Apply the VA invariants after the loop:** track whether VA was seen, then clobber `Locale` / `baseLocale` / `EternalHighPriest` once the loop finishes. Probably the closest match to what the existing comment seems to intend.

Option 3 also matches the related branch in `CalendarParams::validateNationalCalendarParam`, which doesn't have this bug because it doesn't try to override siblings.

## Testing

Whichever fix lands, add the two repro cases above as regression tests in `phpunit_tests/Params/EventsParamsTest.php` (the file added in the M1 PR for #570).

## Out of scope

- Changing the *semantics* of which param wins when both are explicit. That's a product decision; this issue is just about the implementation not matching the apparent intent.

JohnRDOrazio:
## Summary

Fixes #576.

`EventsParams::setParams()` was supposed to force `locale=la_VA` and `eternal_high_priest=false` whenever the caller passes `national_calendar=VA`. It tried to do that by writing back into the `$params` array mid-`foreach`, but PHP `foreach` iterates a snapshot, so those writes never reached subsequent iterations. Whichever sibling key the caller passed *after* `national_calendar` would clobber the VA-mandated value.

### Fix

Track a `$forceVaInvariants` flag during the loop and apply the VA overrides once the loop has finished:

```diff
+    $forceVaInvariants = false;
     foreach ($params as $key => $value) {
         // ...
                 case 'national_calendar':
                     // ...
                     if ($value === 'VA') {
-                        $this->Locale                  = LitLocale::LATIN;
-                        $this->baseLocale              = LitLocale::LATIN_PRIMARY_LANGUAGE;
-                        $this->EternalHighPriest       = false;
-                        $params['eternal_high_priest'] = false; // no-op
-                        $params['locale']              = LitLocale::LATIN; // no-op
+                        $forceVaInvariants = true;
                     } else {
                         $this->NationalCalendar = strtoupper($value);
                     }
         // ...
     }
+    if ($forceVaInvariants) {
+        $this->Locale            = LitLocale::LATIN;
+        $this->baseLocale        = LitLocale::LATIN_PRIMARY_LANGUAGE;
+        $this->EternalHighPriest = false;
+    }
```

Sibling input validation still runs inside the loop, so malformed `locale` / `eternal_high_priest` values are still rejected even when VA is set — VA only overrides the *successfully parsed* sibling values, not the validation itself.

### Tests

Three new regression tests in `EventsParamsTest`:

- `testNationalCalendarVaOverridesEvenWhenItComesFirst` — VA first, then `eternal_high_priest=true, locale=en_US` ⇒ Latin / EHP=false
- `testNationalCalendarVaOverridesWhenItComesLast` — same payload in reverse order ⇒ same result (was already passing)
- `testInvalidSiblingValuesStillRejectedEvenWhenVaIsSet` — `eternal_high_priest='maybe'` alongside VA still raises `ValidationException`

The existing `testNationalCalendarVaForcesLatinAndSkipsAssignment` continues to cover the VA-only payload.

## Test plan

- [x] `composer test:quick` — 926 tests / 2939 assertions (the 8 errors are pre-existing integration tests in `Routes/Readonly/*` requiring a live server; not related to this change)
- [x] `composer lint` clean
- [x] `composer analyse` (PHPStan level 10) clean

https://claude.ai/code/session_018tynbsLbtvgLGQr8ma1QKp

---
_Generated by [Claude Code](https://claude.ai/code/session_018tynbsLbtvgLGQr8ma1QKp)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of national calendar parameter handling to ensure uniform behavior regardless of input order.

* **Tests**
  * Added regression tests for national calendar configuration and validation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Liturgical-Calendar/LiturgicalCalendarAPI/pull/584)

### DIFF
--- a/phpunit_tests/Params/EventsParamsTest.php
+++ b/phpunit_tests/Params/EventsParamsTest.php
@@ -128,6 +128,22 @@ final class EventsParamsTest extends TestCase
         ]);
     }
 
+    public function testRepeatedSetParamsVaClearsPreviousNationalCalendar(): void
+    {
+        // setParams() is public; switching from a real nation to VA on the same
+        // instance must clear NationalCalendar so the post-loop VA invariants
+        // fully describe the request shape (no stale nation override left over).
+        $params = new EventsParams(['national_calendar' => 'IT']);
+        self::assertSame('IT', $params->NationalCalendar);
+
+        $params->setParams(['national_calendar' => 'VA']);
+
+        self::assertNull($params->NationalCalendar);
+        self::assertSame(LitLocale::LATIN, $params->Locale);
+        self::assertSame(LitLocale::LATIN_PRIMARY_LANGUAGE, $params->baseLocale);
+        self::assertFalse($params->EternalHighPriest);
+    }
+
     public function testUnknownNationalCalendarIsRejected(): void
     {
         $this->expectException(ValidationException::class);
--- a/phpunit_tests/Params/EventsParamsTest.php
+++ b/phpunit_tests/Params/EventsParamsTest.php
@@ -82,6 +82,52 @@ final class EventsParamsTest extends TestCase
         self::assertNull($params->NationalCalendar);
     }
 
+    /**
+     * Regression for #576: when national_calendar=VA appears anywhere in the
+     * input, its locale=la_VA / eternal_high_priest=false invariants must win
+     * regardless of where it sits relative to sibling keys.
+     */
+    public function testNationalCalendarVaOverridesEvenWhenItComesFirst(): void
+    {
+        $params = new EventsParams([
+            'national_calendar'   => 'VA',
+            'eternal_high_priest' => true,
+            'locale'              => 'en_US',
+        ]);
+
+        self::assertSame(LitLocale::LATIN, $params->Locale);
+        self::assertSame(LitLocale::LATIN_PRIMARY_LANGUAGE, $params->baseLocale);
+        self::assertFalse($params->EternalHighPriest);
+        self::assertNull($params->NationalCalendar);
+    }
+
+    public function testNationalCalendarVaOverridesWhenItComesLast(): void
+    {
+        $params = new EventsParams([
+            'eternal_high_priest' => true,
+            'locale'              => 'en_US',
+            'national_calendar'   => 'VA',
+        ]);
+
+        self::assertSame(LitLocale::LATIN, $params->Locale);
+        self::assertSame(LitLocale::LATIN_PRIMARY_LANGUAGE, $params->baseLocale);
+        self::assertFalse($params->EternalHighPriest);
+        self::assertNull($params->NationalCalendar);
+    }
+
+    public function testInvalidSiblingValuesStillRejectedEvenWhenVaIsSet(): void
+    {
+        // VA only overrides the *successfully* parsed sibling values; it does
+        // not silence input validation for malformed siblings.
+        $this->expectException(ValidationException::class);
+        $this->expectExceptionMessage('eternal_high_priest');
+
+        new EventsParams([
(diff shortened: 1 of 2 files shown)


JohnRDOrazio:
@coderabbitai review

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #577, opened 2026-05-11
test(coverage): unit-test src/Params and src/Database (issue #570 M1)

JohnRDOrazio:
Adds 95 in-process unit tests covering all seven src/Params classes plus
src/Database/Connection. These run without the dev API server or Postgres,
so they exercise paths that the existing HTTP-based integration suite cannot
hit deterministically (env-driven branches, validation rejections, singleton
lifecycle).

CalendarParams and EventsParams resolve calendar metadata via
Utilities::jsonUrlToObject in their constructors; tests redirect that fetch
to a local file:// URL by overriding Router::\$apiPath against a small
fixture in phpunit_tests/fixtures/api/calendars. MissalsParams tests stand
in MissalsHandler::\$missalsIndex with a hand-built MissalMetadataMap,
restoring the previous value in tearDownAfterClass.

This is the M1 milestone (Params + Database) of the layered plan in #570.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for database connection lifecycle, configuration validation, and error handling.
  * Added multiple API parameter validation test suites covering calendar, events, missals, easter, decrees, and regional data behaviors (validation, parsing, and canonicalization).
  * Added bundled fixtures with calendar metadata, locale mappings, and diocesan calendar definitions to support deterministic parameter tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Liturgical-Calendar/LiturgicalCalendarAPI/pull/577)

PR #585, opened 2026-05-11
fix(params): reject unsupported locale in EventsParams instead of silent Latin fallback (#578)

### LINKED ISSUE #578: EventsParams silently falls back to Latin on unsupported locale, while sibling Params classes throw
## Summary

`EventsParams::setParams()` silently falls back to `LitLocale::LATIN` when given an unsupported locale, while every other params class in `src/Params/` throws a `ValidationException` for the same input. This is an inconsistency in input validation worth reconciling.

Surfaced while writing unit tests for #570 (M1, PR #577) — the test `EventsParamsTest::testUnsupportedLocaleFallsBackToLatin` had to be written to document the current behavior rather than the expected-by-convention behavior.

## Behavior matrix

Input: `['locale' => 'not-a-locale']` (canonicalisable but not in `LitLocale::$AllAvailableLocales`)

| Class | Result |
| --- | --- |
| `EventsParams` | Silently sets `Locale = la_VA`, `baseLocale = la` |
| `DecreesParams` | `ValidationException("Invalid value … for param `locale` …")` |
| `EasterParams` | `ValidationException("Invalid value … for param `locale` …")` |
| `CalendarParams` | `ValidationException("Invalid value … for parameter `locale` …")` |
| `MissalsParams` | `ValidationException("Locale … not supported by this server …")` |
| `RegionalDataParams` | `ValidationException("Invalid value … for param `locale` …")` |

## Where

`src/Params/EventsParams.php` lines ~106-119:

```php
case 'locale':
    $locale = \Locale::canonicalize($value);
    if (null === $locale) {
        throw new ValidationException('Invalid locale string: ' . $value);
    }

    $this->Locale = LitLocale::isValid($locale) ? $locale : LitLocale::LATIN;
    $baseLocale   = \Locale::getPrimaryLanguage($this->Locale);
    if (null === $baseLocale) {
        $description = '"The evil spirit had bound his tongue, ..." — St. John Chrysostom, Homily 32 on Matthew';
        throw new ValidationException($description);
    }
    $this->baseLocale = $baseLocale;
    break;
```

The ternary `LitLocale::isValid($locale) ? $locale : LitLocale::LATIN` is the silent-fallback branch. Compare with `DecreesParams::setParams`, which throws if `LitLocale::isValid($value)` is false.

## Decision needed

Either:

1. **Tighten EventsParams** to throw `ValidationException` like the rest of the family. Pro: consistent contract across the API surface; clients can't accidentally request a German-only events list and silently get Latin back. Con: breaking change for any client relying on the fallback.
2. **Document the fallback as intentional** and explain why `/events` is permissive where the other endpoints are strict. Update the test name / docblock to match, and consider whether the same fallback should apply elsewhere.

Option 1 matches what readers of the codebase would expect by default; option 2 is fine if there's a product reason (e.g., `/events` is meant to "always answer something") but the rationale isn't visible from the code.

## Suggested fix (if option 1)

```diff
-        $this->Locale = LitLocale::isValid($locale) ? $locale : LitLocale::LATIN;
+        if (false === LitLocale::isValid($locale)) {
+            $description = "Invalid value `{$locale}` for param `locale`, valid values are: la, la_VA, "
+                . implode(', ', LitLocale::$AllAvailableLocales);
+            throw new ValidationException($description);
+        }
+        $this->Locale = $locale;
```

## Testing

Once a direction is picked, update `phpunit_tests/Params/EventsParamsTest.php::testUnsupportedLocaleFallsBackToLatin` accordingly — either rename + invert it to assert the exception, or rename to make the "permissive by design" intent explicit.

## Related

- #570 — parent coverage epic
- #576 — separate bug in `EventsParams` where the VA branch's `$params` mutation is a no-op
- PR #577 — the M1 PR where this finding surfaced

JohnRDOrazio:
## Summary

Fixes #578.

`EventsParams::setParams()` silently fell back to `LitLocale::LATIN` when given a canonicalisable but unsupported locale, while every other params class in `src/Params/` (`DecreesParams`, `EasterParams`, `CalendarParams`, `MissalsParams`, `RegionalDataParams`) raises `ValidationException` for the same input. A client asking for an events list in (say) Vietnamese got Latin back with no signal that the requested locale wasn't available.

### Fix

```diff
-    $this->Locale = LitLocale::isValid($locale) ? $locale : LitLocale::LATIN;
+    if (false === LitLocale::isValid($locale)) {
+        $description = "Invalid value `$locale` for param `locale`, valid values are: la, la_VA, "
+            . implode(', ', LitLocale::$AllAvailableLocales);
+        throw new ValidationException($description);
+    }
+    $this->Locale = $locale;
```

Mirrors `DecreesParams::setParams` exactly — same message format, same enum-list disclosure.

### Tests

- **`testUnsupportedLocaleIsRejected`** — renamed and inverted from the previous `testUnsupportedLocaleFallsBackToLatin`, which had been added in M1 (#577) specifically to document the silent fallback as a known wart. Now asserts `ValidationException`.
- **`testEmptyLocaleStringIsRejected`** — empties don't actually hit the upstream null-canonicalize guard because `Locale::canonicalize('')` returns `en_US_POSIX` on this build; they fall into the new unsupported-locale arm instead. Either way the caller gets `ValidationException`.

### Compatibility note

This is a **breaking change** for any client that was relying on the silent fallback. The behaviour was undocumented and inconsistent with the rest of the API surface, so the practical impact should be low — but worth flagging in release notes.

### Stacks with #584

PR #584 (fix for #576) also touches `EventsParams::setParams`. Both PRs are small and the conflict is mechanical; whichever lands first, the second will need a tiny rebase.

## Test plan

- [x] `composer test:quick` — 924 tests / 2931 assertions (the 8 errors are pre-existing integration tests in `Routes/Readonly/*` requiring a live server; not related to this change)
- [x] `composer lint` clean
- [x] `composer analyse` (PHPStan level 10) clean

https://claude.ai/code/session_018tynbsLbtvgLGQr8ma1QKp

---
_Generated by [Claude Code](https://claude.ai/code/session_018tynbsLbtvgLGQr8ma1QKp)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter locale validation: Unsupported and empty locale values are now rejected with validation errors instead of silently falling back to a default locale.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Liturgical-Calendar/LiturgicalCalendarAPI/pull/585)

### DIFF
--- a/phpunit_tests/Params/EventsParamsTest.php
+++ b/phpunit_tests/Params/EventsParamsTest.php
@@ -52,14 +52,29 @@ final class EventsParamsTest extends TestCase
         self::assertSame('en', $params->baseLocale);
     }
 
-    public function testUnsupportedLocaleFallsBackToLatin(): void
+    public function testUnsupportedLocaleIsRejected(): void
     {
-        // Per current behavior, when a locale is canonicalisable but unsupported,
-        // it silently falls back to Latin rather than rejecting outright.
-        $params = new EventsParams(['locale' => 'not-a-locale']);
+        // Regression for #578: EventsParams now matches the rest of the
+        // src/Params/ family — an unsupported (but canonicalisable) locale
+        // raises ValidationException rather than silently falling back to
+        // Latin.
+        $this->expectException(ValidationException::class);
+        $this->expectExceptionMessage('Invalid value');
+        $this->expectExceptionMessage('param `locale`');
 
-        self::assertSame(LitLocale::LATIN, $params->Locale);
-        self::assertSame(LitLocale::LATIN_PRIMARY_LANGUAGE, $params->baseLocale);
+        new EventsParams(['locale' => 'not-a-locale']);
+    }
+
+    public function testEmptyLocaleStringIsRejected(): void
+    {
+        // Locale::canonicalize('') returns 'en_US_POSIX' on this build, which
+        // is unsupported — so empty strings end up in the new unsupported-locale
+        // arm rather than the upstream null-canonicalize guard. Either way, the
+        // user gets a ValidationException with the locale-rejection message.
+        $this->expectException(ValidationException::class);
+        $this->expectExceptionMessage('param `locale`');
+
+        new EventsParams(['locale' => '']);
     }
 
     public function testNationalCalendarFromMetadataIsAccepted(): void
--- a/src/Params/EventsParams.php
+++ b/src/Params/EventsParams.php
@@ -117,7 +117,12 @@ class EventsParams implements ParamsInterface
                             throw new ValidationException('Invalid locale string: ' . $value);
                         }
 
-                        $this->Locale = LitLocale::isValid($locale) ? $locale : LitLocale::LATIN;
+                        if (false === LitLocale::isValid($locale)) {
+                            $description = "Invalid value `$locale` for param `locale`, valid values are: la, la_VA, "
+                                . implode(', ', LitLocale::$AllAvailableLocales);
+                            throw new ValidationException($description);
+                        }
+                        $this->Locale = $locale;
                         $baseLocale   = \Locale::getPrimaryLanguage($this->Locale);
                         if (null === $baseLocale) {
                             $description = '“The evil spirit had bound his tongue, and together with his tongue had fettered his soul.” — St. John Chrysostom, Homily 32 on Matthew';


JohnRDOrazio:
@coderabbitai review

---
_Generated by [Claude Code](https://claude.ai/code)_

JohnRDOrazio:
@coderabbitai review

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #603, opened 2026-05-18
fix(schema): drop redundant idx_api_keys_hash, add idx_api_keys_application_id

JohnRDOrazio:
Closes #601.

## Summary

New migration `Version20260519000000` fixes two `api_keys` index issues flagged by CodeRabbit in PR #599's baseline review:

| Issue | Fix |
|-------|-----|
| `idx_api_keys_hash` is redundant — `key_hash` is `UNIQUE NOT NULL`, so Postgres already maintains a unique B-tree index (`api_keys_key_hash_key`) for it. The extra non-unique index just costs writes and disk. | `DROP INDEX IF EXISTS idx_api_keys_hash` |
| `idx_api_keys_application_id` is missing — `application_id` is `NOT NULL REFERENCES applications(id) ON DELETE CASCADE`, and Postgres does NOT auto-index FK columns. "List keys for app X" and cascade-deletes seq-scan today. | `CREATE INDEX IF NOT EXISTS idx_api_keys_application_id ON api_keys(application_id)` |

## Why a new migration (not editing the baseline)

Per the issue: the baseline `Version20260518120000` is meant to be a faithful snapshot of what shipped to staging on 2026-05-18. Mixing schema improvements into it would break the "baseline = production reality" invariant, and CodeRabbit's findings wouldn't be attributable to a specific decision or independently rollback-able.

## Idempotency

`DROP INDEX IF EXISTS` and `CREATE INDEX IF NOT EXISTS` make the migration safe to re-run against any environment where the fix was applied out of band.

## No init-db.sql changes

PR #602 (issue #600) just stripped the application DDL — including `idx_api_keys_hash` — from `scripts/init-db.sql`. Migrations are now the single source of truth, so this PR is a pure new-migration with no SQL bootstrap touch.

## Behavior on each environment

| Environment | State before | After this migration |
|-------------|--------------|----------------------|
| Fresh (Doctrine creates schema) | Baseline runs → `idx_api_keys_hash` + unique constraint | DROP IF EXISTS removes `idx_api_keys_hash`; FK index added. Final state: one unique index on `key_hash`, one regular index on `application_id`, one on `key_prefix`. |
| Staging (pre-marked baseline, schema bootstrapped via init-db.sql 2026-05-18) | Has both `idx_api_keys_hash` and the unique constraint-backed index | This migration drops the redundant one, adds the FK index. |
| Local dev volumes from before PR #602 | Same as staging (init-db.sql ran with old DDL) | Same fix applies. |

## Test plan

[...]

PR #607, opened 2026-05-18
ci(deploy): retry transient SSH/rsync failures + add curl --retry to migrate

JohnRDOrazio:
PR #605's deploy red-lighted at the rsync step with `kex_exchange_identification: read: Connection reset by peer` — the remote SSH daemon reset the connection before key exchange even completed. The next manual rerun succeeded on the first attempt, confirming it was transient. Three plausible causes: fail2ban backoff window (PR #603 deployed 12 min before), `MaxStartups` contention, or a brief OVH network blip.

Whitelisting GitHub Actions runner IPs upstream isn't practical — the pool is ~10k+ Azure ranges rotating weekly, which is effectively "most of Azure North America" and so isn't really an allowlist. Workflow-side retry is the cheapest mitigation.

## Changes

| Step | Change | Why |
|------|--------|-----|
| `Ensure deploy + cache + logs dirs exist` (SSH) | Wrap ssh in 3-attempt loop, 10s/20s backoff | First SSH op of the deploy — same connection-reset surface as rsync |
| `Deploy via rsync` (rsync over SSH) | Wrap rsync in 3-attempt loop, 10s/20s backoff | The exact failure that triggered this PR. SSH key exchange fails before rsync has any bytes on the wire, so wrapping rsync (vs. rsync `--partial`) is the right level |
| `Run database migrations` (HTTPS curl) | Add `--retry 2 --retry-delay 5` | Transient TCP-layer failures on the curl. `/_ops/migrate` is safe to retry (Doctrine `transactional: true` rolls back partial runs cleanly) |
| `Reset OPcache` | Unchanged | Already has `curl --retry 3 --retry-delay 3` from PR #598 |

## Worst-case timing

- mkdir step: 30s extra wait before failing hard (was: instant fail)
- rsync step: 30s extra wait before failing hard
- migrate curl: 10s extra wait before failing hard

Total worst-case added latency: 70 seconds, only on retry. Happy-path deploys see zero overhead — the loops execute the underlying command once and break.

## No new deps

Hand-rolled bash loop pattern, no third-party action introduced. Keeps the supply-chain footprint identical.

## Retry logic, generalized

Same shape used for both SSH steps:

```bash
attempt=1
max=3
while true; do
  if <command>; then break; fi
  if [ "$attempt" -ge "$max" ]; then
    echo "::error::<cmd> failed after ${max} attempts" >&2
    exit 1
  fi
  delay=$((attempt * 10))
  echo "::warning::<cmd> attempt ${attempt} failed; retrying in ${delay}s..." >&2
  sleep "$delay"
  attempt=$((attempt + 1))
done
```

GitHub Actions surfaces the `::warning::` line as a yellow annotation on the run summary, so even successful-on-retry deploys leave a visible breadcrumb.

## Doctrine `transactional: true` — why migrate is safe to retry

From `doctrine-migrations.php`:

```php
'all_or_nothing'  => true,
'transactional'   => true,
```

If the HTTPS connection drops while Doctrine is mid-migration, the SQL transaction rolls back, the version row never gets inserted into `doctrine_migration_versions`, and the next attempt re-runs the whole migration from scratch. No partial state to clean up.

## Future considerations (not in this PR)

- **fail2ban allowlist** of `api.github.com/meta`'s `actions` ranges via a daily cron on the VPS — only worthwhile if retries-with-backoff prove insufficient. Adds maintenance overhead (the IP set drifts).
- **Self-hosted GitHub runner** inside the VPS network — eliminates the SSH-from-internet hop entirely. Heaviest option, best long-term posture, but a new service to maintain.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/deploy.yaml'))"` — workflow parses cleanly after the edits
- [ ] Next real deploy (next merge to `development`): happy-path completes with zero retries, single rsync invocation in the log
- [ ] First post-merge deploy that hits a transient: `::warning::rsync attempt 1 failed; retrying in 10s...` appears as a yellow annotation, then the deploy succeeds on attempt 2 or 3 — what this PR exists to achieve

## Summary by CodeRabbit

* **Chores**
  * Enhanced deployment robustness with automatic retry logic for transient SSH and network connectivity failures. Directory creation, file synchronization, and database migrations now include built-in recovery with backoff strategies (up to 3 retry attempts), reducing deployment failures caused by temporary connection disruptions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Liturgical-Calendar/LiturgicalCalendarAPI/pull/607?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -214,15 +214,32 @@ jobs:
             "${DEPLOY_PATH}/cache" \
             "${DEPLOY_PATH}/public/engineCache" \
             "${DEPLOY_PATH}/logs")
-          ssh -i ~/.ssh/deploy_key -p "${VPS_PORT}" \
-            -o IdentitiesOnly=yes \
-            -o StrictHostKeyChecking=yes \
-            -o UserKnownHostsFile=~/.ssh/known_hosts \
-            -o ConnectTimeout=10 \
-            -o ServerAliveInterval=15 \
-            -o ServerAliveCountMax=2 \
-            "${VPS_USER}@${VPS_HOST}" \
-            "${remote_cmd}"
+          # Retry on transient SSH failures (connection reset, fail2ban
+          # backoff, MaxStartups contention). 3 attempts with 10s/20s
+          # backoff — total worst-case wait 30s before giving up.
+          attempt=1
+          max=3
+          while true; do
+            if ssh -i ~/.ssh/deploy_key -p "${VPS_PORT}" \
+                -o IdentitiesOnly=yes \
+                -o StrictHostKeyChecking=yes \
+                -o UserKnownHostsFile=~/.ssh/known_hosts \
+                -o ConnectTimeout=10 \
+                -o ServerAliveInterval=15 \
+                -o ServerAliveCountMax=2 \
+                "${VPS_USER}@${VPS_HOST}" \
+                "${remote_cmd}"; then
+              break
+            fi
+            if [ "$attempt" -ge "$max" ]; then
+              echo "::error::ssh mkdir failed after ${max} attempts" >&2
+              exit 1
+            fi
+            delay=$((attempt * 10))
+            echo "::warning::ssh mkdir attempt ${attempt} failed; retrying in ${delay}s..." >&2
+            sleep "$delay"
+            attempt=$((attempt + 1))
+          done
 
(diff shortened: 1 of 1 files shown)


JohnRDOrazio:
@coderabbitai review

PR #663, opened 2026-06-24
refactor(api): eliminate the /calendars HTTP self-call (#659)

### LINKED ISSUE #659: Eliminate the /calendars HTTP self-call in RegionalDataHandler and EventsParams
## Background

`RegionalDataHandler::__construct()` and `EventsParams` build their calendar metadata by making an **HTTP request to the API's own `/calendars` endpoint**:

```php
$metadataObj = Utilities::jsonUrlToObject(Route::CALENDARS->path());
$this->CalendarsMetadata = MetadataCalendars::fromObject($metadataObj->litcal_metadata);
```

This self-call is fragile and has already produced two distinct production/CI failures (both fixed in #658):

1. **Unroutable host** — under `php -S 0.0.0.0:8000`, `SERVER_NAME` is the wildcard `0.0.0.0`, so the self-call targeted `http://0.0.0.0:8000/calendars`, which isn't reliably routable inside a container. (Fixed by normalizing the host to loopback.)
2. **Self-throttling** — the unauthenticated self-call (from loopback) was charged against the per-IP `UNAUTHENTICATED_RATE_LIMIT` bucket, so after N internal calls the API returned 429 to **itself**. (Fixed by exempting loopback from the rate limit.)

Both fixes are band-aids around the underlying smell: **the API loops back through the network to read data it already has locally.** It also adds latency, consumes a worker, and runs *before* the middleware pipeline (so its failures bypass `ErrorHandlingMiddleware`).

## Proposal

Build `MetadataCalendars` **directly from local source data** instead of via HTTP — the same way `MetadataHandler::buildIndex()` already does:

```php
self::$metadataCalendars = new MetadataCalendars();
MetadataHandler::buildNationalCalendarData();
MetadataHandler::buildDiocesanCalendarData();
MetadataHandler::buildWiderRegionData();
MetadataHandler::buildLocales();
```

Extract that index-building into a reusable service/factory that both `MetadataHandler` and the self-call sites (`RegionalDataHandler`, `EventsParams`) can call in-process. This removes the HTTP hop entirely and makes the host-normalization + loopback-exemption workarounds unnecessary for these paths.

## Affected code

- `src/Handlers/RegionalDataHandler.php` (constructor, ~line 73)
- `src/Params/EventsParams.php` (~line 71)
- `src/Handlers/MetadataHandler.php` (`buildIndex()` and the `build*` helpers — candidate to extract)
- `src/Utilities.php` (`jsonUrlToObject` / `rawContentsFromUrl` self-call usage)

## Acceptance

- `RegionalDataHandler` and `EventsParams` no longer issue an HTTP request to `/calendars`.
- Equivalent metadata is produced (existing tests for `/data` and `/events` stay green).
- The wildcard-host normalization and loopback rate-limit exemption remain valid for any *other* self-calls but are no longer required by these paths.

## Context

Surfaced while making the frontend RBAC e2e suite (`LiturgicalCalendarFrontend`) green; see #658 for the two interim fixes.

JohnRDOrazio:
Closes #659.

## Summary

`RegionalDataHandler`, `EventsParams`, and `CalendarParams` built their calendars metadata by issuing an HTTP request to the API's own `/calendars` endpoint and re-parsing the response via `MetadataCalendars::fromObject()`. That self-call is fragile (it produced the two failures fixed in #658 — wildcard host and self-throttling), adds latency, consumes a worker, and runs *before* the middleware pipeline so its failures bypass `ErrorHandlingMiddleware`.

This builds the metadata index **in-process from local source data** instead, via a new single-source-of-truth service.

## Changes

- **New `src/Services/CalendarMetadataProvider.php`** — `create(): MetadataCalendars` builds the index from local source data. The `build*` logic moved here out of `MetadataHandler`.
- **`MetadataHandler`** (serving `GET /calendars`) now delegates to the provider.
- **Three former self-call sites** now build in-process (no HTTP, no `fromObject`):
  - `RegionalDataHandler::__construct`
  - `EventsParams::__construct`
  - `CalendarParams::__construct` *(the issue listed the first two; this is the same anti-pattern in a third spot, fixed for completeness)*
- `Health.php`'s `/calendars` call is **intentionally retained** — it deliberately exercises the live endpoint as a health check.
- The wildcard-host normalization and loopback rate-limit exemption from #658 remain valid for any other self-calls but are no longer required by these paths.

## Single source of truth — provable equivalence

Consumers now use the built object directly instead of round-tripping it through JSON. `CalendarMetadataProviderTest::testBuiltObjectIsIdenticalToNetworkRoundTrip` guarantees that's safe: it serializes the built object exactly as the `/calendars` endpoint does, re-parses it through the same `fromObject()` path consumers used, re-serializes, and asserts **byte-for-byte identical JSON**. The round-trip is idempotent, so "use the built object directly" is indistinguishable from "fetch it over the network".

## Bonus fix — latent locale flakiness

While verifying, I found a pre-existing order-dependent flake (`EventsParamsTest::testEmptyLocaleStringIsRejected`) rooted in a real product bug: an empty `locale` param was run through `\Locale::canonicalize('')`, which resolves to the ambient ICU default (`\Locale::getDefault()`) — and that default can be mutated within a worker by a prior request (`CalendarHandler` calls `\Locale::setDefault()`). So an empty locale was accepted or rejected depending on prior request state. The second commit adds an explicit empty/whitespace guard in both `EventsParams` and `CalendarParams`, making rejection deterministic, plus coverage.

## Acceptance criteria (#659)

- [x] `RegionalDataHandler` and `EventsParams` (and `CalendarParams`) no longer issue an HTTP request to `/calendars`
- [x] Equivalent metadata produced (existing `/data`, `/events`, `/calendars` tests stay green; idempotency proof added)
- [x] Wildcard-host normalization and loopback rate-limit exemption remain valid for other self-calls but are no longer required by these paths

## Validation

- PHPStan level 10 (full): no errors
- phpcs: clean
- `composer test:quick` (full suite): **1328 tests, 152k assertions, 0 failures** (3 pre-existing skips)

## Summary by CodeRabbit

* **New Features**
  * Calendar and regional metadata are now generated directly from bundled data, improving reliability and reducing dependence on live metadata lookups.

* **Bug Fixes**
  * Locale fields now correctly reject empty or whitespace-only values.
  * Metadata responses are now more consistent and deterministic across repeated requests and test environments.

### DIFF
--- a/phpunit_tests/Handlers/EventsHandlerTest.php
+++ b/phpunit_tests/Handlers/EventsHandlerTest.php
@@ -7,23 +7,16 @@ namespace LiturgicalCalendar\Tests\Handlers;
 use LiturgicalCalendar\Api\Handlers\EventsHandler;
 use LiturgicalCalendar\Api\Http\Exception\UnprocessableContentException;
 use LiturgicalCalendar\Api\Http\Exception\ValidationException;
-use LiturgicalCalendar\Api\Router;
 use PHPUnit\Framework\Attributes\CoversClass;
 
 #[CoversClass(EventsHandler::class)]
 final class EventsHandlerTest extends AbstractHandlerTestCase
 {
-    protected function setUp(): void
-    {
-        parent::setUp();
-        // EventsHandler constructs an EventsParams, which fetches
-        // /calendars metadata via file_get_contents during its constructor.
-        // Point Router::$apiPath at the file:// fixture introduced in M1 so
-        // the fetch is deterministic and offline.
-        $fixturePath = realpath(__DIR__ . '/../fixtures/api');
-        self::assertNotFalse($fixturePath, 'M1 calendars fixture must be present');
-        Router::$apiPath = 'file://' . $fixturePath;
-    }
+    // EventsHandler constructs an EventsParams, which now builds the calendars
+    // metadata index in-process from local source data (CalendarMetadataProvider)
+    // rather than fetching /calendars over HTTP. AbstractHandlerTestCase already
+    // pins Router::$apiFilePath to the project root, so the build resolves
+    // against the bundled sourcedata with no HTTP server or fixture needed.
 
     public function testOptionsPreflightSucceeds(): void
     {
--- a/phpunit_tests/Handlers/RegionalDataHandlerTest.php
+++ b/phpunit_tests/Handlers/RegionalDataHandlerTest.php
@@ -7,7 +7,6 @@ namespace LiturgicalCalendar\Tests\Handlers;
 use LiturgicalCalendar\Api\Handlers\RegionalDataHandler;
 use LiturgicalCalendar\Api\Http\Exception\UnprocessableContentException;
 use LiturgicalCalendar\Api\Http\Exception\ValidationException;
-use LiturgicalCalendar\Api\Router;
 use PHPUnit\Framework\Attributes\CoversClass;
 
 /**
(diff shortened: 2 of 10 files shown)


JohnRDOrazio:
@coderabbitai review

PR #688, opened 2026-07-06
fix(ics): escape description newlines once and restore liturgical rank

JohnRDOrazio:
## Summary

Fixes two long-standing regressions in the ICS (`text/calendar`) output:

- **Literal `\n` shown in Google Calendar** — the `DESCRIPTION` property was assembled with pre-escaped literal `'\n'` sequences, a holdover from before `escapeIcal()` existed. Since 958085e1 introduced `escapeIcal()`, its backslash doubling turned them into `\\n`, which RFC 5545 clients unescape to a literal backslash + `n` instead of a line break. The description is now built with real newlines and escaped exactly once by `escapeIcal()`.
- **Liturgical rank (memorial, feast, …) never displayed** — unlike the HTML `X-ALT-DESC` branch, the plain-text `$displayGrade` had no fallback to the localized grade when `grade_display` was `null` (the common case). The fallback existed in the original 2021 implementation and was lost in a refactor. The plain-text branch now mirrors the HTML logic, including the `DedicationLateran` special case.

Also joins only non-empty description parts, so events without commons no longer start with a blank line.

Before / after for a memorial (Latin locale):

```text
DESCRIPTION:De Commune Beatæ Mariæ Virginis\\n\\nalbus
DESCRIPTION:De Commune Beatæ Mariæ Virginis\nmemoria ad libitum\nalbus
```

## Tests

Two regression tests in `CalendarHandlerTest` request the 2025 calendar as `text/calendar` and assert (on the unfolded body) that no double-escaped sequences are present and that the localized rank appears in `DESCRIPTION`. They delete cached `engineCache/*.ics` files first, since a cache file produced by earlier buggy code would otherwise mask regressions.

- `composer test:quick`: 1472 tests pass (3 environment-dependent skips)
- `composer analyse` (PHPStan level 10): clean
- `composer lint`: clean

## Deployment note

⚠️ The response cache under `engineCache/` is keyed by API version + source-data hash only, so deploying this fix does **not** invalidate previously cached `.ics` responses. Clear `engineCache/v*/​*.ics` on the server after deploying (or bump the API version).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved calendar export text so descriptions appear correctly formatted in ICS files, avoiding double-escaped line breaks.
  * Ensured memorial entries include the localized liturgical grade in the calendar description.
  * Updated feast-day labeling in calendar exports so the correct grade is shown consistently, including special cases.

### DIFF
--- a/phpunit_tests/Handlers/CalendarHandlerTest.php
+++ b/phpunit_tests/Handlers/CalendarHandlerTest.php
@@ -144,4 +144,72 @@ final class CalendarHandlerTest extends AbstractHandlerTestCase
             'Fallback published_at must be an RFC3339 UTC timestamp for the ICS CREATED field'
         );
     }
+
+    /**
+     * Request the 2025 calendar as ICS (Latin locale for deterministic grade
+     * strings) and return the response body with RFC 5545 line folding undone,
+     * so assertions can match logical content lines directly.
+     */
+    private function fetchUnfoldedIcs(): string
+    {
+        // The handler serves cached responses from engineCache/ when one
+        // exists for the same API version + source-data hash; a cache file
+        // produced by earlier (possibly buggy) code would mask regressions,
+        // so force a fresh ICS computation.
+        foreach (glob(dirname(__DIR__, 2) . '/engineCache/v*/*.ics') ?: [] as $staleIcs) {
+            unlink($staleIcs);
+        }
+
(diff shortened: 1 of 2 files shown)


JohnRDOrazio:
@coderabbitai review

PR #691, opened 2026-07-07
fix(calendar): move national-calendar events that coexist with their replacement

### LINKED ISSUE #271: [Bug]: if a national calendar moves a celebration that was suppressed in the General Roman Calendar, an error occurs
### What happened?

It seems that celebrations (optional memorials, memorials) that are suppressed in the General Roman Calendar for a given year because of a Sunday, Solemnity, or Feast, are no longer accessible to National Calendars that move the celebration to a different date (usually to give way to a proper memorial or optional memorial).

We should store all suppressed celebrations in the newly created `FestivityCollection::suppressedEvents` collection, instead of simply not creating them.

The specific case that brought this to light, is the National Calendar for Canada, which moves the optional memorial `StRayPenyafort` from Jan 7 to Jan 8 in order to prefer the proper Memorial `StAndreBessette` on Jan 7. In the year 2024, Jan 7 falls on a Sunday so `StRayPenyafort` is not created, and can no longer be moved.

### Version

dev (Edge)

### Which browser or browsers are you seeing the problem on? (can select more than one, or N/A if not applicable)

N/A

### Relevant error messages

_No response_

### LINKED ISSUE #690: [Bug]: St. Camillus has a different feast day in the US vs General Roman Calendar
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

St. Camillus has a feast day that falls on July 18 in the US but July 14 everywhere else. LitCal currently shows July 14 only.

https://en.wikipedia.org/wiki/Camillus_de_Lellis

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
- **OS**:
- **PHP**:
```

### Anything else?

```markdown

```

### Version

dev (Unstable)

JohnRDOrazio:
## Summary

Fixes #690 — St Camillus de Lellis stayed on July 14 in the US calendar instead of moving to July 18.

The `CalEventAction::MoveEvent` case in `handleNationalCalendarEvents()` had a missing branch. It handled:

1. target date occupied/Sunday + event exists → refuse, remove, message ✓
2. target date free + event **absent** (suppressed by a higher-ranking event) → recover from suppressed collection and move ✓
3. target date free + event **still present** → **silent no-op** ✗

Case 3 is exactly St Camillus: Blessed Kateri Tekakwitha's memorial coexists with (does not suppress) his optional memorial on July 14, so he was always still present when the moveEvent item was processed — and nothing happened. The same silent no-op affected `StElizabethPortugal` (stuck on July 4 in years where July 5 is free) and `StVincentDeacon` (stuck on Jan 22). `StPaulCross` kept moving correctly only because the Brébeuf memorial suppresses him first, sending him down working branch 2.

The fix delegates case 3 to `moveLiturgicalEventDate()`, which already implements the event-still-present move (including the "is transferred from … to …" message) — it just was never reached from this call site.

## When it broke

2928a146 (2024-12-05, "Fixes #271"): the original `moveFestivityDate()` helper handled both the present-event and suppressed-event cases; when its guard checks were inlined into the switch case, only the suppressed-recovery path was kept. Production v5.7 has the same behavior (verified: Camillus on July 14 in 2026, absent entirely in 2010 where July 18 fell on a Sunday and the refuse-and-remove branch fired).

## Tests

Two regression tests (written first, red before the fix, green after) exercise both US moveEvent entries whose source events coexist with their replacements:

- US 2026: `StCamillusDeLellis` → 2026-07-18
- US 2025: `StElizabethPortugal` → 2025-07-05

They purge `engineCache/` response caches before computing, so cached pre-fix responses can't mask a regression (the ICS tests' cache clearing is factored into the same `purgeEngineCache()` helper).

- `composer test:quick`: 1474 tests pass (3 environment-dependent skips)
- `composer analyse` (PHPStan level 10): clean
- `composer lint`: clean

## Deployment note

⚠️ `engineCache/` responses are keyed by API version + source-data hash, so previously cached US/CA calendar responses will keep serving the un-moved dates until the cache is cleared or the source data changes. Clear `engineCache/v*/` response files on the server after deploying.

## Summary by CodeRabbit

* **Bug Fixes**
  * National calendar “move event” logic now correctly transfers the celebration from the original date to the computed new date when the liturgical event exists, including the standard “transferred from … to …” message.
  * Fixed U.S. national calendar move scenarios for 2025 and 2026 saints’ memorials.
* **Tests**
  * Added regression coverage for national calendar move behavior (U.S. 2025 and 2026), including a new St Camillus de Lellis test definition.
  * Improved calendar test cache cleanup by adding a reusable cache purge helper used by ICS-related test helpers.

### DIFF
--- a/jsondata/sourcedata/calendars/nations/US/US.json
+++ b/jsondata/sourcedata/calendars/nations/US/US.json
@@ -20,7 +20,7 @@
                 "action": "moveEvent",
                 "missal": "US_2011",
                 "reason": "National Day of Prayer for the Unborn",
-                "since_year": 1970
+                "since_year": 2011
             }
         },
         {
@@ -83,7 +83,7 @@
                 "action": "moveEvent",
                 "missal": "US_2011",
                 "reason": "Independence Day",
-                "since_year": 1970
+                "since_year": 2011
             }
         },
         {
--- a/jsondata/tests/StCamillusDeLellisTest.json
+++ b/jsondata/tests/StCamillusDeLellisTest.json
@@ -1,0 +1,203 @@
+{
+    "name": "StCamillusDeLellisTest",
+    "event_key": "StCamillusDeLellis",
+    "description": "The optional memorial of 'Saint Camillus de Lellis, Priest' should fall on July 18, moved from July 14 to make way for bl Kateri Tekakwitha",
+    "test_type": "exactCorrespondence",
+    "applies_to": {
+        "national_calendar": "US"
+    },
+    "assertions": [
+        {
+            "year": 1999,
+            "expected_value": null,
+            "assert": "eventNotExists",
+            "assertion": "The optional memorial of 'Saint Camillus de Lellis, Priest' should not exist on July 18"
+        },
+        {
+            "year": 2000,
+            "expected_value": "2000-07-18T00:00:00+00:00",
+            "assert": "eventExists AND hasExpectedDate",
+            "assertion": "The optional memorial of 'Saint Camillus de Lellis, Priest' should fall on July 18"
+        },
+        {
+            "year": 2001,
+            "expected_value": "2001-07-18T00:00:00+00:00",
+            "assert": "eventExists AND hasExpectedDate",
+            "assertion": "The optional memorial of 'Saint Camillus de Lellis, Priest' should fall on July 18"
+        },
+        {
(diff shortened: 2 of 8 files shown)


JohnRDOrazio:
@coderabbitai review

PR #718, opened 2026-07-19
Model Opus Dei as a personal prelature, not an Italian diocese

JohnRDOrazio:
## What was wrong

The `opudei_it` entry sat in Italy's diocese list under the name **"Diocesi di Lanusei"**, duplicating the real Lanusei entry (`lanuse_it`).

## Why the original fix was not enough

This PR originally just corrected the name. That left the deeper error in place: the **Prelature of the Holy Cross and Opus Dei** is a *personal prelature* with worldwide jurisdiction, not a territorial diocese of Italy. `catholic_dioceses_latin_rite` groups strictly by country — `{country_iso, country_name_english, dioceses[]}` — so there is no honest place for it there under any name.

## What this PR does now

Removes the entry from Italy's list and adds a new top-level `personal_prelatures` key, a sibling of `catholic_dioceses_latin_rite`:

```json
"personal_prelatures": [
    {
        "prelature_name": "Prelatura personale della Santa Croce e Opus Dei",
        "prelature_id": "opudei_it"
    }
]
```

The field names deliberately avoid `diocese_*`, so the structure no longer claims it is a diocese.

## Verified non-breaking

- **No schema governs `world_dioceses.json`.** Nothing in `jsondata/schemas/` references `catholic_dioceses_latin_rite`, so a new sibling key is unconstrained.
- **Both consumers ignore it.** `CalendarHandler` and `CalendarMetadataProvider` both go through `CatholicDiocesesMap::fromObject()`, which reads only `$data->catholic_dioceses_latin_rite` by name and never rejects unknown properties.
- **The map still works.** `dioceseNameFromId('IT', 'lanuse_it')` still returns `Diocesi di Lanusei`; `dioceseNameFromId('IT', 'opudei_it')` now returns `null`, which is the intent.
- **No API behaviour change.** Valid diocesan calendars are derived from the bundled calendar folders, not from this file, and there is no `opudei_it` folder. `/calendar/diocese/opudei_it/2026` returned `400` before this change and returns `400` after it.
- PHPStan level 10 clean; 732 model/service tests pass.

## Note on the id

`opudei_it` keeps its `_it` suffix. It follows the file-wide `<name>_<iso>` convention and reflects the prelature's seat in Rome; changing it would mean editing the 2935-entry `DiocesanCalendarId` enum in `CommonDef.json` for a purely cosmetic gain. The suffix is now much less misleading, since the entry no longer sits under Italy.

## Scope deliberately not taken

Nothing consumes `personal_prelatures` yet — it is inert, correctly-categorised data. Surfacing prelatures through `/calendars` would ripple into the metadata response shape and into `CalendarSelect` in both component libraries, which group dioceses under nations. That is worth doing only if and when a calendar actually exists for a prelature.

## Summary by CodeRabbit

* **Data Updates**
  * Removed the Opus Dei prelature from Italy’s dioceses list.
  * Added a dedicated personal prelatures listing for “Prelatura personale della Santa Croce e Opus Dei.”
  * Updated the Italian religious structure data to distinguish dioceses from personal prelatures.

### DIFF
--- a/jsondata/schemas/CommonDef.json
+++ b/jsondata/schemas/CommonDef.json
@@ -3517,7 +3517,6 @@
                 "novara_it",
                 "nuoror_it",
                 "opmapa_it",
-                "opudei_it",
                 "ormpli_it",
                 "oriair_it",
                 "orista_it",
--- a/jsondata/world_dioceses.json
+++ b/jsondata/world_dioceses.json
@@ -7085,10 +7085,6 @@
                     "diocese_id": "opmapa_it"
                 },
                 {
-                    "diocese_name": "Diocesi di Lanusei",
-                    "diocese_id": "opudei_it"
-                },
-                {
                     "diocese_name": "Ordinariato Militare per l'Italia",
                     "diocese_id": "ormpli_it"
                 },
@@ -13134,6 +13130,12 @@
                     "diocese_id": "mutare_zw"
                 }
             ]
+        }
+    ],
+    "personal_prelatures": [
+        {
+            "prelature_name": "Prelatura personale della Santa Croce e Opus Dei",
+            "prelature_id": "opudei_it"
         }
     ]
 }


JohnRDOrazio:
I'm not convinced by this fix, which leaves a supranational personal prelature under Italy. It should either be removed completely, or we should restructure the database to accomodate the various ecclesiastical circumscriptions

PR #737, opened 2026-07-23
Ambrosian 8a: partition sourcedata by rite (rite/{roman,ambrosian})

JohnRDOrazio:
## Summary

Milestone **8a** of the Ambrosian-rite rollout: a foundational, **behavior-preserving** reorganization of the JSON source-data tree by liturgical **rite**. This unblocks the diocesan-overlay milestones — Milano/Bergamo/Novara have *both* Ambrosian and Roman-rite parishes, so a single `dioceses/IT/milano_it/` folder can't cleanly hold two rites' calendars. Partitioning by rite at the root solves this structurally and generalizes to future rites.

```
jsondata/sourcedata/rite/
├── roman/      { calendars, decrees, lectionary, missals }   ← all Roman data, moved
└── ambrosian/  { missals/{propriumdetempore,propriumdesanctis_2024}, calendars/dioceses/ }
```

The Ambrosian comune loses its redundant `/ambrosian` segment (now implied by `rite/ambrosian/`).

## What changed (functional diff is tiny)

- **Two new base constants** (`ROMAN_RITE_FOLDER`, `AMBROSIAN_RITE_FOLDER`) + **re-basing six base-folder constants** in `src/Enum/JsonDataConstants.php`. All ~137 derived usages recompose automatically — `JsonData.php` cases are pure aliases, untouched.
- The **one trap** handled correctly: the Ambrosian bases re-base onto `AMBROSIAN_RITE_FOLDER`, *not* off `MISSALS_FOLDER` (which now lives under `rite/roman/`).
- Physical `git mv` of the folders (history-preserving; ~276 renames) + an empty `rite/ambrosian/calendars/dioceses/.gitkeep` scan-root for the next milestone.
- Fixed a **second, hidden path-composition mechanism** the plan's analysis missed: `AmbrosianMissal.php` hard-composed the pre-move `/ambrosian/propriumdesanctis_2024/…` onto `MISSALS_FOLDER`; now derives from `AMBROSIAN_SANCTORALE_FOLDER` (removes a hand-rolled literal).
- Swept every stale `jsondata/sourcedata/…` path reference in docblocks/comments — **zero** remain in `src/` or `phpunit_tests/`.

## Discovery invariant

[...]

PR #738, opened 2026-07-24
feat(ambrosian): diocesan overlays — 4 dioceses, rite-scoped discovery/validation (8b)

JohnRDOrazio:
## Ambrosian Diocesan Overlays (milestone 8b)

Adds diocesan-calendar support for the **Ambrosian rite** (which has no national layer). Four dioceses ship: **milano_it, bergam_it, novara_it** (IT) and **lugano_ch** (CH). Diocesan events overlay the *comune ambrosiano* calendar with diocesan-wins semantics by `event_key`. Builds on 8a (rite-partitioned sourcedata, #737).

Executed as 13 subagent-driven TDD tasks (implementer + reviewer each), golden-master-gated, followed by a final whole-branch review.

### What ships

- **`/calendar/ambrosian/diocese/{id}`** — full calculated diocesan calendar (comune base + diocesan overlay).
- **`/events/ambrosian/diocese/{id}`** — diocesan event catalog merged into the comune catalog.
- **`/calendars`** — announces the 4 Ambrosian dioceses, each tagged with `rite`.
- **Data:** milano_it (31 rows), bergam_it (3), novara_it (2), lugano_ch (4), each with `it_IT`/`la_VA` i18n.
- **Discovery** is rite-aware (scans both `rite/roman` and `rite/ambrosian` trees); diocese validation is **rite-scoped and data-driven** (the hardcoded `AmbrosianRiteProfile::SUPPORTED_DIOCESES` whitelist is retired).
- **Diocese→nation decoupling** so lugano_ch (no `nations/CH`) loads correctly and the IT dioceses don't wrongly pull the Roman IT national calendar.
- **Overlay** (`applyAmbrosianDiocesanCalendar`) runs before precedence resolution; overrides use remove+add (diocesan-wins), and festive diocesan events keep their `ReadingsFestive` shape through the ferial readings backfill.
- **Schema:** `DiocesanCalendar.json` gains optional `metadata.rite` (default `roman`, so existing Roman files stay valid); `LitCalMetadata.json` gains required `rite`.
- **Bonus fix:** restored 6 Roman schema-validation tests that 8a had silently left skipping (their hardcoded glob paths still pointed at the pre-move `sourcedata/calendars/…` layout — repointed to the `JsonData::*` enums), plus a new test validating all 4 Ambrosian diocesan files.

### Verification

- **Golden-master 9/9 byte-identical** — the General Roman Calendar is unchanged. The Ambrosian code paths are strictly additive and rite-guarded; the diff contains no `setlocale`/`putenv`/`gettext` mutations.
- Full suite: 1819 tests, 0 failures (the ~18 errors are environment-only — no live `localhost:8000` server / no Postgres in the sandbox, CI-gated).
- PHPStan level 10, phpcs, php-parallel-lint all clean.
- Final whole-branch review: **Ready to merge**, no Critical/Important findings.

> Golden-master must be run via `vendor/bin/phpunit --filter CalendarGoldenMaster`. Invoking the test *file* directly reorders its methods and trips the pre-existing locale issue below (5/9) — a test-harness ordering artifact, not a regression in this branch.

### Deferred follow-ups (out of scope for this PR)

1. **Pre-existing production bug (worth its own PR against the core Roman path).** `CalendarHandler`'s Latin/default branch selects its gettext catalog but never clears the `LANGUAGE` env var it sets for translated locales, nor resets `setlocale`. In a persistent worker (`PHP_CLI_SERVER_WORKERS`), a `/calendar` Latin (`la`) request following an Italian request in the same worker process can return Italian-translated names. Not introduced or worsened here (Ambrosian names come from JSON i18n, not gettext), but discovered during this work.
2. **`/data` endpoint is not rite-aware** — `GET /data?category=diocese&calendar=milano_it` resolves the diocese in the index but probes the `rite/roman` tree, giving a misleading 404. A 404→404 non-regression on an endpoint untouched by this feature.
3. **`Europe/Zurich` (and `Europe/Vienna`) missing from the `CommonDef.json` Timezone enum** — lugano_ch uses `Europe/Rome` as a stopgap (identical CET offset, so calculations are correct).
4. **`suppressed_events` metadata** lists overridden comune events, which reads like a precedence suppression. Cosmetic — the `litcal` output is correct (diocesan event wins with the right grade).
5. **Data proofing:** ~18–20 uncertain Latin names across the 4 dioceses, and the Nov 24 Andrea Dung-Lac duplicate exclusion, should be proofed against a published ordo (chiesadimilano.it) in the later ordo-validation milestone.

### Minor (follow-up OK, from the final review)

- `/events` prefixes diocesan keys and lets them coexist with the comune event, while `/calendar` uses the plain key and replaces — an intentional catalog-vs-resolved split, worth a documentation note.
- `EventsHandler` indexes the i18n array directly for names, whereas `CalendarHandler` routes through `DiocesanData::setNames()` which throws loudly on a missing key; latent-only (parity is currently perfect and schema-validated).

## Summary by CodeRabbit

- **New Features**
  - Added Ambrosian diocesan calendar support for Milan, Bergamo, Novara, and Lugano.
  - Introduced rite-aware diocesan overlays for Ambrosian calendars and event results, including localized diocese-specific naming, grade overrides, and festive empty-reading handling.
  - Added rite-aware discovery/validation across calendar, event, and `/calendars` responses, with new Ambrosian diocese routing for calendar and events.

- **Validation / Quality**
  - Updated JSON schemas to include diocesan `rite` and expanded automated schema/endpoint validation for Ambrosian diocesan behavior.


JohnRDOrazio:
Note: the pre-existing locale/LANGUAGE production bug called out under "Deferred follow-ups" #1 is now tracked separately in #739.

JohnRDOrazio:
## CI fix + CodeRabbit triage

**CI failure (the actual blocker):** the pre-existing `CalendarTest::testGetCalendarSampleAllCalendars` and `EventsTest::testGetEventsSampleAllPaths` iterate every calendar announced by `/calendars` and requested each via its plain Roman path. Now that the 4 Ambrosian dioceses are announced, those requests hit the Roman rite and are (correctly) rejected 400 by the rite-scoped validation. Fixed by making the sample-all tests rite-aware: Ambrosian dioceses are requested via `/calendar/ambrosian/diocese/...` and `/events/ambrosian/diocese/...`, calendar years below the Ambrosian floor (1976) are skipped, and `national_calendar` is asserted null for Ambrosian dioceses. (These are CI-only tests — they need a live server + Postgres — so they didn't surface in the local sandbox runs.)

### CodeRabbit findings

**Addressed:**

- **Lugano timezone → `Europe/Zurich`** — added `Europe/Zurich` to the `CommonDef.json` Timezone enum and set it on `lugano_ch` (was `Europe/Rome`; same CET/CEST offset, so calculations unchanged). `5e967d38`
- **Diocesan override recorded in `suppressed_events`** — added `LiturgicalEventCollection::removeLiturgicalEventWithoutSuppression()` and used it in `applyAmbrosianDiocesanCalendar()`, so a comune event replaced in place by a diocesan override no longer surfaces as a suppressed celebration. Roman `removeLiturgicalEvent()` untouched; added a regression assertion. `11fdb0a5`
- **`JsonDataAmbrosianDiocesanPathTest` unused `setUpBeforeClass`** — added a `->path()` runtime-contract assertion so the setup is exercised. `dc59c7ee`

**Skipped, with reason:**

- **`loadDiocesanCalendarData()` null branch → throw instead of message+return** — this is pre-existing behavior on the shared Roman code path (not introduced by this PR) and is unreachable for the 4 shipped dioceses (they resolve via `world_dioceses.json`, as the overlay tests prove). Changing it would alter Roman-diocese error semantics, which is out of this PR's scope; better as a separate robustness change.
- **Extract the duplicated diocese lookup + rite-mismatch check into a shared helper (`CalendarParams` / `EventsParams`)** — nitpick; the two blocks live in different classes with different surrounding validation flow, and a shared static helper adds cross-class coupling for ~8 lines. Not "minimal"; deferrable.
- **Memoize the `MetadataCalendars` instance in `CalendarMetadataProvider::create()`** — the `/data` PUT/PATCH/DELETE endpoints write calendar definitions, so a process-lifetime cache would serve stale metadata after a write (and complicate test isolation). The gain over a filesystem scan is marginal and not worth the correctness risk; skipped.

Golden-master 9/9, PHPStan level 10, phpcs, and parallel-lint all clean locally.

JohnRDOrazio:
The "catalog-vs-resolved split" minor follow-up (diocesan overrides re-declaring comune keys + remove/re-add, and the resulting `/events` vs `/calendar` key mismatch) is now tracked as a tidier-fix enhancement in #740: introduce an in-place diocesan "change-grade"/setProperty action instead.

Issue #739, opened 2026-07-24
CalendarHandler::prepareL10N() leaks LANGUAGE across requests — Latin calendar can return a prior request's language in persistent workers


JohnRDOrazio:
## Summary

`CalendarHandler::prepareL10N()` sets process-global locale state (`setlocale(LC_ALL, …)`, `putenv("LANGUAGE=…")`, `\Locale::setDefault(…)`) for translated locales, but its **Latin / default branch never clears that state**. Because the API runs persistent workers (`PHP_CLI_SERVER_WORKERS`, ≥6), locale state set by one request survives into the next request handled by the same worker process. As a result, a `/calendar` **Latin (`la` / `la_VA`) request that follows a translated request (e.g. `it`) in the same worker can return names in the previous request's language instead of Latin.**

## Location

`src/Handlers/CalendarHandler.php` — `prepareL10N()` (declared at line ~5013, called at line ~5493).

```php
LitLocale::$PRIMARY_LANGUAGE = $baseLocale;
if ($baseLocale !== LitLocale::LATIN_PRIMARY_LANGUAGE) {
    // ... build $localeArray ...
    $runtimeLocale = setlocale(LC_ALL, $localeArray);          // ~5039
    // ...
    putenv("LANGUAGE={$languageEnv}");                          // ~5056  <-- sets, never cleared for Latin
    \Locale::setDefault($normalizedLocale);                     // ~5059
    LitLocale::$RUNTIME_LOCALE = $normalizedLocale;
} else {
    LitLocale::$RUNTIME_LOCALE = LitLocale::LATIN_PRIMARY_LANGUAGE;   // ~5063  <-- does NOT reset setlocale / LANGUAGE / ICU default
}
// ...
bindtextdomain('litcal', ...); textdomain('litcal');           // ~5074  gettext now reads the STALE LANGUAGE
```

glibc `gettext()` selects its message catalog primarily from the `LANGUAGE` environment variable (above `LC_MESSAGES`). For a Latin request, `prepareL10N()` leaves `LANGUAGE` at whatever a prior translated request set it to, so `gettext()` returns that prior language's translations rather than falling through to the untranslated (Latin) `msgid`.

## Impact

- **Correctness bug, not just cosmetic.** Latin (`la` / `la_VA`) calendar output can contain another language's saint/feast names depending on which request the worker handled immediately before. This affects the General Roman Calendar and the Vatican calendar (Latin by nature).
[...]

PR #744, opened 2026-07-24
fix(events): manage LANGUAGE in EventsHandler::setLocale to stop locale leak (#743)

### LINKED ISSUE #743: EventsHandler::setLocale() doesn't manage LANGUAGE — /events localized fields can leak a prior request's language (sibling of #739)
## Summary

`EventsHandler::setLocale()` sets the process locale for a request but does not manage the `LANGUAGE` environment variable the way `CalendarHandler::prepareL10N()` now does (after #742 / #739). Because glibc `gettext()` selects its message catalog from `LANGUAGE` **above** `LC_MESSAGES`, a `LANGUAGE` value left by a prior request in the same persistent worker leaks into a later `/events` response's localized fields — the `/events` analog of #739.

## Location

`src/Handlers/EventsHandler.php` — `setLocale()` (declared line ~313, called line ~871):

```php
private function setLocale(): void
{
    $localeArray = [ /* request locale + region variants */ ];
    setlocale(LC_ALL, $localeArray);          // return value not checked
    bindtextdomain('litcal', Router::$apiFilePath . 'i18n');
    textdomain('litcal');
    LiturgicalEventAbstract::setLocale($this->EventsParams->Locale);
}
```

Compared with the (now-fixed) `CalendarHandler::prepareL10N()`, this method:

1. **Never sets `putenv("LANGUAGE=…")`** to pin the *current* request's catalog. So if any earlier request in the worker set `LANGUAGE` — e.g. `CalendarHandler::prepareL10N()` does `putenv("LANGUAGE=it_IT:…")` for a translated `/calendar` request — it persists, and gettext honors it over the `LC_MESSAGES` that `setLocale()` just set.
2. **Doesn't check `setlocale()`'s return value.** If the requested locale isn't installed on the host, `setlocale()` returns `false` and the process locale is left at whatever the previous request set.
3. **Has no Latin/reset branch.** The Latin locale (`la`/`la_VA`) is generally not installed, so `setlocale()` fails for Latin and the process stays on the prior locale.

## Impact

The `/events` catalog's localized fields go through gettext (`LiturgicalEventAbstract`: `common_lcl` via `LitCommons::fullTranslate()` / `_('or')` / `_('From the Common')`, `grade_lcl` via `LitGrade::i18n()`). A concrete leak:

1. Worker handles `GET /calendar/nation/IT/{year}` (Italian) → `prepareL10N()` sets `LANGUAGE=it_IT:…`.
2. Same worker then handles `GET /events/nation/US` with `Accept-Language: fr` → `setLocale()` sets `setlocale(LC_ALL, fr_FR)` but leaves `LANGUAGE=it_IT`. gettext returns **Italian** for the French request's `common_lcl` / grade labels.

Note the Latin `/events` output is largely shielded (those helpers branch on an explicit `$locale === LATIN` and use hardcoded Latin rather than gettext), so the most exposed case is **translated-after-translated in a different language**, and translated-after-a-`LANGUAGE`-setting-`/calendar`-request.

Also relevant: `/events` does **not** appear to have the `ServerCache` short-circuit that `/calendar` has, so — unlike #739, which is masked until a cache miss — this can surface on ordinary consecutive requests.

## Suggested fix

Mirror `CalendarHandler::prepareL10N()` (as fixed in #742): make `setLocale()` deterministic per request —

- For a translated locale: check `setlocale()`'s result, and `putenv("LANGUAGE=…")` (+ optionally `\Locale::setDefault`) so the current request's catalog is pinned regardless of prior state.
- For Latin / the default: reset `setlocale(LC_ALL, 'C')` and `putenv('LANGUAGE')` (unset) so localized output falls through deterministically.

Consider extracting the shared logic so `CalendarHandler` and `EventsHandler` don't drift (they already duplicate the `$localeArray` construction and the `LANGUAGE` handling).

## Acceptance criteria

- A translated `/events` request returns its own language regardless of a prior request (in the same process) in a different language or one that set `LANGUAGE`.
- A regression test analogous to `CalendarHandlerLocaleLeakTest` (issue a translated request, then an `/events` request in a different language in the same process, assert the localized fields are correct).
- No change to single-request behavior; PHPStan / phpcs / golden-master unaffected.

## Context

Sibling of #739 (fixed for `CalendarHandler` in #742). Discovered while fixing #739.

JohnRDOrazio:
## Summary

Fixes #743 — the `/events` analog of #739. `EventsHandler::setLocale()` set the process locale per request but never managed the `LANGUAGE` env var. Because glibc `gettext()` reads `LANGUAGE` **above** `LC_MESSAGES`, a value left by a prior request in the same persistent worker (e.g. `CalendarHandler::prepareL10N()` sets `LANGUAGE=it_IT:…` for a translated `/calendar` request) leaked into a later `/events` response's gettext-backed localized fields (`grade_lcl`, `common_lcl`). Unlike `/calendar`, `/events` has **no `ServerCache` short-circuit**, so this surfaces on ordinary consecutive requests.

## Fix

`setLocale()` is now deterministic per request, mirroring the fixed `prepareL10N()`:

- **Translated locale with an installed system locale** → pin the current request's gettext catalog via `putenv("LANGUAGE=…")` and update ICU's default (`\Locale::setDefault`).
- **`setlocale()` fails, or Latin/default** → reset `setlocale(LC_ALL, 'C')`, clear `LANGUAGE`, reset ICU so localized output falls through to the untranslated (English) msgid base.

### Deliberate divergence from `prepareL10N()`

A failed `setlocale()` **degrades gracefully instead of throwing 503**, because:

1. **English is the source language** — no catalog of its own, and on many hosts bare `en` is not an installed system locale (only `en_US.utf8` etc.), so `setlocale()` returns `false`.
2. **`/events` has no cache**, so `setLocale()` runs on every request — a strict throw would 503 every English `/events` request.

The reset-on-failure path still clears the leak vector, so an English (or any uninstalled-locale) request no longer inherits a prior request's `LANGUAGE`. Adopting the strict throw first surfaced this — it broke existing `EventsHandlerRiteRoutingTest` cases with `503 instead of 200`.

## Tests

Adds `EventsHandlerLocaleLeakTest` (analogous to `CalendarHandlerLocaleLeakTest`), each case verified **red → green**:

- After a translated Italian `/calendar` request pollutes the worker, a **French** `/events` request returns French (not the leaked Italian).
- Same for an **English** `/events` request (covers the `setlocale`-failure path).
- A **Latin** `/events` request clears the leaked `LANGUAGE` env var.

## Verification

- New leak tests: 3/3 (fail on pre-fix code).
- `EventsHandlerTest` + `EventsHandlerRiteRoutingTest`: pass (previously-failing rite tests recovered).
- Full suite: **0 failures** (remaining errors are pre-existing, live-server/DB-dependent integration tests).
- PHPStan level 10 clean · phpcs clean · php-parallel-lint clean.

## Follow-up (deferred)

Extracting the shared `$localeArray` + `LANGUAGE` handling so `CalendarHandler` and `EventsHandler` don't drift is intentionally **not** in this PR — the two build `$localeArray` differently, #742 (the `prepareL10N()` Latin-reset) isn't merged to `development` yet, and the two now intentionally diverge on `setlocale()`-failure handling. Recommend a dedicated follow-up once #742 lands, standardizing on graceful degradation (which would also fix the latent `en`-on-cache-miss 503 in `CalendarHandler`). Detailed in the [issue comment](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/743#issuecomment-5073180844).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed language settings persisting between requests in long-running workers.
  * Ensured translated event responses consistently honor the requested language after requests in other languages.
  * Prevented Latin-language requests from inheriting translations from previous requests.
  * Improved fallback behavior when a requested locale cannot be initialized.

* **Tests**
  * Added regression coverage for sequential Italian, French, English, and Latin requests.

### DIFF
--- a/phpunit_tests/Handlers/EventsHandlerLocaleLeakTest.php
+++ b/phpunit_tests/Handlers/EventsHandlerLocaleLeakTest.php
@@ -1,0 +1,136 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests\Handlers;
+
+use LiturgicalCalendar\Api\Handlers\CalendarHandler;
+use LiturgicalCalendar\Api\Handlers\EventsHandler;
+use LiturgicalCalendar\Api\Http\Enum\ReturnTypeParam;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Regression for issue #743: EventsHandler::setLocale() leaks the LANGUAGE env var
+ * across requests. glibc gettext() reads LANGUAGE above LC_MESSAGES, so a LANGUAGE
+ * value left by a prior request in the same persistent worker (e.g. a translated
+ * /calendar request, whose CalendarHandler::prepareL10N() sets it) would leak into a
+ * later /events response's gettext-backed localized fields (grade_lcl, common_lcl) —
+ * the /events analog of #739. Unlike /calendar, /events has no ServerCache
+ * short-circuit, so this surfaces on ordinary consecutive requests.
+ */
+final class EventsHandlerLocaleLeakTest extends AbstractHandlerTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+        // Force Router::isLocalhost() true so CalendarHandler::handle() bypasses the
+        // response cache and actually runs prepareL10N() (which sets LANGUAGE) — the
+        // pollution the leak requires.
+        $_SERVER['SERVER_NAME'] = 'localhost';
+        // Start each test from a clean process-global locale so the French baseline
+        // below is genuinely French regardless of what earlier tests in this process
+        // left behind.
+        setlocale(LC_ALL, 'C');
+        putenv('LANGUAGE');
+    }
+
+    protected function tearDown(): void
+    {
+        unset($_SERVER['SERVER_NAME']);
+        // Never let this test's deliberate locale pollution leak into later tests.
+        setlocale(LC_ALL, 'C');
+        putenv('LANGUAGE');
+        parent::tearDown();
+    }
+
+    /**
+     * Map each event to its gettext-backed localized fields — the fields that leak.
+     *
+     * @return array<string,array{grade_lcl:?string,common_lcl:?string}>
+     */
+    private function localizedFields(ResponseInterface $response): array
+    {
+        $events = $this->decodeJsonBody($response)['litcal_events'];
+        self::assertIsArray($events);
+        self::assertNotEmpty($events);
+
+        $out = [];
+        foreach ($events as $event) {
+            self::assertIsArray($event);
+            $out[$event['event_key']] = [
+                'grade_lcl'  => $event['grade_lcl'] ?? null,
+                'common_lcl' => $event['common_lcl'] ?? null,
+            ];
+        }
+        return $out;
+    }
+
+    public function testFrenchEventsKeepTheirLanguageAfterAnItalianRequest(): void
+    {
+        // Baseline: a French /events request with no prior pollution → French fields.
+        $frBaseline = $this->localizedFields(
+            ( new EventsHandler() )->handle($this->requestFor('GET', '/events', ['Accept-Language' => 'fr']))
+        );
+
+        // Pollute the worker: a translated (Italian) /calendar request leaves
+        // LANGUAGE=it_IT:... via CalendarHandler::prepareL10N().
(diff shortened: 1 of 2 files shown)


JohnRDOrazio:
Deferred shared-locale-helper follow-up is now tracked in #745 (blocked on #742 landing first).

PR #770, opened 2026-08-11
Echo the rite in /calendar and /events settings

### LINKED ISSUE #760: Echo the rite in the /calendar response
## Summary

A `/calendar` response carries no indication of the rite it was computed under. Clients that need to
label or branch on the rite have to track it themselves, out of band, and can get it wrong.

## Why it matters

The **rite-level** Ambrosian calendar is indistinguishable from the General Roman calendar by payload
alone:

- `settings` has neither `national_calendar` nor `diocesan_calendar` (correctly — it is neither)
- `metadata` carries no rite field

```console
$ curl -s 'http://localhost:8000/calendar/ambrosian' | jq '.settings | keys'
["ascension","corpus_christi","epiphany","eternal_high_priest","holydays_of_obligation","locale","return_type","year","year_type"]
```

A diocesan or national calendar is self-describing — the response names the calendar — but
`/calendar/ambrosian` and `/calendar` differ only in their contents.

Found while making `WebCalendar` in `liturgy-components-js` caption an Ambrosian calendar correctly
(Liturgical-Calendar/liturgy-components-js#5). Because the payload could not answer "which rite is this?",
the component had to take the rite from the `ApiClient` that issued the request and thread it through to
render time. That works, but it means:

- every consumer that renders a calendar has to plumb the rite alongside the data
- the pairing is only as reliable as the client's own bookkeeping. With two requests in flight after a
  rite change, a response can be paired with the wrong rite unless the client pins the rite per request
- a consumer that receives calendar data from anywhere other than its own request — a cache, a message,
  a server-rendered payload — has no way to recover the rite at all

## Suggestion

Echo the rite in the response, in whichever of the two places fits the existing conventions better:

```jsonc
// in settings, alongside the other request parameters that were applied
"settings": { "year": 2026, "locale": "la", "rite": "roman", ... }

// or in metadata, alongside the other descriptive fields
"metadata": { "version": "...", "rite": "ambrosian", ... }
```

`settings` seems the closer match: the rite is a request parameter (a path segment) that the API resolves
and applies, like `year_type` and `locale`, both of which are already echoed there.

Emitting it for **every** rite, including `roman`, would make it usable without a presence check. Since
`Router::extractRiteSegment()` already accepts `roman` explicitly and defaults to it, the value is always
known.

## Compatibility

Additive. Existing consumers ignore unknown fields, and no current response shape changes meaning — the
absence of a rite today is not information anyone can be relying on, since it is absent for every rite.

JohnRDOrazio:
Closes #760.

## What

`settings.rite` is now echoed on both `/calendar` and `/events`, for **every** rite —
`roman` included — so consumers can read it without a presence check.

```console
$ curl -s 'http://localhost:8000/calendar/ambrosian' | jq '.settings.rite'
"ambrosian"

$ curl -s 'http://localhost:8000/calendar' | jq '.settings.rite'
"roman"

$ curl -s 'http://localhost:8000/events/ambrosian' | jq '.settings'
{ "locale": "la_VA", "national_calendar": null, "diocesan_calendar": null, "rite": "ambrosian" }
```

`Router::extractRiteSegment()` already accepts `roman` explicitly and defaults to it, so the
value is always known.

## Why `/events` too

The issue is scoped to `/calendar`, but `/events` echoes the same `settings` block and has the
identical blind spot: `/events/ambrosian` sets neither `national_calendar` nor
`diocesan_calendar`, so it was indistinguishable from `/events`. Covering both keeps the two
read endpoints consistent.

## Placement

`settings`, per the issue's reasoning — the rite is a request parameter (a path segment) that
the API resolves and applies, like `year_type` and `locale`, both already echoed there.

## Changes

| File | Change |
|------|--------|
| `src/Handlers/CalendarHandler.php`      | emit `settings->rite` from `CalendarParams->Rite` |
| `src/Handlers/EventsHandler.php`        | emit `settings['rite']` from `EventsParams->Rite`  |
| `jsondata/schemas/CommonDef.json`       | new shared `Rite` definition                       |
| `jsondata/schemas/LitCal.json`          | `rite` added + required in all three settings variants |
| `jsondata/schemas/LitCalEventsPath.json`| `rite` added + required in all three `oneOf` branches  |
| `jsondata/schemas/LiturgicalCalendar.xsd` | `<Rite>` in the Settings sequence                |

`openapi.json` needs no change — its `/calendar` and `/events` responses `$ref` the schema files,
which are authoritative.

### XML placement

`<Rite>` sits between `HolydaysOfObligation` and the optional
`NationalCalendar`/`DiocesanCalendar` elements, keeping the mandatory elements contiguous.
`Utilities::transformKey()` maps `rite` → `Rite` with no special-casing.

### Golden-master fixtures

Regenerated. The diff across all nine fixtures is exclusively the new field:

```console
$ git diff -U0 -- phpunit_tests/fixtures/golden-master/ | grep -E '^[+-]' | grep -v '^[+-][+-][+-]' | sort | uniq -c
      6 -        }
      6 +        },
      6 +        "rite": "roman"
      3 +        "rite": "roman",
```

I confirmed the regression check is meaningful rather than trivially green: with the fixtures
reset to `HEAD` and the handler change applied, `CalendarGoldenMasterTest` fails 9/9 on exactly
this field.

## Compatibility

Additive. Existing consumers ignore unknown fields, and the absence of a rite today is not
information anyone can rely on, since it is absent for every rite.

One caveat for XML consumers: the XSD `Settings` sequence is ordered and now requires `Rite`, so
a client validating against the **new** XSD will reject an **old** response. That only matters
during the window between deploying the schema and deploying the code.

## Verification

- `vendor/bin/phpunit --exclude-group slow` — 2068 tests green, against a server built from this
  branch. (Against the default `:8000` server, which runs the main checkout, `CalendarTest::testGetCalendarReturnsXML`
  fails on the old payload vs. the new XSD — expected. The one remaining failure when pointing the
  suite at a worktree port is `ExecuteValidationTest`, the documented worktree-port artifact where
  the Dockerised WebSocket server cannot reach the host port.)
- `composer analyse` (PHPStan level 10) — no errors
- `composer lint` — clean
- `composer lint:openapi` — valid
- New `phpunit_tests/Handlers/SettingsRiteEchoTest.php` — 14 tests covering both handlers across
  general/national/diocesan/ambrosian-comune/ambrosian-diocesan routes, plus the XML element and
  the "identifiable without calendar keys" case that motivated the issue.
- Ad-hoc check (not committed): the generated XML validates against the updated XSD for all three
  Roman routes.

## Two pre-existing issues found along the way

Neither is touched here; both predate this change and deserve their own issues if you agree.

1. **Ambrosian XML has never validated against the XSD.** The `LitColor` enumeration in
   `LiturgicalCalendar.xsd` lists only `green|red|white|purple|rose`, but Ambrosian events use
   `morello`, so `/calendar/ambrosian?return_type=XML` fails schema validation on the colour
   `<Option>` — independently of this PR.

2. **The golden-master generator can mask its own regression check.**
   `CalendarGoldenMasterGenerateTest` writes the fixtures and sorts before
   `CalendarGoldenMasterTest` in a default run, so a full-suite run rewrites the baseline before
   the check reads it. It is in group `golden-master-generate`, but that group is not excluded by
   default.

Also worth flagging at deploy time, not a code issue: `computeEngineCacheDataVersion()` hashes only
`jsondata/sourcedata` and `i18n/*.mo`, never the PHP source. This is a code-only change to the
serialiser, so a deployed instance will keep serving cached rite-less calendars until `engineCache/`
is cleared (or `API_VERSION` bumps).

## Summary by CodeRabbit

* **New Features**
  * Calendar and events configurations now require a liturgical rite: `roman` or `ambrosian`.
  * JSON responses include the selected rite in their settings.
  * XML responses now include the corresponding `<Rite>` element.
  * Ambrosian calendars can be identified correctly, including comune routes without additional calendar keys.

* **Documentation**
  * Configuration examples and supported calendar formats now document the required rite setting.

### DIFF
--- a/jsondata/schemas/CommonDef.json
+++ b/jsondata/schemas/CommonDef.json
@@ -407,6 +407,15 @@
                 "mobile"
             ]
         },
+        "Rite": {
+            "type": "string",
+            "title": "Rite",
+            "description": "the liturgical rite the calendar was computed under, as selected by the optional leading rite path segment (`roman` when the segment is absent)",
+            "enum": [
+                "roman",
+                "ambrosian"
+            ]
+        },
         "Epiphany": {
             "type": "string",
             "title": "Epiphany",
--- a/jsondata/schemas/LitCal.json
+++ b/jsondata/schemas/LitCal.json
@@ -196,7 +196,8 @@
                     "StJoseph": false,
                     "StsPeterPaulAp": false,
                     "AllSaints": true
-                }
+                },
+                "rite": "roman"
             },
             "metadata": {
                 "version": "5.3",
@@ -896,6 +897,9 @@
                 },
                 "holydays_of_obligation": {
                     "$ref": "./CommonDef.json#/definitions/HolyDaysOfObligation"
+                },
+                "rite": {
+                    "$ref": "./CommonDef.json#/definitions/Rite"
                 }
             },
             "required": [
@@ -907,7 +911,8 @@
                 "return_type",
                 "year_type",
                 "eternal_high_priest",
-                "holydays_of_obligation"
+                "holydays_of_obligation",
+                "rite"
             ]
         },
         "NationalCalendarSettings": {
@@ -953,6 +958,9 @@
                 "holydays_of_obligation": {
                     "$ref": "./CommonDef.json#/definitions/HolyDaysOfObligation"
                 },
+                "rite": {
+                    "$ref": "./CommonDef.json#/definitions/Rite"
+                },
                 "national_calendar": {
                     "$ref": "./CommonDef.json#/definitions/Nation"
                 }
@@ -967,7 +975,8 @@
                 "year_type",
                 "eternal_high_priest",
                 "national_calendar",
-                "holydays_of_obligation"
+                "holydays_of_obligation",
+                "rite"
             ]
         },
         "DiocesanCalendarSettings": {
@@ -1013,6 +1022,9 @@
                 "holydays_of_obligation": {
                     "$ref": "./CommonDef.json#/definitions/HolyDaysOfObligation"
                 },
+                "rite": {
+                    "$ref": "./CommonDef.json#/definitions/Rite"
+                },
                 "national_calendar": {
                     "$ref": "./CommonDef.json#/definitions/Nation"
                 },
@@ -1030,6 +1042,7 @@
                 "year_type",
                 "eternal_high_priest",
                 "holydays_of_obligation",
+                "rite",
                 "national_calendar",
                 "diocesan_calendar"
             ]
--- a/jsondata/schemas/LitCalEventsPath.json
+++ b/jsondata/schemas/LitCalEventsPath.json
@@ -22,10 +22,13 @@
                   "diocesan_calendar": {
                     "$ref": "./CommonDef.json#/definitions/DiocesanCalendarId",
                     "description": "indicates the Diocesan Calendar used to produce the data in `litcal_events`"
+                  },
+                  "rite": {
+                    "$ref": "./CommonDef.json#/definitions/Rite"
                   }
                 },
                 "additionalProperties": false,
-                "required": [ "locale", "national_calendar", "diocesan_calendar" ]
+                "required": [ "locale", "national_calendar", "diocesan_calendar", "rite" ]
               },
               {
                 "properties": {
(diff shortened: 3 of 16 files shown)


JohnRDOrazio:
The two pre-existing findings in the description are now filed:

- #771 — design: should liturgical-colour validity be rite-scoped, and where? (recommendation: keep one `LitColor` enum, add `RiteProfile::colors()` + rite-scoped source-data validation)
- #772 — bug: Ambrosian XML fails XSD validation (`morello` missing from the enumeration), plus the Ambrosian `AllSouls` row tagged `purple` instead of `morello`

Both are independent of this PR, which stays scoped to echoing the rite.

JohnRDOrazio:
@coderabbitai review

PR #778, opened 2026-08-14
fix(schemas,ambrosian): make Ambrosian calendars validate as XML; AllSouls is morello

JohnRDOrazio:
## Ambrosian XML never validated, and `AllSouls` was the wrong colour

Fixes #772. Two independent colour defects, both surfaced from #770.

### 1. Ambrosian XML has never validated against the XSD

`/calendar/ambrosian` served as XML failed `LiturgicalCalendar.xsd` validation outright — and nothing noticed, because no test had ever validated Ambrosian XML. The pre-existing coverage (`Routes/Readonly/CalendarTest::testGetCalendarReturnsXML`) exercises only the Roman `/calendar`, and does it over HTTP against a running server, which in this project is bound to a different checkout.

Working TDD from a failing Ambrosian-XML test surfaced **three** pieces of accumulated drift, not one:

| Drift | Fix |
|-------|-----|
| `ColorEnumType` restricted to the five Roman colours, so every `morello` failed the enumeration facet | Added `morello` + `black` — `LitColor` and `CommonDef.json` already had them |
| `IsAliturgical` / `IsDominical` undeclared, rejected as unexpected elements | Declared as optional booleans — already optional in `LitCal.json` |
| `openapi.json`'s copy of the XML `Color/option` enum had drifted identically | Same two values added |

The XSD enumeration is deliberately the **union** across rites. XML Schema cannot key a facet off a sibling element, so rite-scoping colour validity belongs in the validators and is tracked separately in #771; the intent is documented inline.

### 2. Ambrosian `AllSouls` was `purple` rather than `morello`

The Commemoration of All the Faithful Departed (2 November) was authored with the Roman `purple`, while all fifteen other violet-family Ambrosian rows already used `morello`. Per *Ordinamento Generale del Messale Ambrosiano* n. 320 the proper colour of this celebration is `morello`; `black` is admitted only as an *optional* alternative on ferial days and is **excluded on Sundays** (and on vigil Masses that open a Sunday). `morello` is therefore the correct unconditional primary colour.

Modelling the conditional `black` alternative is **not** attempted here — the colour field is scalar-per-event, and expressing "primary `morello`, optional `black` when the day is neither a Sunday nor opens one" is a schema change. Filed as a follow-up.

`AllSouls` was the only Roman-only colour anywhere in the Ambrosian tree: the full census is `white` ×263, `red` ×86, `morello` ×15→16, `purple` ×1→0, `rose` ×0. That makes the rite-scoped assertion planned in #771 green.

[...]

Issue #771, opened 2026-08-11
Rite-scoped liturgical colours: one LitColor enum, or per-rite validity?


JohnRDOrazio:
## Summary

`LitColor` is a single flat enum of 7 values shared by both rites. Nothing states which colours are
licit in which rite, so a Roman-only colour can land in Ambrosian data (or vice versa) with nothing
to catch it.

## Evidence this is not theoretical

Colour usage in the rite-partitioned source-data trees:

| | white | red | green | purple | rose | morello | black |
|---|---|---|---|---|---|---|---|
| `sourcedata/rite/roman/`     | 385 | 105 | 32 | 15 | 2 | — | — |
| `sourcedata/rite/ambrosian/` | 263 | 86  | —* | **1** | — | 15 | — |

\* Ambrosian green is synthesised in `AmbrosianTemporale`, not carried in data.

The lone Ambrosian `purple` is tracked separately in the companion issue — the point here is that
the type system and the schemas both permitted it.

Two observations from the same table:

- `morello` and `black` are Ambrosian-only; `rose` and (apparently) `purple` are Roman-only.
- `LitColor::BLACK` is referenced by **no data in either tree**. It was added in Ambrosian Plan 3
  ahead of data that never materialised.

## Recommendation (for discussion)

**Keep one `LitColor` enum.** Splitting the PHP type is the wrong lever:

- PHP enums cannot extend anything, so there is no `AmbrosianLitColor extends LitColor`. The
  alternative is two unrelated enums plus a marker interface, which turns `LiturgicalEvent::$color`
  (typed `LitColor[]`) into a union across the whole model layer, serialisation and `i18n()`, and
  costs `match` exhaustiveness plus `EnumToArrayTrait`.
- The enum's job — "a liturgical colour this API can translate and render" — genuinely is
  rite-independent. *Which colours are licit in a given rite* is a validation rule, not a type.

Put the rite-scoping on the seam that already exists, in descending order of payoff:

1. **`RiteProfile::colors(): LitColor[]`.** `src/Models/Calendar/Rite/RiteProfile.php` already
   dispatches `temporaleEngine()`, `precedenceResolver()` and `missalResolver()` exactly this way.
   One authoritative statement per rite of the Missal's colour list (Ambrosian: praenotanda §350),
   consumable by the `/data` write path — where `metadata.rite` is known — and by data-integrity
   tests. This is the piece that would have caught the stray row.

[...]

PR #783, opened 2026-08-14
feat(ambrosian): conditional black for Masses for the Dead (OGMA n. 320)

### LINKED ISSUE #781: Model the conditional `black` alternative for Ambrosian Masses for the Dead (OGMA n. 320)
## Summary

The `color` field on a liturgical event is **scalar-per-event**: an unconditional array of colours. That cannot express a colour whose admissibility depends on the day the event falls on — which the Ambrosian rite requires for Masses for the dead.

Surfaced while fixing #772 (`AllSouls` was tagged `purple`; corrected to `morello` in #778) and deliberately scoped out of both #772 and #771, since it needs a schema change rather than a data or validation fix.

## The rule

Per *Ordinamento Generale del Messale Ambrosiano* n. 320:

- **`morello`** is used in Advent (excluding the Sunday of the Incarnation and any pre-Christmas Sunday), in Lent up to but excluding the Saturday *in Traditione symboli*, and in votive Masses for the forgiveness of sins; it *may* be used in offices and Masses for the dead.
- **`black`** may be used on the ferias of Lent (Saturday excluded) and in offices and Masses for the dead — **except on Sundays**.

So for the Commemoration of All the Faithful Departed (2 November):

- **on a ferial day** → `morello`, or `black` *ad libitum*;
- **on a Sunday** (as in 2025) → `morello` only; `black` is excluded.

The vigil case matters too. The Milan Curia specified for Saturday 2 November 2019 that the evening vigil Masses were of the Commemoration of All the Faithful Departed, with Credo, in `morello` and **not** `black` — consistent with n. 320's Sunday exclusion, since Vespers already opens the Sunday. The condition is therefore *"the day is not a Sunday **and** does not open a Sunday"*, not simply "not a Sunday".

**Historical note:** the `black` faculty is a recent *reintroduction*, not an unbroken survival. After the 2008 reform of the Ambrosian Lectionary, Card. Tettamanzi restored the optional use of black in substitution for morello on the ferias of Lent (Holy Week excluded), at funerals, in Masses for the dead, and on the Commemoration of All the Faithful Departed. (In the Ambrosian *vetus ordo* the *Commemoratio Omnium Fidelium Defunctorum* takes black with no alternative, with its own three formularies.)

## What this needs

> **Superseded shape — see the decisions in the comments.** An earlier draft of this section proposed a separate `color_alternative` field on the response. That is *not* the design: `color` is already an array of alternatives (`["rose", "purple"]` for Gaudete/Laetare is exactly an *ad libitum* case), so the engine appends the permitted alternative to `color` and **no response-schema change is needed**.

Resolved response, per computed date:

```jsonc
{ "color": ["morello", "black"] }   // 2 November on a feria — faculty applies
{ "color": ["morello"] }            // 2 November on a Sunday — faculty excluded
```

The work is therefore confined to the **source-data** side, which must carry the condition rather than an unconditional two-entry array:

```jsonc
{ "event_key": "AllSouls", "color": ["morello"],
  "color_ad_libitum": [ { "color": "black", "when": "not_sunday_and_not_sunday_vigil" } ] }
```

Touches `PropriumDeSanctisEvent`, `CommonDef.json` / `PropriumDeSanctis.json`, and the engine's colour-resolution step. `LitCal.json`, `LiturgicalCalendar.xsd` and `openapi.json` are **not** affected — they already type `color` as an array of `LitColor`, and #778 already added `black` to the XSD and OpenAPI enumerations.

**Compatibility:** consumers already handle multi-entry `color` arrays (71 rows ship today), so appending is the conservative option. The proper colour leads; the *ad libitum* one is appended.

## Interactions

- **#771** (rite-scoped colour validity, PR #780) — `RiteProfile::colors()` already lists `black` as Ambrosian-admissible precisely so this work will not trip the validation guard, even though no unconditional row uses it today. A conditional-colour field will need the same rite-scoped check applied to `color_alternative.colors`.
- **#772 / #778** — added `black` to the XSD and `openapi.json` colour enumerations (the union across rites), so the vocabulary is already serialisable; only the *shape* is missing.
- The grade/colour flags in `propriumdesanctis_2024` remain provisional pending the ordo-proofing pass, so this is a good moment to settle the shape before those values are finalised.

## Decisions

All settled — see comments for the reasoning.

1. **Closed enum of named predicates**, evaluated by the engine (not a general expression language). Initial vocabulary: `not_sunday_and_not_sunday_vigil` (Masses for the dead) and `lenten_feria_excluding_saturday` (ferial-Lent faculty — check whether the Holy Week exclusion needs to be carried in the name).
2. **The API resolves per-year**, keeping liturgical law in the API rather than copying it into every client.
3. **The Roman rite skips `black` entirely** — never used in practice. Already the case: `RomanRiteProfile::colors()` and `CommonDef.json#/definitions/RomanLitColor` both omit it (#771 / PR #780), so no change is required.
4. **The resolved alternative is appended to `color`**, not carried in a parallel field.

JohnRDOrazio:
## Conditional `black` for Ambrosian Masses for the Dead

Closes #781, implementing the four decisions taken on the issue.

`color` is an unconditional array, so it cannot express a colour whose admissibility depends on the day — which the Ambrosian rite requires. Per *Ordinamento Generale del Messale Ambrosiano* n. 320, `black` is admitted in offices and Masses for the dead only *ad libitum*, and **never on a Sunday**. The Milan Curia extended the same exclusion to the vigil, ruling for Saturday 2 November 2019 that the evening Masses of All Souls were `morello` and not black, since Vespers already opens the Sunday.

### What the API now returns

```jsonc
// 2 November 2026 (Monday) — faculty applies
{ "event_key": "AllSouls", "color": ["morello", "black"] }

// 2 November 2025 (Sunday) — n. 320 excludes black
{ "event_key": "AllSouls", "color": ["morello"] }

// 2 November 2024 (Saturday) — the evening Mass opens the Sunday
{ "event_key": "AllSouls", "color": ["morello"] }
```

### The four decisions, as implemented

| Decision | Implementation |
|---|---|
| Closed enum of named predicates, engine-evaluated | `AdLibitumColorCondition`, a plain `enum` in JSON Schema |
| API resolves per-year | `resolveAdLibitumColors()` runs after `setDate()` in the sanctorale assembly |
| Roman rite skips `black` | Already true since #771 — no change needed |
| Append to `color`, no parallel field | `["rose", "purple"]` for Gaudete/Laetare is the same pattern |

**The response schemas need no change at all.** `LitCal.json`, `LiturgicalCalendar.xsd` and `openapi.json` already type `color` as an array of `LitColor`, and #778 already added `black` to the XSD and OpenAPI enumerations. `color_lcl` tracks the resolved set for free, because `LiturgicalEvent::__construct()` derives it from whatever `color` it is handed. The proper colour keeps position 0 — existing multi-colour rows are inconsistent about ordering (`white,red` ×62 vs `red,white` ×6), so this is pinned by test rather than inferred.

The conditional is **source-data only**: `color_ad_libitum` on `PropriumDeSanctis`, resolved into `color` before the `LiturgicalEvent` is built. The response carries the outcome and never the condition.

Only one predicate is defined. The ferial-Lent faculty named in the issue is deferred until a data row needs it: it additionally requires season awareness, which the sanctorale resolution point does not have. Adding a member to the enum and its `match` is the whole cost of the next one.

### Rite-scoped validation reaches it too

#771's colour guard keys on `color` being an **array**; inside a `color_ad_libitum` entry it is a bare **string**, so an illicit conditional colour would have slipped through all three checkers. The payload walk in `RegionalDataHandler`, `RiteSourceColorTest` and `SchemaValidationTest` now normalises both shapes. Verified load-bearing — planting `purple` as the ad libitum colour:

```console
Swaggest\JsonSchema\Exception\EnumException: Enum failed,
  enum: ["white","red","green","morello","black"], data: "purple"
ambrosian source data must not use colors licit only in the other rite (found: purple)
```

`color_ad_libitum` is deliberately **not** added to the `/data`-writable schemas: the engine resolves it only in the Ambrosian sanctorale assembly, so permitting it on diocesan or national payloads would promise behaviour the calculation does not implement. `RegionalDataHandlerTest` pins that boundary — such a payload is rejected as an unknown property.

### A test-isolation trap worth knowing about

`AmbrosianAdLibitumColorTest` resets `LiturgicalEvent::$internal_index` before each build. That looks like housekeeping but is load-bearing:

`engineCache/` keys on `md5(serialize(CalendarParams))` and stores the **serialized response, `event_idx` values included**. `event_idx` comes from a process-lifetime static counter, and `LitCal.json` caps it at 2000 on the assumption of a fresh process per request — true in production, false inside one PHPUnit process. Building nine calendars pushed the counter past the cap, and the cached 2025 response was then handed to `CalendarHandlerAmbrosianResponseSchemaTest`, **defeating that test's own reset** and failing its schema assertion with `event_idx 5386`. Resetting before each build keeps what lands in the shared cache within the documented bounds.

### Modelling note (the one judgment call)

The API reports colours per **day**, not per Mass, so Saturday is excluded outright. Strictly, n. 320 would still permit black at a Saturday *morning* Mass for the dead — only the evening Mass falls under the Sunday exclusion. A single day-level answer has to cover both, and this takes the restrictive reading, which is the one the Curia published for the case it actually ruled on. Distinguishing them would require per-Mass colours, which the model does not have.

### Verification

| Check | Result |
|-------|--------|
| `composer test:quick` | 2232 tests, 172587 assertions, 11 skipped, **0 failures** (run twice: cold and warm `engineCache/`) |
| `composer analyse` (PHPStan L10) | `[OK] No errors` |
| `composer lint` (phpcs) | clean |
| `composer parallel-lint` | 572 files, no syntax errors |

Baseline on `development` was 2222 / 0 failures; the delta is exactly the 10 new tests. Every new assertion was confirmed red before its fix — the weekday cases failed against the unmodified engine while the Sunday and Saturday cases already passed, which is the correct shape for a change that only ever *adds* a colour.

## Summary by CodeRabbit

- **New Features**
  - Added support for conditional, date-dependent liturgical colors in source data.
  - Ambrosian All Souls observance now includes black alongside morello on non-Sundays, including Saturdays; vigils inherit the resolved colors.
  - Added schema support for optional conditional color definitions.

- **Bug Fixes**
  - Improved validation for scalar and array color values, including conditional entries.
  - Ensured resolved conditional colors appear consistently in localized results.

- **Tests**
  - Added coverage for date resolution, ordering, localization, vigil inheritance, and schema validation.

### DIFF
--- a/jsondata/schemas/CommonDef.json
+++ b/jsondata/schemas/CommonDef.json
@@ -72,6 +72,43 @@
                 ]
             }
         },
+        "LitColorAdLibitum": {
+            "type": "array",
+            "title": "Ad Libitum Liturgical Colour",
+            "description": "source-data only: liturgical colours admitted *ad libitum*, each gated by a named condition that the engine evaluates against the computed date. The engine appends the permitted colours to the event's `color` array, so the API response carries the outcome and never the condition. Used for colours whose admissibility depends on the day — e.g. the Ambrosian rite admits `black` in Masses for the dead only on days that are neither a Sunday nor the vigil opening one (Ordinamento Generale del Messale Ambrosiano n. 320). See issue #771 for the rite-scoped palettes these colours are also validated against.",
+            "uniqueItems": true,
+            "minItems": 1,
+            "items": {
+                "type": "object",
+                "properties": {
+                    "color": {
+                        "type": "string",
+                        "description": "the colour admitted ad libitum when the condition holds",
+                        "enum": [
+                            "white",
+                            "red",
+                            "green",
+                            "purple",
+                            "rose",
+                            "morello",
+                            "black"
+                        ]
+                    },
+                    "when": {
+                        "type": "string",
+                        "description": "the named predicate gating this colour; a closed vocabulary evaluated by the engine, mirroring the `AdLibitumColorCondition` PHP enum",
+                        "enum": [
+                            "not_sunday_and_not_sunday_vigil"
+                        ]
+                    }
+                },
+                "required": [
+                    "color",
+                    "when"
+                ],
+                "additionalProperties": false
+            }
+        },
         "LitCommon": {
             "type": "array",
             "title": "Liturgical Common",
@@ -8267,7 +8304,9 @@
                     "type": "object",
                     "oneOf": [
                         {
-                            "required": ["if_weekday"],
+                            "required": [
+                                "if_weekday"
+                            ],
                             "additionalProperties": false,
                             "properties": {
                                 "if_weekday": {
@@ -8285,7 +8324,9 @@
                             }
                         },
                         {
-                            "required": ["if_grade"],
+                            "required": [
+                                "if_grade"
+                            ],
                             "additionalProperties": false,
                             "properties": {
                                 "if_grade": {
@@ -8299,7 +8340,9 @@
                     "type": "object",
                     "oneOf": [
                         {
-                            "required": ["move"],
+                            "required": [
+                                "move"
+                            ],
                             "additionalProperties": false,
                             "properties": {
                                 "move": {
@@ -8309,7 +8352,9 @@
                             }
                         },
                         {
-                            "required": ["move_to"],
+                            "required": [
+                                "move_to"
+                            ],
                             "additionalProperties": false,
                             "properties": {
                                 "move_to": {
--- a/jsondata/schemas/CommonDef.json
+++ b/jsondata/schemas/CommonDef.json
@@ -98,7 +98,7 @@
                         "type": "string",
                         "description": "the named predicate gating this colour; a closed vocabulary evaluated by the engine, mirroring the `AdLibitumColorCondition` PHP enum",
                         "enum": [
-                            "not_sunday_and_not_sunday_vigil"
+                            "not_sunday"
                         ]
                     }
                 },
--- a/jsondata/schemas/PropriumDeSanctis.json
+++ b/jsondata/schemas/PropriumDeSanctis.json
@@ -26,7 +26,10 @@
                     "$ref": "./CommonDef.json#/definitions/LitGrade"
                 },
                 "grade_display": {
-                    "type": ["string", "null"]
+                    "type": [
+                        "string",
+                        "null"
+                    ]
                 },
                 "common": {
                     "$ref": "./CommonDef.json#/definitions/LitCommon"
@@ -44,6 +47,9 @@
                 "is_bvm": {
                     "type": "boolean",
                     "description": "Marks a celebration of the Blessed Virgin Mary, used by the Ambrosian precedence classifier to rank a memorial of the BVM above a memorial of a saint within the comune memorial tier. Optional; omitted for events where the distinction does not apply. Not serialized in API responses."
+                },
+                "color_ad_libitum": {
+                    "$ref": "CommonDef.json#/definitions/LitColorAdLibitum"
                 }
             },
             "required": [
--- a/jsondata/sourcedata/rite/ambrosian/missals/propriumdesanctis_2024/propriumdesanctis.json
+++ b/jsondata/sourcedata/rite/ambrosian/missals/propriumdesanctis_2024/propriumdesanctis.json
@@ -211,7 +211,7 @@
     { "month": 10, "day": 28, "event_key": "StsSimonJudeApostles", "grade": 4, "common": [ "Proper" ], "calendar": "AMBROSIAN", "color": [ "red" ] },
     { "month": 10, "day": 29, "event_key": "StHonoratusOfVercelli", "grade": 2, "common": [ "Pastors:For a Bishop" ], "calendar": "AMBROSIAN", "color": [ "white" ] },
     { "month": 11, "day": 1, "event_key": "AllSaints", "grade": 6, "common": [ "Proper" ], "calendar": "AMBROSIAN", "color": [ "white" ] },
-    { "month": 11, "day": 2, "event_key": "AllSouls", "grade": 6, "common": [ "Proper" ], "calendar": "AMBROSIAN", "color": [ "morello" ] },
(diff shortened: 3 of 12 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
## Correction: the condition excludes Sunday only

@JohnRDOrazio is right that `color` is per **celebration**, not per day — my earlier "day granularity" reasoning was wrong, and the restrictive reading it produced would have denied the faculty at a Saturday Mass for the dead that n. 320 permits. Pushed in `0c7eecaf`.

Checking what the engine actually emits settles the vigil question. All Souls on Saturday 2 November 2024:

```console
2024-11-01 Fri  AllSouls_vigil                   morello   (is_vigil_for=AllSouls)
2024-11-02 Sat  AllSouls                         morello
2024-11-02 Sat  AfterPentecostDedication2_vigil  green     (opens the Sunday)
2024-11-03 Sun  AfterPentecostDedication2        green
```

The Mass opening the following Sunday is **that Sunday's own vigil** — a separate event carrying that Sunday's colours, never All Souls'. And when All Souls itself falls on a Sunday, `AllSouls_vigil` is the Saturday evening that opens it and inherits `morello` only, since a vigil inherits the colours of the celebration it belongs to.

So the Sunday-vigil case needs no special handling: evaluating the condition against the celebration's own date is exactly right.

| Change | |
|---|---|
| `NOT_SUNDAY_AND_NOT_SUNDAY_VIGIL` → `NOT_SUNDAY` | `[6, 7]` → `!== 7` |
| 2019 and 2024 (Saturdays) | now expect `["morello", "black"]` |
| `testAllSoulsVigilInheritsResolvedColors` | new — pins the inheritance in both directions, so the reasoning above is regression-tested rather than left in a comment |

I also corrected the rationale for appending to `color`. The array expresses alternatives **within one celebration** — either a choice of common (`["white", "red"]` for a martyr-bishop) or an *ad libitum* option (`["rose", "purple"]` on Gaudete/Laetare) — not multiple colours in a day, which is how I had originally framed it.

Verification after the change: 2234 tests, 0 failures; PHPStan L10 and phpcs clean.

PR #785, opened 2026-08-14
feat(tests): scope liturgical tests to a rite (#767)

### LINKED ISSUE #767: Liturgical tests cannot be scoped to a rite: Ambrosian calendars are untestable
## Summary

Liturgical tests (`jsondata/tests/*.json`) cannot be scoped to a rite. `applies_to` accepts only national and diocesan calendars, so **no Ambrosian calendar can be tested** — and for Ambrosian *dioceses*, which `applies_to` can already name, the runner builds a path the API rejects.

This surfaced while making the frontend rite-aware (LiturgicalCalendarFrontend#446, which exposes the Ambrosian rite in the calendar-subscription UI). The rite is now a first-class dimension everywhere except the test suite.

### 1. Rite-level calendars are inexpressible

`jsondata/schemas/LitCalTest.json` → `definitions.AppliesToOrExcludes` sets `additionalProperties: false` and permits only `national_calendar`, `national_calendars`, `diocesan_calendar`, `diocesan_calendars`, with an `anyOf` requiring at least one of them. There is no key for a rite.

The Ambrosian rite-level calendar (`GET /calendar/ambrosian`, announced as `ambrosian_calendars` in `/calendars` metadata) is neither national nor diocesan, so it cannot be the subject of a test at all.

`Services/TestScopeResolver::mapAppliesTo()` defaults anything unscoped to `['general_roman_calendar_test', 'general_roman_calendar']`, so an Ambrosian test written without a scope silently runs against the **General Roman Calendar** and fails every assertion.

That failure mode is not hypothetical. A `StIgnatiusOfLoyolaTest.json` was found on disk asserting the Ambrosian `StIgnatiusOfLoyola` (a real key — the Roman equivalent is `StIgnatiusLoyola`) on 31 July across 1999–2030. The assertions are correct for the Ambrosian calendar, but with no way to say so the test evaluates against the GRC, where that key does not exist, and all 32 assertions fail. It was deleted rather than fixed, because it is not currently fixable.

### 2. Ambrosian dioceses are expressible but unrunnable

`Health::buildCalendarRequestPath()` (`src/Health.php:988-992`) emits:

```php
'nationalcalendar'  => "/nation/$calendar/$year?year_type=CIVIL",
'diocesancalendar'  => "/diocese/$calendar/$year?year_type=CIVIL",
```

No rite segment. But the router requires one for Ambrosian dioceses. Verified against a running instance:

```text
/calendar/diocese/lugano_ch/2026?year_type=CIVIL              -> 400
/calendar/ambrosian/diocese/lugano_ch/2026?year_type=CIVIL    -> 200
/calendar/ambrosian/2026?year_type=CIVIL                      -> 200
/calendar/diocese/romamo_it/2026?year_type=CIVIL              -> 200   (Roman control)
```

So a test scoped `{"diocesan_calendar": "lugano_ch"}` — which the schema accepts today — would request a URL that 400s. The four Ambrosian dioceses (`milano_it`, `bergam_it`, `novara_it`, `lugano_ch`) are all affected.

## Scope of the change

Rite scoping is not only a schema concern — `TestScopeResolver` maps `applies_to` onto **OpenFGA object types**, so authorization is in the blast radius.

| Area | File | Change |
| ---- | ---- | ------ |
| Schema | `jsondata/schemas/LitCalTest.json` | Add a rite key to `AppliesToOrExcludes` (blocked by `additionalProperties: false`) |
| Validation | `src/Test/TestItem.php` (`checkAppliesToExcludesConditions`) | Accept and validate the new key |
| Scope → FGA | `src/Services/TestScopeResolver.php` (`mapAppliesTo`) | Map rite scope to an object type |
| Authorization | `authz/openfga-expectations.json` + the FGA model | Likely a new `rite_calendar_test` type alongside the existing `national_calendar_test` / `diocesan_calendar_test` / `general_roman_calendar_test` |
| Runner | `src/Health.php` (`buildCalendarRequestPath`) | Emit the rite segment; fixes gap 2 even for already-expressible diocesan scopes |
| Frontend | `admin-tests.php` editor | Offer the rite in the scope picker |

## Design questions

1. **Is the rite an independent axis, or a property of the calendar being named?** `{"rite": "ambrosian"}` for the rite-level calendar, versus inferring the rite from the calendar id (the API already knows `lugano_ch` is Ambrosian). Inference would fix gap 2 with no schema change and no new FGA type — worth considering first, since it is much smaller. An explicit key would still be needed for gap 1.
2. **Does the rite-level calendar need its own FGA object type**, or can it reuse `general_roman_calendar_test` renamed to something rite-neutral? The latter is a migration; the former grows the model.
3. **Should an unscoped test keep defaulting to the General Roman Calendar?** That default is what made the deleted test fail silently rather than loudly. An explicit required scope would be a breaking change to the 10 existing tests, but would remove a whole class of silent misconfiguration.

## Acceptance

- A test can be scoped to the Ambrosian rite-level calendar and runs against `/calendar/ambrosian`.
- A test scoped to an Ambrosian diocese requests `/calendar/ambrosian/diocese/{id}` and passes.
- The `StIgnatiusOfLoyola` case above is expressible and green.
- Existing national / diocesan / GRC tests are unaffected.
- OpenFGA authorization covers whatever new scope is introduced.

## Related

- LiturgicalCalendarFrontend#446 — makes the frontend rite-aware; the change that exposed this.
- #760 — Echo the rite in the `/calendar` response (adjacent: the runner would benefit from the response stating its rite).

JohnRDOrazio:
Closes #767.

Liturgical tests could not be scoped to a rite, with two consequences: the Ambrosian rite-level calendar
(`GET /calendar/ambrosian`) was untestable at all, and a test scoped to one of the four Ambrosian dioceses —
a scope the schema already accepted — produced a URL the router rejects with a 400.

The three design questions in the issue were answered in its comments; this implements those answers.

## Scope shape

`applies_to` is now **required** on every test, with a **required `rite`**:

```json
{ "applies_to": { "rite": "roman" } }
{ "applies_to": { "rite": "roman", "national_calendar": "US" } }
{ "applies_to": { "rite": "ambrosian" } }
{ "applies_to": { "rite": "ambrosian", "diocesan_calendar": "lugano_ch" } }
```

`AppliesToOrExcludes` splits into `AppliesToScope` (rite required, calendar keys optional) and `ExcludesScope`
(unchanged, and it rejects `rite` — the rite is already pinned by `applies_to`, so excluding one would leave
nothing to test).

Making the rite mandatory is the point rather than a side effect: `TestScopeResolver` previously defaulted an
unscoped test to the General Roman Calendar, which is exactly how the deleted `StIgnatiusOfLoyolaTest` came to
fail all 32 of its assertions silently.

## Authorization

New `rite_calendar_test` OpenFGA object type, keyed by rite id (`roman`, `ambrosian`). It generalises
`general_roman_calendar_test`, whose single fixed id `general_roman_calendar` denoted exactly the Roman
rite-level calendar.

| `applies_to`                           | FGA scope                     |
|----------------------------------------|-------------------------------|
| `{"diocesan_calendar": "<id>"}`         | `diocesan_calendar_test:<id>`  |
| `{"national_calendar": "<id>"}`         | `national_calendar_test:<id>`  |
| `{"rite": "<rite>"}` (no calendar key)  | `rite_calendar_test:<rite>`    |
| absent / unrecognised                   | `rite_calendar_test:roman`     |

The diocesan and national branches keep precedence and stay keyed by calendar id alone — those ids are unique
across rites today, and the matching *data* resource types (`diocesan_calendar`, `national_calendar`) are keyed
the same way, so rite-qualifying only the test scope would make the two halves of the model disagree.

**The migration is additive.** `general_roman_calendar_test` stays in the model and in every PHP allow-list, so
existing tuples keep authorizing and a rollback to pre-#767 code is safe. `scripts/migrate-rite-test-tuples.php`
copies each grant onto `rite_calendar_test:roman`; it is copy-only by default and `--prune` is gated behind full
rollout. Runbook section added to `docs/ops/test-scope-migration-runbook.md`.

## Runner

`Health::buildCalendarRequestPath()` now takes a `Rite` and always emits the canonical explicit segment:

| category           | path                                                |
|--------------------|-----------------------------------------------------|
| `ritecalendar`     | `/{rite}/{year}?year_type=CIVIL`                     |
| `nationalcalendar` | `/{rite}/nation/{calendar}/{year}?year_type=CIVIL`   |
| `diocesancalendar` | `/{rite}/diocese/{calendar}/{year}?year_type=CIVIL`  |

`ritecalendar` is a new category for the rite-level calendar; the historical `calendar === 'VA'` marker (which
already meant "the universal calendar", not `/nation/VA`) resolves the same way.

`Health::resolveRite()` prefers an explicit `rite` on the WebSocket message, and otherwise reads it from
`/calendars` metadata for diocesan calendars. That second step fixes gap 2 for the **existing** UnitTestInterface
client with no client release required; `rite` is optional on the message, so `ACTION_PROPERTIES` is unchanged.

## Failing loudly

`/calendar` responses already echo `settings.rite` (#760). `LitTestRunner` now compares it against the test's
declared rite and reports one clear error on mismatch instead of running assertions that would all fail for the
wrong reason:

```text
StIgnatiusOfLoyolaTest is scoped to the ambrosian rite, but the calendar under test was computed under the roman rite
```

`getCalendarName()` also learns to name the rite-level calendar ("the General Roman Calendar" / "the Ambrosian
Calendar") instead of hard-coding "the Universal Roman Calendar".

## Data

All ten existing tests are scoped `rite: roman` — no behaviour change for any of them.

`StIgnatiusOfLoyolaTest.json` is restored, scoped `rite: ambrosian`. **The copy found on disk was itself wrong:**
it asserted the memorial on 31 July for every year 1999–2030, but in the Ambrosian calendar the memorial is
suppressed when 31 July falls on a Sunday. Verified against a running instance — 2005, 2011, 2016 and 2022 return
no `StIgnatiusOfLoyola` event at all — so those four assertions are now `eventNotExists`, matching the pattern
already used in `StCamillusDeLellisTest.json`.

`SchemaValidationTest::testRealTestSourceValidation` was validating only `glob(...)[0]` — one file, whichever
sorts first. Widened to a data provider over the whole corpus, so a test file that never gained a rite fails
loudly instead of hiding behind its neighbours.

## Verification

Full suite green: **2463 tests, 478878 assertions, 14 skipped, 0 failures**. PHPStan level 10 clean, phpcs clean,
`composer lint:openapi` back to zero warnings (matching the pre-change baseline), markdownlint clean.

Acceptance criteria checked end-to-end against a running instance, driving the real `Health` path builder and the
real `LitTestRunner`:

```text
=== 1. StIgnatiusOfLoyolaTest (rite: ambrosian) against the Ambrosian rite-level calendar ===
  32 passed / 0 failed across 32 years
  path used: /ambrosian/2026?year_type=CIVIL

=== 2. Ambrosian diocesan request path ===
  milano_it   /ambrosian/diocese/milano_it/2026?year_type=CIVIL   OK
  bergam_it   /ambrosian/diocese/bergam_it/2026?year_type=CIVIL   OK
  novara_it   /ambrosian/diocese/novara_it/2026?year_type=CIVIL   OK
  lugano_ch   /ambrosian/diocese/lugano_ch/2026?year_type=CIVIL   OK

=== 3. Rite-mismatch guard ===
  Ambrosian test vs Roman calendar -> OK: error: StIgnatiusOfLoyolaTest is scoped to the ambrosian rite,
  but the calendar under test was computed under the roman rite

=== 4. Existing Roman tests still pass ===
  MaryMotherChurchTest       /roman/2019?year_type=CIVIL                   SUCCESS
  NativityJohnBaptistTest    /roman/2022?year_type=CIVIL                   SUCCESS
  ThanksgivingDayTest        /roman/nation/US/2024?year_type=CIVIL         SUCCESS
  rotter_nl_HLaurentius...   /roman/diocese/rotter_nl/2024?year_type=CIVIL SUCCESS
```

## Follow-ups (deliberately out of scope)

- **Frontend `admin-tests.php` scope picker.** Requiring `applies_to.rite` is a breaking change for that editor's
  PUT/PATCH payloads, so `LiturgicalCalendarFrontend` needs a companion change before this lands anywhere the two
  are deployed together.
- **Removing `general_roman_calendar_test`**, once every deployment runs merged code — steps listed in the runbook.
- **Rite-qualifying diocesan/national calendar ids** in the resource model, if a calendar id ever needs to exist
  under two rites. That is a change to the whole resource model, not to tests.

Design spec: `docs/superpowers/specs/2026-08-14-rite-scoped-liturgical-tests-design.md`.

## Summary by CodeRabbit

- **New Features**
  - Added rite-aware calendar requests and unit-test execution for Roman and Ambrosian rites.
  - Added rite-specific authorization scopes, dashboard visibility, and qualified calendar test identifiers.
  - Added validation for rite-scoped applicability and response-rite mismatches.
  - Added Ambrosian liturgical test coverage and updated existing tests with explicit Roman scope.

- **Documentation**
  - Added design guidance and an operational migration runbook for rite-scoped tests.

- **Refactor**
  - Separated applicability and exclusion rules, requiring a declared rite.
  - Added migration support for existing calendar-test permissions.

### DIFF
--- a/authz/openfga-expectations.json
+++ b/authz/openfga-expectations.json
@@ -9,7 +9,8 @@
     "general_roman_calendar",
     "national_calendar_test",
     "diocesan_calendar_test",
-    "general_roman_calendar_test"
+    "general_roman_calendar_test",
+    "rite_calendar_test"
   ],
   "forbidden_types": [
     "test_definition"
--- a/docs/ops/test-scope-migration-runbook.md
+++ b/docs/ops/test-scope-migration-runbook.md
@@ -251,3 +251,66 @@ The migration writes the new tuple **before** deleting the old one, so:
   for this; manual OpenFGA tuple writes or a reverse migration script would be
   needed. Prefer rolling forward rather than rolling back once all environments
   are migrated.
+
+---
+
+## Follow-up migration — `general_roman_calendar_test` → `rite_calendar_test`
+
+Issue #767 made the rite a first-class scope for liturgical tests. The Ambrosian
+rite-level calendar (`GET /calendar/ambrosian`) is neither national nor diocesan,
+so `general_roman_calendar_test` — a type with exactly one id,
+`general_roman_calendar` — could not name it. `TestScopeResolver` now emits
+`rite_calendar_test:<rite>` for every rite-level test:
+
+| Type                          | Object id                                | Status                        |
+|-------------------------------|------------------------------------------|-------------------------------|
+| `general_roman_calendar_test` | Fixed id `general_roman_calendar`        | Deprecated; kept in the model |
+| `rite_calendar_test`          | A `Rite` value: `roman` / `ambrosian`    | Successor                     |
+
+The shape of this migration mirrors the one above, with one deliberate
+difference: it is **copy-only by default**. The legacy tuples are left in place
+so that rolling the API back to pre-#767 code keeps authorizing.
+
+### Step 1 — apply the model
+
+`rite_calendar_test` is added to `scripts/openfga-model.additive.json` alongside
+`general_roman_calendar_test`, with an identical relation set. As above, the
+authoritative model lives in `cdcf-infra`: land the equivalent change to
+`cdcf-infra/auth/models/LiturgicalCalendar.json`, then upload the new model
+version on the VPS.
+
+### Step 2 — copy the tuples
+
+```bash
+# Dry run first — prints every tuple that would be copied.
+php scripts/migrate-rite-test-tuples.php
+
+# Apply.
+php scripts/migrate-rite-test-tuples.php --apply
+```
+
+Each `general_roman_calendar_test:general_roman_calendar` tuple gains a
+counterpart on `rite_calendar_test:roman`, preserving user and relation.
+Re-running is a no-op. An id other than `general_roman_calendar` on the legacy
+type is reported and left untouched — the script never guesses a rite — and the
+run exits with status `2` so the anomaly is not lost in CI output.
+
+### Step 3 — prune (later)
+
+Only once **every** deployment runs post-#767 code:
+
+```bash
+php scripts/migrate-rite-test-tuples.php --apply --prune
+```
+
+Then drop `general_roman_calendar_test` from the model and from the PHP
+allow-lists that still name it:
+
+- `src/Services/TestScopeResolver.php` (docblock only — it no longer emits it)
+- `src/Services/ResourceExistenceChecker.php`
+- `src/Services/ResourceAdminService.php`
+- `src/Repositories/AccessRequestRepository.php`
+- `authz/openfga-expectations.json`
+- `jsondata/schemas/openapi.json`
+
+This step is **out of scope for the current change** and tracked as a follow-up.
--- a/docs/ops/test-scope-migration-runbook.md
+++ b/docs/ops/test-scope-migration-runbook.md
@@ -6,7 +6,7 @@ This branch (`feat/calendar-scoped-tests`) replaces the flat `test_definition`
 OpenFGA type with three calendar-scoped types:
 
 | New type                      | Scope                   | Object id                         |
-| ----------------------------- | ----------------------- | --------------------------------- |
+|-------------------------------|-------------------------|-----------------------------------|
 | `national_calendar_test`      | National-calendar tests | ISO nation code, e.g. `IT`        |
 | `diocesan_calendar_test`      | Diocesan-calendar tests | Diocese id, e.g. `ROMA`           |
 | `general_roman_calendar_test` | GRC / unscoped tests    | Fixed id `general_roman_calendar` |
@@ -254,22 +254,38 @@ The migration writes the new tuple **before** deleting the old one, so:
 
 ---
 
-## Follow-up migration — `general_roman_calendar_test` → `rite_calendar_test`
+## Follow-up migration — rite-qualified test scopes
 
-Issue #767 made the rite a first-class scope for liturgical tests. The Ambrosian
-rite-level calendar (`GET /calendar/ambrosian`) is neither national nor diocesan,
-so `general_roman_calendar_test` — a type with exactly one id,
-`general_roman_calendar` — could not name it. `TestScopeResolver` now emits
-`rite_calendar_test:<rite>` for every rite-level test:
+Issue #767 made the rite a first-class scope for liturgical tests. Two things
+follow, and one migration covers both.
 
-| Type                          | Object id                                | Status                        |
-|-------------------------------|------------------------------------------|-------------------------------|
-| `general_roman_calendar_test` | Fixed id `general_roman_calendar`        | Deprecated; kept in the model |
-| `rite_calendar_test`          | A `Rite` value: `roman` / `ambrosian`    | Successor                     |
+**The rite-level calendar needed a type.** The Ambrosian rite-level calendar
+(`GET /calendar/ambrosian`) is neither national nor diocesan, so
+`general_roman_calendar_test` — a type with exactly one id,
+`general_roman_calendar` — could not name it.
+
+**Scoped calendar ids needed a rite.** A bare calendar id does not identify a
+calendar: the source tree is partitioned as
+`jsondata/sourcedata/rite/{rite}/calendars/...`, so `lugano_ch` could name an
+Ambrosian calendar or a Roman one, and granting `diocesan_calendar_test` on a
+bare `lugano_ch` would be an ambiguous grant. National scopes are qualified too,
+for one uniform rule — even though only the Roman rite has a national tier.
+
+| Type                          | Object id before         | Object id after                       |
+|-------------------------------|--------------------------|---------------------------------------|
+| `general_roman_calendar_test` | `general_roman_calendar` | → `rite_calendar_test:roman`          |
+| `rite_calendar_test`          | *(new)*                  | A `Rite` value: `roman` / `ambrosian` |
+| `national_calendar_test`      | `US`                     | `roman/US`                            |
+| `diocesan_calendar_test`      | `lugano_ch`              | `ambrosian/lugano_ch`                 |
+
+The rite of an existing national or diocesan tuple is inferred from the
+rite-partitioned source tree, which is the authority on which rite a calendar is
+defined under. An id defined under two rites is reported and skipped — the script
+never guesses which grant was meant.
 
 The shape of this migration mirrors the one above, with one deliberate
-difference: it is **copy-only by default**. The legacy tuples are left in place
-so that rolling the API back to pre-#767 code keeps authorizing.
+difference: it is **copy-only by default**. The superseded tuples are left in
+place so that rolling the API back to pre-#767 code keeps authorizing.
 
 ### Step 1 — apply the model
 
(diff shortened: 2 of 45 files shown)


JohnRDOrazio:
Thanks — three of the four applied in `cc4f4ce4`, one declined with reasoning.

**1. Reject Ambrosian national-calendar scopes — applied.** Good catch, and it is the same
expressible-but-unrunnable trap the issue is about, just one level up. Rejected in both places:

- `LitCalTest.json` — a conditional subschema on `AppliesToScope` forbidding `national_calendar` /
  `national_calendars` when `rite` is `ambrosian`. (Written as `"national_calendar": false` under
  `then.properties` rather than `not`/`anyOf`/`required`; the latter form is equivalent but trips Redocly's
  `no-required-schema-properties-undefined`, and this repo's `composer lint:openapi` baseline is zero warnings.)
- `TestItem::checkAppliesToConditions()` — the same rule at runtime, phrased generally (`Rite::ROMAN !== $rite`)
  since national calendars are a Roman-rite structure, not an Ambrosian-specific exception.

Verified across the matrix — `ambrosian`+national rejected, `ambrosian`+diocesan and all Roman combinations
still accepted — with rejection coverage added in `TestItemTest`.

**2. Restore the uninitialized metadata state — applied, with a caveat.** `Health::$metadata` is
`private static MetadataCalendars $metadata;` — a typed static with no default, and PHP offers no way to return
one to the genuinely uninitialised state. The `else` branch therefore installs an *empty* `MetadataCalendars`,
which addresses the actual hazard you identified: a later test in the same process can no longer resolve
`lugano_ch` or `rotter_nl` from data this helper invented. Documented in the helper's docblock.

**3. Make the rite probe observable — applied, both sites.** `ResourceAdminServiceTest` now returns
`rite_calendar_test:ambrosian` for the rite editor and admin probes and asserts both; `TestScopesHandlerTest`
replaces the first element of the queue instead of prepending a ninth response, so it is exactly one response
per probe.

**4. `#[Group('slow')]` on four test classes — declined for three of the four.**

Applied to `SchemaValidationTest::testRealTestSourceValidation`, which is the one place the finding is right:
that method carried a legacy `@group slow` docblock that PHPUnit 12 does not honour, so `--group slow` selected
zero tests there. It now uses the attribute and is actually selectable. CI runs the full suite
(`composer test:coverage`, no group exclusions), so corpus coverage is unchanged.

Not applied to `TestItemTest`, `AccessRequestRepositoryConstantsTest`, or `ResourceExistenceCheckerTest`. None of
them declares any group today, and all 44 of their tests together run in 12 ms:

```text
$ vendor/bin/phpunit phpunit_tests/Test/TestItemTest.php \
    phpunit_tests/Repositories/AccessRequestRepositoryConstantsTest.php \
    phpunit_tests/Services/ResourceExistenceCheckerTest.php
Time: 00:00.012  OK (44 tests, 117 assertions)
```

`CLAUDE.md` reserves `slow` for tests with *measurable* runtime cost — multi-year calendar calculations,
rate-limit window waits, full-schema-corpus validation — and explicitly notes that integration tests are not
automatically slow. These are pure in-memory unit tests, so marking them `slow` would exclude them from
`composer test:quick` for no reason and misrepresent what the group means. The guideline the finding cites is
about *how* to declare `slow` when a test is slow, not about which classes should be.

CI is green and `composer lint:openapi` is back to zero warnings.

JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
Scope ids are now rite-qualified for the national and diocesan test types as well, per review (`68f55c2e`):

```text
diocesan_calendar_test:ambrosian/lugano_ch
diocesan_calendar_test:roman/rotter_nl
national_calendar_test:roman/US
```

A bare calendar id does not identify a calendar — the source tree is partitioned as `jsondata/sourcedata/rite/{rite}/calendars/...`, so a second `lugano_ch` under `roman/` is a directory away, and a grant on the bare id would silently widen to cover it. National scopes are qualified too, so there is one rule rather than a carve-out; `national_calendar_test:ambrosian/*` is rejected rather than merely unused.

The migration script grew accordingly: it now rewrites existing `national_calendar_test` / `diocesan_calendar_test` tuples too, inferring each calendar's rite from the rite-partitioned source tree and skipping — never guessing — anything defined under two rites or none.

While updating the rejection path I also found a pre-existing bug: a rejected id always listed the General Roman Calendar's valid ids regardless of type, so rejecting `national_calendar_test:IT` reported `Valid ids: temporale, EDITIO_TYPICA_1970, …`. Replaced with a type-aware `AccessRequestRepository::validIdsLabelForType()`.

The **data** resource types (`national_calendar`, `diocesan_calendar`, `wider_region`) are deliberately unchanged here and tracked in #786 — same ambiguity, but the blast radius is production calendar-editing grants plus `/data` rite-awareness. Scoping that issue turned up a live bug worth a look: `ResourceExistenceChecker` globs only the Roman partition, so all four Ambrosian dioceses report as not existing and the reconciler sweep would purge their grants.

Full suite 2472 green, PHPStan level 10 clean, phpcs clean, `lint:openapi` at zero warnings, e2e acceptance 41/41.

Issue #787, opened 2026-08-14
/tests is not rite-aware: the rite is not part of a test's identity, and the collection cannot be filtered by it


JohnRDOrazio:
## Summary

`/calendar` and `/events` take an optional leading rite segment (`/calendar/ambrosian/...`). `/data` does not
(#786). `/tests` does not either, and unlike `/data` it is not merely an addressing gap: **a test's identity
is its bare file stem**, so the rite is not part of what names a test.

#785 made the rite a required property *inside* each test (`applies_to.rite`) and rite-qualified the
authorization scopes. The resource identity was deliberately left alone. This issue tracks whether it should
be, and the collection-level ergonomics that follow.

## The routes today

`src/Router.php:575-600`:

```text
GET  | POST                      /tests
GET  | POST | PUT | PATCH | DELETE   /tests/{test_name}
```

`{test_name}` maps directly to `jsondata/tests/{test_name}.json`, and `TestScopeResolver::isSafeName()`
restricts it to `[A-Za-z0-9_-]+`. There is no rite segment and no filtering on the collection.

## Gap 1 — the rite is not part of a test's identity

Test names are globally unique file stems, so two tests that differ *only* by rite have to be disambiguated
by hand in the name. That is already live: `StIgnatiusOfLoyolaTest` is the **Ambrosian** memorial of St
Ignatius of Loyola. A Roman-rite test for the same saint (`StIgnatiusLoyola` there — a different event key)
would need some other name, chosen by convention rather than enforced by structure.

The existing name pattern in `LitCalTest.json` allows an optional lower-case prefix —

```text
^(?:[a-z_]+?_){0,1}[A-Z][a-zA-Z1-9]+[0-9]{0,2}(?:_vigil)?Test$
```

— which is how `rotter_nl_HLaurentius...Test` scopes itself by diocese in the name. So the convention
"encode the scope in the name" already exists informally. Whether the rite should join it, become a path
segment, or stay implicit is the open question.

Note this is *not* an authorization hole: `TestScopeResolver` reads `applies_to.rite` from the file, so
`/tests/StIgnatiusOfLoyolaTest` already authorizes against `rite_calendar_test:ambrosian`. The gap is
identity and discoverability, not access control.

## Gap 2 — the collection cannot be filtered by rite

`GET /tests` returns every test. A client that wants "the Ambrosian tests" — UnitTestInterface when the user
[...]

PR #788, opened 2026-08-14
feat(authz): rite-qualify the data resource scopes and make /data rite-aware (#786)

### LINKED ISSUE #786: Authorization: data resource types are keyed by an ambiguous bare calendar id (and Ambrosian dioceses read as deleted)
## Summary

The OpenFGA **data** resource types are keyed by a bare calendar id: `diocesan_calendar:lugano_ch`,
`national_calendar:US`, `wider_region:europe`. A bare calendar id does not identify a calendar. The source
tree is partitioned by rite —

```text
jsondata/sourcedata/rite/roman/calendars/dioceses/{NATION}/{diocese}/
jsondata/sourcedata/rite/ambrosian/calendars/dioceses/{NATION}/{diocese}/
```

— so nothing stops the same diocese being defined under both rites. `lugano_ch` only *happens* to be
Ambrosian-only today. A grant on `diocesan_calendar:lugano_ch` is therefore ambiguous, and would silently
widen to cover a Roman `lugano_ch` calendar the moment one was added.

#785 fixed this for the **test** scopes, which now carry the rite in the object id
(`diocesan_calendar_test:ambrosian/lugano_ch`). It deliberately left the data resource types alone, because
rite-qualifying those touches `/data` authorization and live calendar-editing grants. This issue tracks
finishing the job so the two halves of the authorization model agree.

## A live bug, found while investigating

`ResourceExistenceChecker::exists()` (`src/Services/ResourceExistenceChecker.php:67-74`) resolves a diocesan
calendar by globbing `JsonData::DIOCESAN_CALENDARS_FOLDER`, which is **Roman-only** —
`JsonDataConstants::DIOCESAN_CALENDARS_FOLDER` derives from `ROMAN_RITE_FOLDER`, while the Ambrosian
partition has its own separate constant `AMBROSIAN_DIOCESAN_CALENDARS_FOLDER`
(`src/Enum/JsonDataConstants.php:185`, `:441`) that the checker never consults.

Every Ambrosian diocese therefore reports as not existing:

```text
ResourceExistenceChecker::exists('diocesan_calendar', ...)
  rotter_nl    (roman    ) => true
  romamo_it    (roman    ) => true
  boston_us    (roman    ) => true
  lugano_ch    (ambrosian) => false
  milano_it    (ambrosian) => false
  bergam_it    (ambrosian) => false
  novara_it    (ambrosian) => false
```

That matters because the reconciler sweep purges on exactly this predicate
(`src/Services/Outbox/ResourceTuplePurgeReconciler.php:69-72`):

```php
if ($this->checker->exists($type, $id)) {
    continue; // resource still present — operational tuples are live
}
$enqueued += $this->purge->purgeForObject($object);
```

**So a legitimate `diocesan_calendar:milano_it` editor grant looks orphaned and gets purged by any
reconciler run.** This is independent of the modelling question below and is arguably the more urgent half:
it is a silent revocation of real permissions, and it is live today for all four Ambrosian dioceses.

## `/data` is not rite-aware either

Verified against a running instance:

```text
/data/diocese/rotter_nl   -> 422   (reached the handler)
/data/diocese/lugano_ch   -> 404   (not addressable at all)
```

So the Ambrosian diocesan calendars cannot currently be read or written through `/data` under any path.
There is no rite segment on the `/data` routes the way there is on `/calendar`. Whatever id shape the
authorization model settles on has to line up with how `/data` addresses these calendars, so the two should
be designed together.

## Scope of the change

| Area                | File                                                     | Change                                                                  |
| ------------------- | -------------------------------------------------------- | ----------------------------------------------------------------------- |
| Existence check     | `src/Services/ResourceExistenceChecker.php`               | Consult both rite partitions; stop reporting Ambrosian dioceses as gone |
| Scope → FGA         | `src/Http/Middleware/OpenFgaAuthorizationMiddleware.php`  | `OBJECT_TYPE_MAP` + object-id composition (lines 44-48)                 |
| Handler             | `src/Handlers/RegionalDataHandler.php`                    | Category → object type mapping (around line 961)                        |
| Request validation  | `src/Repositories/AccessRequestRepository.php`            | `isValidObjectIdForType()`, `validIdsLabelForType()`                    |
| Dashboards          | `src/Services/ResourceAdminService.php`                   | `ADMIN_OBJECT_TYPES` / `VIEWER_OBJECT_TYPES` consumers                  |
| Routes              | `src/Router.php`                                          | Rite segment on `/data`, if that is the direction                       |
| Tuple migration     | `scripts/` + `docs/ops/`                                  | Rewrite live grants; runbook section                                    |
| Frontend            | `LiturgicalCalendarFrontend`                              | Admin calendar pages construct these object ids                         |

The blast radius is larger than #785's: these are **production calendar-editing grants**, not test-authoring
scopes, so the migration needs the same copy-then-prune discipline and a rollback story.

## Design questions

1. **Same `<rite>/<id>` shape as the test scopes?** #785 established
   `TestScopeResolver::qualify()` / `::parseQualifiedId()` with `/` as the separator. Reusing it verbatim
   keeps one format across the model; the helpers would want to move somewhere more neutral than
   `TestScopeResolver` if the data types adopt them.

2. **Does `wider_region` need qualifying?** Wider regions exist only in the Roman rite today, and unlike
   national calendars they are a transversal layer rather than a calendar. Qualify for uniformity, or leave
   bare and document why?

3. **Should `/data` gain a rite segment first?** The 404 above suggests `/data` rite-awareness is a
   prerequisite — there is little point authorizing `diocesan_calendar:ambrosian/lugano_ch` while no route
   can address that calendar. This may want to be sequenced as: `/data` rite segment → object-id change →
   tuple migration.

4. **Fix the existence-checker bug separately and first?** It is a small, self-contained fix that stops
   real permission loss, and it does not depend on any of the above. Splitting it out would let it ship
   immediately.

## Acceptance

- A grant can distinguish an Ambrosian `lugano_ch` from a Roman one.
- `ResourceExistenceChecker` reports all four Ambrosian dioceses as existing, and the reconciler no longer
  purges their tuples.
- Existing production grants are migrated without any editor losing access; the migration is idempotent and
  copy-only until every deployment runs merged code.
- The data resource types and the test scope types use one consistent id format.

## Related

- #785 — rite-scoped liturgical tests; introduced `rite_calendar_test` and the `<rite>/<id>` format for the
  test scopes, and explicitly deferred the data resource types to this issue.
- #767 — the parent issue for rite-scoped tests.
- #760 — the rite is echoed in the `/calendar` response.

JohnRDOrazio:
Closes #786.

Completes the OpenFGA tuple alignment started in #785, which did the *test* scopes and deferred the data
resource types here. Four ordered commits, in the sequence the issue proposes — the standalone bug fix first,
so it can ship or be cherry-picked on its own.

## 1. The live purge bug (`ff6ce656`)

`ResourceExistenceChecker::exists()` globbed only `JsonData::DIOCESAN_CALENDARS_FOLDER`, which derives from
`ROMAN_RITE_FOLDER`. The Ambrosian partition has its own constant the checker never consulted, so every
Ambrosian diocese answered "does not exist" — and `ResourceTuplePurgeReconciler` purges on exactly that
predicate. A legitimate editor grant on any of the four was revoked on any sweep.

The lookup now spans both rite partitions. It deliberately answers *"is this diocese defined under **any**
rite"* rather than a specific one: this decides what gets purged, so a false negative destroys a grant while
a false positive merely keeps a stale tuple around for the next sweep.

**A second bug, found by the same test:** `exists('diocesan_calendar', '')` returned `true`, because the
empty id made the glob match every nation directory. Ids are now validated against the diocese-id character
set before being interpolated into a glob pattern — which also stops a metacharacter in a tuple id matching
unintended directories.

Regression cover at both levels: the checker directly, and a reconciler sweep wired to the **real** checker.
Every existing reconciler test stubs it, which is exactly why none of them caught this.

## 2. `/data` rite-awareness (`aed9c8ef`)

`/data` could not address the Ambrosian calendars at all. `data` joins `calendar`, `events` and the root
route in `Router::extractRiteSegment()`, so the handler still sees the same 2-part shape and every
downstream count-based branch is untouched.

`JsonData` gains rite resolvers (`diocesanCalendarFileFor()` and friends) so the ~9 diocesan path sites ask
for "the diocesan file for this rite" rather than repeating an `if`. Only the diocesan tier needs them —
there are no `AMBROSIAN_NATIONAL_*` or `AMBROSIAN_WIDER_REGION_*` constants, because the Ambrosian rite has
neither tier. `RegionalDataParams` rejects a category the rite has no tier for, mirroring
`EventsParams::validateRiteCompatibility()`.

Verified against a server running this branch:

```text
/data/ambrosian/diocese/lugano_ch    200  (was 404)
/data/ambrosian/diocese/milano_it    200  (was 404)
/data/roman/diocese/lugano_ch        422  rite disagreement
/data/ambrosian/diocese/rotter_nl    422  rite disagreement
/data/ambrosian/nation/US            400  no such tier
/data/ambrosian/widerregion/Europe   400  no such tier
/data/diocese/rotter_nl              200  unchanged
/data/roman/diocese/rotter_nl        200  equivalent to bare
```

`data` is deliberately **not** added to `canonicalRiteUrl()`: that header advertises the explicit form for
cacheable read routes, and a `Link: rel="canonical"` on a `PUT` is noise. Easy to add if you'd rather have
the symmetry.

## 3. Rite-qualified object ids (`d25a2e6e`)

| Object type              | Id before   | Id after              |
|--------------------------|-------------|-----------------------|
| `diocesan_calendar`      | `lugano_ch` | `ambrosian/lugano_ch` |
| `national_calendar`      | `US`        | `roman/US`            |
| `wider_region`           | `Europe`    | `roman/Europe`        |
| `general_roman_calendar` | `temporale` | *(unchanged)*         |

`general_roman_calendar` keeps bare ids: they are `temporale`, `decrees` and missal editions — not calendars,
and Roman by construction.

**`wider_region` is qualified**, which the issue left open as a design question. The Ambrosian rite has no
wider regions, but they live in the same rite-partitioned tree as the other two, so exempting them would
qualify two of three sibling types and leave a carve-out to remember rather than a rule to state. Reverting
is a one-line change to the mapping plus its migration branch.

New `RiteScopedObjectId` holds the format; `TestScopeResolver`'s #785 helpers delegate to it, so a repository
no longer depends on a *test* service for its id format.

**No OpenFGA model change** — the object types are unchanged and only their ids move.

Also fixes a third instance of the stale valid-ids message, this time in `PermissionAdminHandler`: rejecting
`national_calendar:IT` reported `Valid ids: temporale, EDITIO_TYPICA_1970, …`.

## 4. Migration, runbook, and a Vatican purge bug (`e532fd10`)

`scripts/migrate-rite-data-tuples.php` rewrites live grants onto their qualified ids — copy-only by default,
`--prune` gated behind full rollout, idempotent. `member_nation` tuples are rewritten on **both** sides:
their user side is a `national_calendar:` object, not a `user:`.

Only a *diocese's* rite is inferred from the source tree; national calendars and wider regions are Roman by
definition. Writing that inference surfaced **a third live purge bug of the same shape**: the Vatican is
announced as a national calendar but is still served by the General Roman Calendar and has no
`nations/VA/VA.json` of its own yet, so

```text
ResourceExistenceChecker::exists('national_calendar', 'VA') => false
```

and the reconciler would purge any `national_calendar:VA` grant — even though the id is accepted as valid
when the grant is requested. `exists()` now treats VA as present, with a comment marking the special case for
removal once the Vatican gains its folder.

## Verification

Full suite **2495 green**, PHPStan level 10 clean, phpcs clean, markdownlint clean, `lint:openapi` valid with
zero warnings, plus the live `/data` probe above.

One note on the run: two `Routes/Auth/LoginRateLimitTest` cases failed once mid-session and passed on re-run
and in isolation — my own repeated probing had filled the live API's rate-limit bucket. Not a regression, but
worth knowing they are order- and traffic-sensitive.

## Follow-ups

- **Prune the superseded tuples** — this PR's and #785's — once every deployment runs merged code. Both
  runbook sections describe the step; they can be pruned independently.
- **#787** — `/tests` rite-awareness, which deferred its path-segment decision to whatever this change
  settled. `/data` now takes the segment, so that question has an answer to follow.
- **LiturgicalCalendarFrontend#459** covers the test editor; the **admin calendar pages** construct these
  data object ids and need the same treatment before the prune step.
- **Remove the Vatican special case** in `ResourceExistenceChecker` once `nations/VA/VA.json` exists.

Design spec: `docs/superpowers/specs/2026-08-14-rite-qualified-data-scopes-design.md`.

## Summary by CodeRabbit

* **New Features**
  * Added rite-aware calendar data access, including Ambrosian diocesan calendars and explicit Roman routing.
  * Calendar resource identifiers now support formats such as `roman/IT` and `ambrosian/lugano_ch`.
  * Added validation restricting national and wider-region resources to the Roman rite.
  * Improved resource detection, including Vatican national-calendar handling.

* **Documentation**
  * Added design guidance and an operational runbook for migrating existing calendar permissions.

* **Chores**
  * Added a dry-run migration utility with optional application and cleanup of legacy permission identifiers.

### DIFF
--- a/docs/ops/test-scope-migration-runbook.md
+++ b/docs/ops/test-scope-migration-runbook.md
@@ -389,7 +389,9 @@ Editors should retain access. Spot-check a diocesan grant of each rite:
 # Mirrors what OpenFgaClient sends: the bearer token when OPENFGA_API_TOKEN is set,
 # and the pinned model id — omitting authorization_model_id checks against the store's
 # latest model rather than the one the API is running.
-curl -s "$OPENFGA_API_URL/stores/$OPENFGA_STORE_ID/check" \
+# -fsS so an HTTP 4xx/5xx fails the command: plain `curl -s` exits 0 on an error
+# response, which would read as a successful verification.
+curl -fsS "$OPENFGA_API_URL/stores/$OPENFGA_STORE_ID/check" \
   -H 'Content-Type: application/json' \
   ${OPENFGA_API_TOKEN:+-H "Authorization: Bearer $OPENFGA_API_TOKEN"} \
   -d '{
@@ -399,8 +401,11 @@ curl -s "$OPENFGA_API_URL/stores/$OPENFGA_STORE_ID/check" \
           "relation": "editor",
           "object": "diocesan_calendar:ambrosian/lugano_ch"
         }
-      }'
+      }' | jq -e '.allowed'
 ```
+
+`jq -e` exits non-zero when `allowed` is false or absent, so a denied check fails the
+command too rather than printing `{"allowed":false}` and exiting 0.
 
 ### Step 3 — prune (later)
 
--- a/docs/ops/test-scope-migration-runbook.md
+++ b/docs/ops/test-scope-migration-runbook.md
@@ -334,3 +334,70 @@ allow-lists that still name it:
 - `jsondata/schemas/openapi.json`
 
 This step is **out of scope for the current change** and tracked as a follow-up.
+
+---
+
+## Follow-up migration — rite-qualified *data* resource scopes
+
+Issue #786 applied the same reasoning to the calendar **data** resource types that #767
+applied to the test scopes. A bare calendar id does not identify a calendar: the source
+tree is partitioned as `jsondata/sourcedata/rite/{rite}/calendars/...`, so `lugano_ch`
+could name an Ambrosian calendar or a Roman one, and a grant on the bare id would
+silently widen to cover whichever was added later.
+
+| Type                     | Object id before | Object id after       |
+|--------------------------|------------------|-----------------------|
+| `diocesan_calendar`      | `lugano_ch`      | `ambrosian/lugano_ch` |
+| `national_calendar`      | `US`             | `roman/US`            |
+| `wider_region`           | `Europe`         | `roman/Europe`        |
+| `general_roman_calendar` | `temporale`      | *(unchanged)*         |
+
+`general_roman_calendar` is untouched: its ids are `temporale`, `decrees` and missal
+editions, which are not calendars and are Roman by construction.
+
+**No model change is needed.** Unlike the test-scope migration above, the object *types*
+are unchanged — only their ids move — so there is no new type to add and no model version
+to upload.
+
+These are **production calendar-editing grants**, not test-authoring scopes. Treat the
+rollback story accordingly: the unqualified ids stay valid in every PHP allow-list until
+the prune step, so a rollback to pre-#786 code keeps authorizing.
+
+### Step 1 — copy the tuples
+
+```bash
+# Dry run first — prints every tuple that would be rewritten.
+php scripts/migrate-rite-data-tuples.php
+
+# Apply.
+php scripts/migrate-rite-data-tuples.php --apply
+```
+
+National calendars and wider regions exist only in the Roman rite, so their rite is a
+constant. A diocese's rite is inferred from the rite-partitioned source tree. A diocese id
+defined under two rites, or under none, is reported and left untouched, and the run exits
+with status `2` so the anomaly is not lost in CI output.
+
+`member_nation` tuples are rewritten on **both** sides: their user side is a
+`national_calendar:` object rather than a `user:`, so it needs qualifying too.
+
+### Step 2 — verify
+
+Editors should retain access. Spot-check a diocesan grant of each rite:
+
+```bash
+curl -s "$OPENFGA_API_URL/stores/$OPENFGA_STORE_ID/check" \
+  -H 'Content-Type: application/json' \
+  -d '{"tuple_key":{"user":"user:<sub>","relation":"editor","object":"diocesan_calendar:ambrosian/lugano_ch"}}'
+```
+
+### Step 3 — prune (later)
+
+Only once **every** deployment runs post-#786 code:
+
+```bash
+php scripts/migrate-rite-data-tuples.php --apply --prune
+```
+
+Then the unqualified ids can be rejected outright rather than merely superseded. The same
+step for the #785 test scopes is described above; the two can be pruned independently.
--- a/docs/ops/test-scope-migration-runbook.md
+++ b/docs/ops/test-scope-migration-runbook.md
@@ -386,9 +386,20 @@ with status `2` so the anomaly is not lost in CI output.
 Editors should retain access. Spot-check a diocesan grant of each rite:
 
 ```bash
+# Mirrors what OpenFgaClient sends: the bearer token when OPENFGA_API_TOKEN is set,
+# and the pinned model id — omitting authorization_model_id checks against the store's
+# latest model rather than the one the API is running.
 curl -s "$OPENFGA_API_URL/stores/$OPENFGA_STORE_ID/check" \
   -H 'Content-Type: application/json' \
-  -d '{"tuple_key":{"user":"user:<sub>","relation":"editor","object":"diocesan_calendar:ambrosian/lugano_ch"}}'
+  ${OPENFGA_API_TOKEN:+-H "Authorization: Bearer $OPENFGA_API_TOKEN"} \
+  -d '{
+        "authorization_model_id": "'"$OPENFGA_MODEL_ID"'",
+        "tuple_key": {
+          "user": "user:<sub>",
+          "relation": "editor",
+          "object": "diocesan_calendar:ambrosian/lugano_ch"
+        }
+      }'
 ```
 
 ### Step 3 — prune (later)
--- a/docs/superpowers/specs/2026-08-14-rite-qualified-data-scopes-design.md
+++ b/docs/superpowers/specs/2026-08-14-rite-qualified-data-scopes-design.md
@@ -54,6 +54,13 @@ of this id exist under any rite", and only returns false when none does.
 `national_calendar` and `wider_region` remain Roman-only lookups (no Ambrosian tier exists), tolerating an
 optional `roman/` prefix once step 3 lands.
 
+**A second false negative, found while writing the migration.** The Vatican is announced as a national
+calendar but is still served by the General Roman Calendar and has no `nations/VA/VA.json` of its own yet,
+so `exists('national_calendar', 'VA')` returned false and the reconciler would purge any
+`national_calendar:VA` grant — the same silent revocation, and the id is accepted as valid when the grant is
+requested. `exists()` special-cases VA, with a comment marking the case for removal once the Vatican gains
+its folder.
+
 ## 2. `/data` rite-awareness
 
 `Router::extractRiteSegment()` already strips a leading rite segment for `calendar`, `events` and the root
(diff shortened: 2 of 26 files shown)


JohnRDOrazio:
Both review findings verified against the code and applied in `a5f3659b`, along with the codecov patch gap.

**Check example missing auth and model id — valid.** `OpenFgaClient` sends `Authorization: Bearer` when `OPENFGA_API_TOKEN` is set (`OpenFgaClient.php:433`) and `authorization_model_id` on every call (`:125`, `:161`, `:208`, `:305`). An operator following the runbook would have got a 401 where a token is configured, or silently checked against the store's latest model rather than the one the API is pinned to. The example now mirrors what the client actually sends, with the token header conditional so it stays copy-pasteable when no token is set.

**`general_roman_calendar` missing from the object_type enums — valid, and broader than reported.** It is in `AccessRequestRepository::VALID_OBJECT_TYPES` but was absent from the schema. You flagged two enums; there were **four**. The same constant also gates the `object_type` query filter via `PermissionAdminHandler:318`, so `GET /admin/permissions` and `GET /admin/permissions/check` documented a narrower filter set than they accept. All four now list it.

**Coverage:** 32 uncovered added lines → 12.

- `JsonDataRiteResolversTest` pins each rite resolver to the case it must return — these decide which partition `/data` reads and writes, so a wrong one would silently send an Ambrosian write into the Roman tree.
- `RiteScopedObjectIdTest` covers qualify/parse/calendarId, including that the same calendar id under two rites yields distinct object ids (the point of the format) and that unqualified legacy ids pass through, which the migration window depends on.
- `RegionalDataHandlerTest` gains diocesan reads for both rites, the i18n read, and both directions of the rite-mismatch rejection.

What remains uncovered is deliberate: the diocesan create/update/delete path sites, which have no diocesan write tests to begin with and appear in the diff only because a constant became a resolver; and four `Router` glue lines that only a full route dispatch reaches, which are covered in CI where the `Routes/*` suite runs against an instrumented server.

Suite 2520 green, PHPStan level 10 clean, phpcs clean, markdownlint clean, `lint:openapi` valid.

JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
Second pass: one applied, one declined — and a note on the risk rating.

**`curl -s` masks HTTP errors — valid, applied in `938fb953`.** Good catch: a failed Check would have exited 0 and read as a successful verification. Now `-fsS`, and I went one further and piped through `jq -e '.allowed'`, because `-fsS` alone still exits 0 on a *denied* check — `{"allowed":false}` is a perfectly good HTTP 200. Both failure modes now fail the command.

**`#[Group('slow')]` on the three new test classes — declined**, same as the equivalent finding on #785. None of them declares a group today, and all 43 of their tests run in **208 ms together**:

```text
$ vendor/bin/phpunit phpunit_tests/Enum/JsonDataRiteResolversTest.php \
    phpunit_tests/Services/RiteScopedObjectIdTest.php \
    phpunit_tests/Handlers/RegionalDataHandlerTest.php
Time: 00:00.208  OK (43 tests, 74 assertions)
```

`CLAUDE.md` reserves `slow` for tests with *measurable* runtime cost — multi-year calendar calculations, rate-limit window waits, full-corpus schema validation — and explicitly notes that integration tests are not automatically slow. Marking these would misclassify them and drop them from `composer test:quick` for no benefit. The guideline being cited is about *how* to declare `slow` when a test is slow, not which classes should be.

**On the 🟠 High merge-risk rating:** I don't think it reflects what is actually outstanding. Both 🟠 Major findings from the first pass are marked *"✅ Addressed in commit a5f3659"* in this same review, and the review body reports **"Actionable comments posted: 2"** — both 🟡 Minor, one of which is the misclassification above. The summary prose ("incorrect authorization data placement or purging, unsafe filesystem probing, broken client validation, unreliable migration verification") maps one-to-one onto the four findings *ever* raised on this PR, including the two already fixed, rather than the two still open. Happy to be shown otherwise if there's a fifth finding I have missed.

All checks green on `a5f3659b` including **codecov/patch**, which was the one that failed on the first push.

JohnRDOrazio:
@coderabbitai review

PR #791, opened 2026-08-16
chore: delete the dead scripts/openfga-model.additive.json

JohnRDOrazio:
Deletes `scripts/openfga-model.additive.json`.

## Why it is dead

The OpenFGA authorization model moved to `cdcf-infra` (`auth/models/LiturgicalCalendar.json`) on 2026-08-04. Since then this file has been applied by **nothing**: no composer script, no CI workflow, no PHPUnit test, and not even `scripts/setup-openfga.sh`, whose own header says it "reads store/model IDs only […] the authorization model itself is owned by cdcf-infra […] this script no longer creates or updates the model".

`docs/ops/rbac-create-governance-runbook.md` already asserts, in its opening note, that *"neither `scripts/openfga-model.json` nor `scripts/openfga-model.additive.json` exists in this repo any more"*. That sentence was simply false. This PR makes it true.

## Why it was not merely stale

[#785](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/pull/785) added the new `rite_calendar_test` object type **here**, instead of opening the required `cdcf-infra` PR. The file looks exactly like a live surface — well-formed model JSON, sitting next to the operational scripts — so the edit read as done. The result:

- `rite_calendar_test` shipped in `ResourceAdminService::VIEWER_OBJECT_TYPES` / `TEST_OBJECT_TYPES`, but exists in no deployed model.
- `listObjects` on an unknown type returns `400 type_not_found` → `RuntimeException`.
- The `catch (\RuntimeException)` in `ResourceAdminService` wraps the **whole loop** and returns `array_fill_keys(VIEWER_OBJECT_TYPES, [])`, so one unknown type empties *every* type's scope list for *every* user — dashboard card gating collapses globally, not just for rite-scoped tests.
- Frontend e2e red: `rbac/07-dashboard-card-scoping:143`, `rbac/12`, `rbac/13:153`, `rbac/13:201`, `rbac/15-dashboard-tests-card-matrix:36`.

The type is being added where it counts in **[CatholicOS/cdcf-infra#31](https://github.com/CatholicOS/cdcf-infra/pull/31)**. That PR, not this one, is what actually unblocks the e2e specs; this one removes the trap.

The file had also drifted from the deployed model in ways that would silently reintroduce retired surface if anyone *did* apply it: it still carries the `test_definition` type and a `deleter` relation on every type — both of which `authz/openfga-expectations.json` now explicitly forbids (`forbidden_types`, `forbidden_relations: {"*": ["deleter"]}`).

## Reference check before deleting

Full-repo grep for the filename (excluding `vendor/`, `.git/`) found it referenced only from documentation, plus one comment:

| Reference | Kind | Action |
| --- | --- | --- |
| `docs/ops/test-scope-migration-runbook.md:292` | **Operative** runbook step | Updated — Step 1 no longer tells an operator to edit the file, and records why that instruction was dangerous. |
| `.github/deploy/rsync-exclude.txt:83` | Comment only (no exclude rule) | Updated — the justification for shipping `scripts/` no longer claims an `openfga-model*.json` is among the things shipped. |
| `docs/ops/rbac-create-governance-runbook.md` | Historical record, under a note saying it is retained unchanged | Left alone. |
| `docs/superpowers/plans/*`, `docs/superpowers/specs/*` | Dated design/plan records | Left alone. |

Nothing executable referenced it, so deletion breaks nothing. The `superpowers/` docs are records of decisions as they were made, not instructions — the same reason `rbac-create-governance-runbook.md` keeps its original steps verbatim beneath a note saying the world has moved on.

`composer lint:md` passes (CaptainHook pre-commit, not bypassed).

## Summary by CodeRabbit

* **Documentation**
  * Updated deployment guidance to clarify that authorization models are managed and uploaded through the infrastructure repository.
  * Updated the test-scope migration runbook with the revised authorization-model deployment process.
* **Chores**
  * Removed the repository-managed authorization model file.
  * Clarified deployment exclusions while retaining server-side operational scripts.


JohnRDOrazio:
## CodeRabbit finding declined, with reasoning

**Finding:** add an explicit `scripts/openfga-model*.json` exclusion to `.github/deploy/rsync-exclude.txt`.

**Declined — it would invert the intended outcome.** That file documents its own semantics at lines 11-16:

> GOTCHA (consequence of #2): adding a pattern here does NOT clean up copies already on the server from earlier
> deploys — the new pattern protects them from `--delete`, fossilizing them.

So the two options are:

- **No exclude rule (current PR):** the file is gone from the repo, so it is absent from the runner checkout, and
  `rsync --delete` removes the stale copy from the server on the next deploy. The cleanup happens by itself.
- **With an exclude rule:** that same stale server copy becomes protected from `--delete` and fossilizes
  indefinitely.

Fossilizing it is the specific failure this PR exists to end. `scripts/openfga-model.additive.json` was declared
dead by `docs/ops/rbac-create-governance-runbook.md:16-19`, yet it looked live enough that API #785's model change
was written into it instead of into a cdcf-infra PR — which is why `rite_calendar_test` never reached the deployed
model and several RBAC e2e specs went red.

Verified on this branch: no `openfga-model*.json` remains tracked or on disk (`git ls-files scripts/ | grep
openfga-model` → empty), so the rewritten comment block is accurate as written.

No change made.

Issue #790, opened 2026-08-16
Authorization: PATCH /tests authorizes against the stored scope but writes the payload's applies_to


JohnRDOrazio:
## Summary

`PATCH /tests/{rite}/{name}` authorizes against the scope resolved from the **stored file**, but then writes the
**payload's** `applies_to` — with no check that the resulting scope is one the caller is permitted to write. A
scoped `test_editor` can therefore move a test into a scope they do not hold.

This predates #787 and came in with #785; #787 only added the rite dimension, so the same gap existed with bare
object ids. It was found while reviewing #787 and is filed separately rather than fixed there, to keep an
authorization fix out of an addressing refactor.

## Mechanism

`src/Http/Middleware/OpenFgaAuthorizationMiddleware.php`, `forTestScopes()`:

```php
$resolved = $resolver->resolve($rite, $testId);
if (
    $resolved === null
    && strtoupper($request->getMethod()) === 'PUT'   // <-- payload fallback is PUT-only
    && TestScopeResolver::isSafeName($testId)
) {
    $resolved = $resolver->resolveFromPayload($request->getParsedBody());
}
```

For `PATCH` the file exists, so the scope always comes from the **stored** `applies_to`. The payload fallback is
deliberately gated to `PUT`, where it is correct ("the scope that authorizes the create is the scope the created
resource will carry").

`src/Handlers/TestsHandler.php::handlePatchRequest()` then validates the schema, the rite/path agreement, the
`scope` echo, the name match and file existence — but never checks the **new** `applies_to` against the caller's
permissions before `writeTestToDisk()`.

## Reproduction

1. Grant a user `editor` on `national_calendar_test:roman/IT`.
2. `PATCH /tests/roman/SomeItTest` with a body whose `applies_to` is `{"rite": "roman", "national_calendar": "US"}`.
3. Authorization runs against `national_calendar_test:roman/IT` (the stored scope) and passes.
4. The write succeeds; the stored test now resolves to `national_calendar_test:roman/US`.

The operation is one-way: the editor cannot subsequently edit the test, because authorization then resolves to the
new scope. The effect is pushing content into another scope, not gaining persistent access to it.

## Impact

Moderate. Requires an authenticated `test_editor` holding some scope. Liturgical tests drive CI and validation
rather than published calendar output, so this is an integrity issue rather than a data-exposure one. There is no
audit signal that it happened.

## Possible fix

[...]

PR #792, opened 2026-08-16
feat(tests): make /tests rite-aware (#787)

### LINKED ISSUE #787: /tests is not rite-aware: the rite is not part of a test's identity, and the collection cannot be filtered by it

JohnRDOrazio:
Closes #787.

Completes the rite-awareness arc: #785 made the rite a required property *inside* each test and rite-qualified
the test authorization scopes; #786 settled the path-segment convention on `/data`. #787 deferred its own segment
decision to whatever #786 chose, and this is that decision applied to `/tests`.

## What changes

The test corpus moves from a flat `jsondata/tests/{name}.json` to `jsondata/tests/{rite}/{name}.json`, so test
names are unique **within a rite** rather than globally. A Roman `StIgnatiusOfLoyolaTest` can now coexist with the
Ambrosian one that already ships — with no naming convention to remember and no way to confuse them.

| Path | Result |
| ------------------------------ | ------------------------------------- |
| `GET /tests`                   | collection, **all rites** (unchanged) |
| `GET /tests/{rite}`            | collection for that rite (new)        |
| `/tests/{rite}/{test_name}`    | single test — GET/PUT/PATCH/DELETE    |
| `GET /tests/{test_name}`       | **400** — rite segment required       |

`tests` deliberately does **not** join `Router::extractRiteSegment()`. That helper resolves an absent segment to
Roman, which is right for `/calendar`, `/events` and `/data` — each addresses a single calendar, so "no rite
stated" can only mean "the default one". `/tests` has a collection whose historical meaning is "every test
regardless of rite", so it needs a third state: `extractTestsRite()` returns `?Rite`, where `null` means all rites.

`canonicalRiteUrl()` does not gain `tests` either — there is no bare form left to canonicalise.

Every test in both responses now carries a read-only, server-computed `scope` (`{object_type, object_id}`), so
clients stop re-deriving it. The frontend's `deriveScope()` hand-copy had already drifted twice in one release.

## ⚠️ This is a hard break

`GET /tests/{test_name}` was valid and is now a 400. There is no deprecation window. Both clients need to ship
with or before this:

- LiturgicalCalendarFrontend#459 — `admin-tests` editor
- UnitTestInterface#39 — the test runner

## Notable decisions

- **A `scope` echoed back on write is accepted, not rejected.** The first design 422'd any payload carrying
  `scope`. Both client analyses pushed back independently and were right: no legitimate client *originates* a
  scope value, so the only realistic source is a benign echo of a GET response, and rejecting it would break the
  ordinary load-edit-save cycle. A **disagreeing** scope is still a 422, and an accepted echo is stripped before
  persisting. It is silent divergence that must be impossible, not the echo.
- **A path rite segment contradicting the body's `applies_to.rite` is a 422.** The directory is the address,
  `applies_to.rite` is the content, and the handler refuses to let them diverge.
- **No OpenFGA model change and no tuple migration.** #785 already rite-qualified the test scopes, so unlike #786
  this carries none of that weight.
- **`TestTupleMigration` now refuses ambiguity rather than guessing.** It previously took the first rite partition
  that resolved; since this PR makes name collisions legal, and the migration script deletes the legacy tuple after
  rewriting it, that could have silently moved a grant to the wrong scope. It now reports and skips, matching
  `migrate-rite-data-tuples.php`.

## Operator note

`docs/ops/test-scope-migration-runbook.md` gains a step for a real hazard: any test created through
`PUT /tests/{name}` since #785 lives flat on a deployed host's volume, and after this ships becomes invisible to
`GET /tests`, unaddressable and undeletable through the API, with its FGA tuples still live. The runbook now
documents how to detect and relocate those files. Its `PATCH` verification steps were also re-pathed — two of them
had started failing, and a third was passing for the wrong reason.

## Verification

- `composer test:quick` — 2352 tests, 1 failure: `ExecuteValidationTest.php:129`, the documented baseline when the
  API runs on a non-8000 port (the Docker WebSocket container cannot reach it). Run against a server booted from
  this branch, because `phpunit_tests/Routes/*` otherwise hits `:8000`, which serves a different checkout.
- `composer analyse` (PHPStan level 10, 314 files), `composer lint`, `composer lint:md`, `composer lint:openapi` —
  all clean.
- Each of the six tasks passed its own spec-and-quality review, then the whole branch passed a final review that
  traced every verb end to end. Several findings were caught by mutation-testing tests that looked like coverage
  and were not.

## Known and deliberately out of scope

- **#790** — `PATCH /tests` authorizes against the stored file's scope but writes the payload's `applies_to`, with
  no check that the new scope is one the caller holds. Pre-existing from #785; filed separately rather than mixed
  into an addressing refactor.
- A calendar filter on the collection (`?national_calendar=US`). Both clients re-implement that selection
  client-side — the same duplication argument that motivated `scope` — but the rite dimension is what #787 exists
  to settle.

## Summary by CodeRabbit

- **New Features**
  - Added rite-specific test endpoints for Roman and Ambrosian rites, with an all-rites index retained.
  - Test responses now include resolved authorization scope information.
  - Added validation for rite and scope mismatches during test creation and updates.
  - Added extensive calendar and liturgical test coverage.

- **Documentation**
  - Updated API examples, OpenAPI descriptions, and migration runbooks for rite-qualified paths and legacy data handling.

- **Bug Fixes**
  - Improved migration reporting for missing or ambiguous test mappings.


JohnRDOrazio:
## CodeRabbit triage: 8 applied, 6 declined

Every finding was verified against the current code before acting; declined ones are listed with reasons rather
than silently dropped.

### Applied

| Finding | What was actually wrong |
| --- | --- |
| Runbook "Detect" section | The `find` command errors when run from the working directory the surrounding text tells the operator to be in, printing nothing — indistinguishable from the documented "no hazard" result. Fixed the cwd instruction and the empty-output caveat. |
| Runbook steps 4b/4c | The example PATCH body (`{"missal":"IT_1983"}`) fails `LitCalTest.json` unconditionally — missing every required field, and `missal` is not a property. Those steps could never have returned 200/204. Replaced with a valid body and corrected the expected status. |
| Design spec routing table | Last row said 3+ path segments is a 400; it is a **405** — `AbstractHandler::validateRequestMethod()` fires before the handler's own guard. The OpenAPI text had been corrected earlier but the spec table was missed. |
| `TestsHandlerTest` Ambrosian-PUT test | The Roman fixture path was asserted absent but never registered for cleanup, so a failing assertion could leave a stray fixture in the shipped corpus. |
| `migrate-test-tuples.php` + runbook | The unresolved-ID summary heading said only "test JSON file not found", but the script can now also report `ambiguous` (a name defined under two rites). Per-tuple output was already correct; the heading was not. |
| `openapi.json` PUT/PATCH | Documented 422 but not 400, though `validatePayloadAgainstTestSchema()` produces a reachable 400. Added the 400 response to both. |
| `testCollectionCarriesTheResolvedScope` | Asserted *every* returned test carries the Ambrosian scope — true only because that partition holds exactly one file. Now asserts on the named test. |
| PUT/PATCH docblocks *(follow-on)* | Surfaced while checking the runbook finding: both said schema-validation failure returns 422. It returns **400** (`ValidationException` → `StatusCode::BAD_REQUEST`). The 422s come from the semantic guards afterwards. |

### Declined

- **Plan-document `collectTests()` snippet.** The discrepancy is real, but a plan is a point-in-time record of what
  was *planned*. Editing it to match what shipped destroys its value as a record.
- **`#[Group('slow')]` on `OpenFgaAuthorizationMiddlewareTest` and `LitTestRunnerTest`.** Measured: 33 tests in
  **0.020s** and 14 tests in **0.019s**, both fully mocked. This repo reserves `slow` for measurable runtime cost;
  mislabelling these would silently drop 47 tests from `composer test:quick`.
- **A throwing rite accessor replacing `$this->rite ?? Rite::default()`.** The idiom appears ~5x in the file, the
  null-rite case is already blocked by a guard in `handle()`, and this is a behaviour-adjacent change to code that
  has passed full review. Style preference, not a latent bug.
- **Reusable `TestsNamePathParam` component**, **`TestsRitePathParam` `$ref`ing `CommonDef.json`**, and
  **`RouterPipelineTest` endpoint-level coverage** — polish with no defect behind them.

### Validation

`composer analyse` 314/314 · `composer lint` clean · `composer lint:md` clean · `composer lint:openapi` valid,
0 warnings (the 11 pre-existing scoped ignores untouched) · `TestsHandlerTest` 31/31.

Commits: `2f9af8f0`, `314ae703`

JohnRDOrazio:
## CodeRabbit finding applied — and it led to three more

**Finding:** the ambiguous-tuple remediation should write the scoped replacement first, preserving the original
tuple's `user` and `relation`, and delete the legacy `test_definition:{name}` tuple only after that write succeeds.

**Valid, and slightly understated.** The instructions did not mention deleting the legacy tuple *at all* — so an
operator following them left the migration half-done — and never said to preserve `user` and `relation`. The
script's own apply path already does exactly the right thing
(`scripts/migrate-test-tuples.php:137-147`, "write new tuple first, then remove old", copying `user`/`relation`
verbatim); the manual path just didn't mirror it.

The section now spells out both steps, in order, with the reason the order matters: delete-first leaves the grantee
holding no grant at all in between, and a failed write revokes real access with nothing recorded to restore it
from. It also records the consequence of stopping halfway — skipping the delete is tolerable while the
`test_definition` type is retained, but skipping the write means the grant vanishes when that type is pruned.

### Three adjacent bugs found while verifying

- **Step 3b** described the written tuple as `national_calendar_test:IT`. Those ids have been rite-qualified since
  #785 — it is `national_calendar_test:roman/IT`.
- **The dry-run sample output** used bare ids, the superseded
  `general_roman_calendar_test:general_roman_calendar`, and test names (`IT_2024`, `ROMA_DIOCESE_2024`) that
  **cannot satisfy** `LitCalTest.json`'s name pattern — so no operator could ever have seen that output. All five
  sample names now validate against the actual regex (checked programmatically, not by eye — `IT_2024` confirms
  as invalid).
- **The sample showed only the "no test file found" unresolved line**, never the `ambiguous` case the surrounding
  prose documents. Both now appear, with the script's exact wording.

Verified afterwards: no bare test-scope ids remain anywhere in the runbook.

`composer lint:md` clean. Docs only, no behaviour change.

Commit: `a2b84ce2`

PR #796, opened 2026-08-17
fix(data): report an i18n/locales mismatch as 422, not 500

JohnRDOrazio:
Fixes the API half of [LiturgicalCalendarFrontend#462](https://github.com/Liturgical-Calendar/LiturgicalCalendarFrontend/issues/462). The frontend half is [LiturgicalCalendarFrontend#464](https://github.com/Liturgical-Calendar/LiturgicalCalendarFrontend/pull/464) — independent of this, either can land first.

## The bug

`i18n` must key exactly the locales `metadata.locales` announces. JSON Schema cannot express "the keys of this object equal the values of that array", so the RegionalData models enforce it and signal a breach with a plain `\ValueError`:

```php
if (implode(',', $i18nProps) !== implode(',', $this->metadata->locales)) {
    throw new \ValueError('keys of i18n parameter must be the same as the values of metadata.locales');
}
```

Nothing on the write path caught it. `RegionalDataHandler`'s three `fromObject()` calls sat bare, so the error escaped and became a 500 — blaming the server for what is squarely a client-data problem. The two sibling checks immediately above each of those calls already throw `UnprocessableContentException` (missing `i18n`, empty `litcal`), so 422 is plainly the intended shape for this family.

It was reachable in practice: the frontend's diocesan editor submitted exactly that payload whenever a save beat its secondary-locale translation fetch.

## What changed

`\ValueError` → `UnprocessableContentException` at all three write sites, via a small `payloadValueError()` helper carrying the rationale.

**Why in the handler and not in the models:** the same DTOs load calendars from disk. There a locales/i18n mismatch is *our* bug in stored source data, not the caller's, and must stay a server error. Only the request path may reinterpret it as unprocessable content. Moving the exception type into `validateTranslations()` would have conflated the two.

**All three categories**, not just diocesan: `DiocesanData`, `NationalData` and `WiderRegionData` each have their own `fromObject()` call and the identical model-side check. The defect was never diocese-specific — only its discovery was.

`loadTranslations()` (the other `validateTranslations()` caller) has no callers anywhere in `src/`, so the constructor path is the whole surface.

## Test plan

- [x] Two new tests in `RegionalDataHandlerTest`, diocesan and national, each declaring two locales with `i18n` covering one. They assert the **message**, not merely the exception type: several unrelated conditions on this path also raise `UnprocessableContentException`, so a bare type assertion could pass for the wrong reason.
- [x] **Confirmed to fail against the unpatched handler** with `Failed asserting that exception of type "ValueError" matches expected exception ...UnprocessableContentException` — so these are genuine regression tests, not tautologies.
- [x] `composer test:quick` — 2354 tests, 172818 assertions, **no failures**. 5 errors, all in `phpunit_tests/WebSocket/ExecuteUnitTestTest.php`, which needs `composer ws:start`; unrelated to this path.
- [x] `composer lint` (phpcs) and `composer analyse` (PHPStan level 10) — clean.

Wider-region has no payload fixture in the test class, so it is covered by inspection rather than by a test; the docblock says so explicitly rather than implying coverage it does not have.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for regional data updates to ensure internationalization keys match all declared locales.
  - Invalid diocesan, national, and wider-region payloads now return clear, actionable validation errors instead of server errors.
  - Added coverage for invalid locale configurations across supported regional data update scenarios.
  - Standardized validation responses for malformed regional data submissions, providing more consistent feedback when updates cannot be processed.

### DIFF
--- a/phpunit_tests/Handlers/RegionalDataHandlerTest.php
+++ b/phpunit_tests/Handlers/RegionalDataHandlerTest.php
@@ -696,6 +696,52 @@ final class RegionalDataHandlerTest extends AbstractHandlerTestCase
     }
 
     /**
+     * A payload whose `i18n` covers fewer locales than `metadata.locales` announces must be
+     * rejected as unprocessable content, NOT crash with an uncaught \ValueError (issue #462).
+     *
+     * JSON Schema cannot express "the keys of this object equal the values of that array", so
+     * the invariant is enforced in the RegionalData model layer, which signals a breach with a
+     * plain \ValueError. Nothing on the write path caught it, so a client that submitted a
+     * short `i18n` — which the diocesan editor did whenever a save beat its secondary-locale
+     * translation fetch — got a 500 blaming the server for its own bad request.
+     *
+     * Covered here for the diocesan and national categories; wider-region reaches the same
+     * model check through the identical `fromObject()` write path.
+     */
+    public function testPutDiocesanCalendarRejectsI18nNarrowerThanDeclaredLocales(): void
+    {
+        $payload                        = self::diocesanPayloadWithColor('ambrosian', 'morello', 'novara_it', 'Diocesi di Novara');
+        $payload['metadata']['locales'] = ['it_IT', 'la_VA'];
+        // `i18n` still carries it_IT alone — exactly the mismatch the model rejects.
+
+        $this->expectException(UnprocessableContentException::class);
+        // Assert on the message, not merely the type: many unrelated conditions on this path
+        // also raise UnprocessableContentException, so a bare type assertion could pass for
+        // the wrong reason.
+        $this->expectExceptionMessage('keys of i18n parameter must be the same as the values of metadata.locales');
+
+        ( new RegionalDataHandler(['diocese', 'novara_it']) )
+            ->handle($this->requestFor('PUT', '/data/diocese/novara_it', [], $payload));
+    }
+
+    /**
+     * The national category has its own `fromObject()` call, so it needs its own guard — the
+     * defect was never diocese-specific.
+     */
+    public function testPutNationalCalendarRejectsI18nNarrowerThanDeclaredLocales(): void
+    {
+        $payload                        = self::mtNationalCalendarPayload();
+        $payload['metadata']['locales'] = ['en_MT', 'it_IT'];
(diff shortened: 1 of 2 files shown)


JohnRDOrazio:
Went through the three items. One was real and is now fixed; the two inline suggestions I'm declining, with reasons.

### ✅ Codecov — patch coverage (fixed in 480e1997)

Real, and it pointed at exactly the gap the PR body admitted to: wider-region was covered "by inspection" only, so its `catch` never executed. Added `testPutWiderRegionCalendarRejectsI18nNarrowerThanDeclaredLocales` with a fixture modelled on the shipped `wider_regions/Europe/Europe.json`.

Two schema requirements worth recording, since both bit on the first attempt: `national_calendars` is a `name => ISO code` map and is schema-required, and `makePatron` requires `metadata.url_lang_map` alongside `url`.

Confirmed it fails against the pre-fix handler with the uncaught `ValueError`, so it is a real regression test. All three categories now have an executing test.

### ❌ `metadata.timezone` → `Europe/Vatican` — declining

`Europe/Rome` is correct here and `Europe/Vatican` would be wrong:

- The fixture's dioceses are **Novara** and the **Diocese of Rome** — Italian dioceses, not the Holy See. `Europe/Vatican` is a distinct IANA zone for Vatican City.
- Every diocese in the shipped Italian source data uses `Europe/Rome`; `grep` over `jsondata/sourcedata/rite/` finds no `Europe/Vatican` in any diocesan calendar. Changing the fixture would make it diverge from the data it stands in for.
- The schema only requires "a timezone identifier from the IANA Time Zone Database", so both parse — this is a correctness question, not a validation one.

It is also a pre-existing shared fixture this PR does not touch, used by the rite-colour tests, so changing it is out of scope either way.

### ❌ `#[Group('slow')]` on the new tests — declining

- `RegionalDataHandlerTest` has **zero** `Group` attributes on any of its tests, including the rite-colour tests immediately above these, which follow the same handler-invocation shape. Adding it to only the new tests would be inconsistent with the file.
- They are not slow: all three run in **0.19s** total.
- `slow` is excluded by `composer test:quick`, which is the suite that runs most often. Grouping a regression test out of the fast suite is the opposite of what it is for.

PR #797, opened 2026-08-17
fix(auth): authorize the union of scopes when PATCH re-scopes a test (#790)

### LINKED ISSUE #790: Authorization: PATCH /tests authorizes against the stored scope but writes the payload's applies_to

JohnRDOrazio:
Closes #790.

## The gap

`PATCH /tests/{rite}/{name}` authorized against the scope resolved from the **stored file**, then wrote the
**payload's** `applies_to` — with no check that the resulting scope was one the caller may write.

`OpenFgaAuthorizationMiddleware::forTestScopes()` gated its `resolveFromPayload()` fallback to `PUT` only, and
`TestsHandler::handlePatchRequest()` had only consistency guards. So an editor on
`national_calendar_test:roman/IT` could PATCH an IT test into `roman/US` — pushing content into a scope they do
not hold. One-way: they could not edit it afterwards, since authorization then resolved to the new scope.

Pre-existing since #785; found while reviewing #787 and filed separately rather than mixed into an addressing
refactor.

## The fix — authorize the union

A PATCH that changes `applies_to` now requires `editor` on **both** the stored scope and the payload-derived
scope. This mirrors the PUT rule: you may write only where you may write. Re-scoping stays possible; it just
requires rights on both ends.

Implemented as a second, PATCH-only middleware instance (`forTestScopePayloadTarget()`) piped after
`forTestScopes()`. `process()` is unmodified — both instances run, both can deny, and the `admin` bypass applies
to both consistently.

The alternative — forbidding `applies_to` changes on PATCH outright — was considered and rejected: it would have
forced a delete-and-recreate for a legitimate operation.

## Entangled change

`assertPayloadScopeAgrees()` compared an echoed `scope` against the **stored** file. Once re-scoping is allowed
that is incoherent: a caller who re-scopes *and* correctly re-derives `scope` would 422, while one echoing the
now-obsolete scope would be accepted. It now resolves from the payload — what the test will resolve to after the
write — which also removed a disk read. All four cases are pinned by tests: unchanged echo accepted, new-scope
echo accepted, stale echo 422, create unchanged. `scope` is still never persisted.

## Two hardening changes found in review

- **`/tests` writes are now JSON-only.** The route accepted `JSON, YAML, FORMDATA`, but `JsonBodyParserMiddleware`
  populates the parsed body only for `application/json` — so after the union fix a YAML PATCH would have failed
  closed with a 403. `openapi.json` already declared only `application/json` for these bodies, so the Router was
  out of sync with its own documented contract; this aligns them. A monorepo-wide search (frontend,
  UnitTestInterface, components libraries, docs, wiki) found zero clients sending YAML or form bodies to `/tests`.
  YAML *responses* are unaffected — only request content types changed.
- **`resolveFromPayload()` now fails closed.** It previously could not fail for any array input, defaulting to
  `rite_calendar_test:<rite>`, so a payload with no `applies_to` was authorized against that lenient default and
  only a handler 422 stopped the write. It now returns `null`. `resolve()` deliberately keeps its lenient default
  — it runs over stored files including legacy ones written before `applies_to.rite` was required, and a test pins
  that behaviour so the distinction cannot be lost by accident.

## Verification

Both rounds were mutation-tested independently by a reviewer, not just by the implementer:

- Emptying the payload-target relation map → 6 failures including the exploit test.
- Reverting the payload resolver to stored-file resolution → caught, including by the union test, which was
  strengthened to assert *which* object ids were checked (it previously asserted only that `check()` ran twice,
  and was blind to exactly this mutant).
- Making `resolve()` strict → caught by the pre-existing lenient-default test, confirming the legacy path is
  actively guarded.

`composer analyse` (PHPStan level 10, 314 files), `composer lint`, `composer lint:openapi` (0 warnings) all clean.
Targeted suites 114/114; full suite excluding slow/golden-master, 2359 tests, 0 failures.

## Known follow-up

`Router::route()` is `never`-returning and not unit-testable, so the content-type change itself has no automated
coverage — the wiring was verified manually and the handler's reaction is pinned by a test. Extracting the
`case 'tests':` content-type setup into a testable private method (as `configureAuthorizationPipeline()` already
was, for the same reason) would close it. Disclosed rather than papered over.

## Summary by CodeRabbit

* **Bug Fixes**
  * PATCH requests now validate permissions against both the existing and newly requested test scope.
  * Invalid or outdated scope information in test updates is rejected safely.
  * Test payloads missing valid `applies_to` details now fail closed.
  * YAML and form-data request bodies for test updates are rejected; JSON is required.

* **Documentation**
  * Updated API documentation to clarify scope validation and re-scoping permission requirements.


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
## CodeRabbit triage: 1 applied, 1 declined

### Applied — 403 missing from PATCH (and DELETE)

Valid, and slightly understated. `PUT` already `$ref`s `Forbidden403`; `PATCH` did not — and PATCH is the verb this
PR makes *most* able to 403, since it gains a second independent authorization failure (holding `editor` on the
stored scope but not on the payload-derived target).

The same omission applied to **DELETE**, which CodeRabbit did not mention: it authorizes with the `admin` relation
(`['PUT' => 'editor', 'PATCH' => 'editor', 'DELETE' => 'admin']`), so it can 403 too. Same one-line class of fix,
so both are declared now.

Declared responses after the change:

```text
PUT     201 400 401 403 409 415 422 503
PATCH   200 400 401 403 404 415 422 503
DELETE  204 401 403 404
```

Applied as a surgical text edit — a JSON round-trip reformatted all 11k lines, since the file uses 2-space
indentation. Diff is 6 insertions, nothing else touched. `composer lint:openapi` still valid with only the 11
pre-existing scoped ignores.

### Declined — "preserve `application/yaml` for PATCH, as covered by `TestsHandlerTest`"

The cited test asserts the **opposite** of what the finding claims. `phpunit_tests/Handlers/TestsHandlerTest.php`
`testPatchWithYamlContentTypeIsRejectedAsUnsupportedMediaType` (`:234-254`) pins that a YAML body is now **rejected**
with 415. It was added by this PR precisely to make the restriction deliberate rather than incidental.

The finding's own alternative — "update the established API contract and corresponding tests if YAML is
intentionally no longer supported" — is what this PR did, and the contract needed no updating because it already
said JSON: `openapi.json` declared only `application/json` for these request bodies *before* this PR. The Router was
out of sync with its own published contract; this aligned them.

Why YAML had to go rather than be preserved: `JsonBodyParserMiddleware` populates `getParsedBody()` only for
`application/json`, and the new payload-target authorization reads `getParsedBody()`. A YAML PATCH would therefore
have failed closed with a **403** before the handler ever parsed it — a worse outcome than an honest 415. A
monorepo-wide search (frontend, UnitTestInterface, components libraries, docs, wiki) found zero clients sending
YAML or form bodies to `/tests`.

On scope: the restriction changes only `setAllowedRequestContentTypes`. `setAllowedAcceptHeaders([JSON, YAML])` is
untouched, so YAML **responses** still work. And of the methods the handler implements, only PUT and PATCH carry a
request body — so "limit it to the intended routes" is already the effect.

Commit: `f5f82901`

Issue #806, opened 2026-08-18
Proposal: a versioned, self-describing WebSocket contract to replace the empirically-grown Health handshake


JohnRDOrazio:
## Summary

The `Health` WebSocket protocol was built incrementally, a working piece at a time, and the vocabulary was never designed as a whole. It works — but the naming now actively misleads readers, the same word carries two meanings, and the source-data layout is duplicated across repos with nothing detecting drift.

This issue proposes a coherent v2 contract. It is a **design proposal, not a merged decision** — the point is to agree the shape before anyone writes code. The client-side counterpart is Liturgical-Calendar/UnitTestInterface#42.

Filed at the maintainer's request after a run of four incidents that share one root cause.

## Why now: the same bug has fired four times

| When | What broke | Root cause |
| --- | --- | --- |
| API #737 → UTI #38 | UTI's hardcoded source-data paths | client duplicates the server's on-disk layout; the fix commit says *"lockstep with API #737"* |
| API #795 | all 12 `.vscode` schema globs matched nothing | same layout, a third copy, never chased |
| API #800 | Ambrosian temporale unvalidatable by any client | the server's path→schema match table wasn't updated with the data |
| this week | a reviewer proposed a breaking `category` change on UTI#41 | the only written spec (`@phpstan-type`) disagrees with the code — see #805 |

Every one is the same thing: **the layout and the vocabulary live in several places that must be edited in lockstep, and nothing fails loudly when they diverge.**

## The specific ambiguities

**1. `category` is one property name carrying two disjoint enums — which overlap.**

- On `executeValidation` it selects a *schema-resolution strategy*: `universalcalendar`, `sourceDataCheck`, `resourceDataCheck` (+ four more the server still accepts, see #805).
- On `validateCalendar` / `executeUnitTest` it names a *calendar type*: `nationalcalendar`, `diocesancalendar`, `ritecalendar`.

`nationalcalendar` and `diocesancalendar` appear in **both**, meaning different things. And `universalcalendar` *looks* like a member of the calendar-type enum but belongs to the strategy enum.

**2. `sourceFile` is polymorphic with no discriminator.** Across the two clients it is, depending on `category` and sometimes on the `validate` prefix:

| shape | example |
| --- | --- |
| absolute API URL | `http://…/calendars` |
| absolute API URL with query | `http://…/data/nation/IT?locale=it` |
[...]

Issue #800, opened 2026-08-18
Health: the Ambrosian temporale is missing from getPathToSchemaFile(), so its source-data check cannot be validated


JohnRDOrazio:
## Summary

`Health::getPathToSchemaFile()` maps known source-data file paths to the schema they must validate against. It registers the Ambrosian **sanctorale**:

```php
JsonData::AMBROSIAN_SANCTORALE_FILE->value => LitSchema::PROPRIUMDESANCTIS->path(),
```

but **not** the Ambrosian **temporale**. `JsonData::AMBROSIAN_TEMPORALE_FILE` — `jsondata/sourcedata/rite/ambrosian/missals/propriumdetempore/propriumdetempore.json` — falls through to `default => null`.

The asymmetry looks like a simple oversight: the sanctorale arm was added deliberately (`HealthSchemaMappingTest` documents it as "Plan 5 / Task 6"), the temporale arm alongside it was not, and every Roman missal has both its temporale and its sanctorale registered:

```php
JsonData::MISSALS_FOLDER->value . '/propriumdetempore/propriumdetempore.json' => LitSchema::PROPRIUMDETEMPORE->path(),
JsonData::MISSALS_FOLDER->value . '/propriumdesanctis_1970/propriumdesanctis_1970.json' => LitSchema::PROPRIUMDESANCTIS->path(),
```

## Effect

A WebSocket `executeValidation` message naming the Ambrosian temporale gets a `null` schema, so `processValidationData()` takes the else branch and reports:

```text
executeValidation validation->sourceFile (JSON): Unable to detect schema for dataPath … and category …
```

The file-exists and json-valid assertions still pass; only schema validation reports an error. So the Ambrosian temporale is currently the one shipped missal file whose source data no client can schema-validate.

## Why now

Found while implementing UnitTestInterface#39 (rite-awareness — selecting and testing the Ambrosian rite-level calendar). Once UTI can select the Ambrosian rite calendar, its universal source-data checks point at the Ambrosian missal pair. The sanctorale check goes green; the temporale check cannot, purely because of this missing arm. It is a blocker for a clean Ambrosian run from the UI.

## Fix

Add the missing arm next to the sanctorale one:

```php
JsonData::AMBROSIAN_TEMPORALE_FILE->value => LitSchema::PROPRIUMDETEMPORE->path(),
```

[...]

PR #815, opened 2026-08-20
feat(validations): enumerate every checkable source artifact

JohnRDOrazio:
Refs #806. **Plan 1 of 2 for section B.**

`GET /validations` shipped in #811 advertising 18 hand-listed static items. This grows it to **77** by
*enumerating* per-calendar source data — national calendars, wider regions, diocesan calendars and test
definitions — so a client can derive its whole check list from one fetch instead of constructing
identifiers itself, which is the duplication #806 exists to end.

**No message shape changes in this PR.** `executeValidation`, `validateCalendar` and `executeUnitTest`
are untouched and the WebSocket protocol still speaks its legacy vocabulary. The typed `target` that
consumes these ids is the next PR. Nothing published in #811 was renamed.

## Where the items come from

Enumerated from `CalendarMetadataProvider::create()` — the same builder that serves `/calendars` — so
the two lists cannot disagree: a calendar that exists is a calendar that is checkable. Test definitions
come from globbing `*Test.json`, matching `TestsHandler`'s own glob.

| Id form                             | Source                                   | Schema                     |
|-------------------------------------|------------------------------------------|----------------------------|
| `nation:roman:IT`                   | `MetadataCalendars::$national_calendars` | `NationalCalendar.json`    |
| `widerregion:roman:Europe`          | `MetadataCalendars::$wider_regions`      | `WiderRegionCalendar.json` |
| `diocese:{rite}:milano_it`          | `MetadataCalendars::$diocesan_calendars` | `DiocesanCalendar.json`    |
| `test:{rite}:StIgnatiusOfLoyolaTest`| `jsondata/tests/{rite}/*Test.json`        | `LitCalTest.json`          |

77 items = 18 static + 5 nations ×2 + 16 dioceses ×2 + 3 wider regions ×2 + 11 tests. Each file-kind
item gets an `:i18n` folder sibling; tests have none.

## `region` semantics

`region` answers one question: *is this item specific to one nation's calendar?*

- `nation:roman:IT` → `'IT'`; `diocese:roman:romamo_it` → its nation, `'IT'`.
- `test:{rite}:X` → `null`; a test definition is not nation-scoped.
- `widerregion:roman:Europe` → `null`, **and this is a documented limitation, not an oversight.** A
  wider region spans several nations and a scalar cannot say so. Clients already have the mapping —
  `/calendars` gives each national calendar its `wider_region` — so a client scoping to one calendar
  reads that field. Widening `region` to a list would be a wire-contract change and belongs to plan 2
  if it is wanted at all.

## The Vatican is deliberately excluded

`CalendarMetadataProvider` hardcodes a `calendar_id = 'VA'` national entry, but the Vatican is served by
the General Roman Calendar and has **no source data of its own by design** — there is no
`.../nations/VA/` and there is not meant to be. Listing it would hand every client a target whose
`exists` step fails forever. It is skipped on `ResourceExistenceChecker::VATICAN_NATIONAL_CALENDAR_ID`,
the constant that already encoded this fact (widened to `public` so the two places that know it are
linked rather than duplicating a `'VA'` literal).

This is a categorical exclusion of one statically-known id, not a filesystem check — the endpoint still
never stats a target to decide whether to list it.

## Fixed counts replaced by bounds

`assertCount(18, …)` and an exact 9-and-9 both described a *hand-listed* inventory. Now that it is partly
enumerated they would fail whenever a calendar is added — a false alarm rather than drift. They are
replaced by a floor plus a **structural** invariant (folder ⟺ `:i18n` suffix ⟺ `LitSchema::I18N`) that is
strictly more informative than the number was, plus an id-uniqueness assertion the old count never
provided. Coverage is pinned by name in the drift tests, not by arithmetic.

## Drift coverage

The section A drift test walks the source tree and asserts every file found has an entry, which guards
the hand-listed half. The enumerated half needs the mirror guarantee, and now has it: every calendar the
index reports must have an inventory entry. A diocese added to source data, picked up by `/calendars` and
missed here would otherwise be unvalidatable with no failure anywhere — #800 exactly, arrived at from the
other direction.

Two properties worth calling out:

- The missing-national set is asserted to be **exactly** `[VA]`, not "VA is among the misses", so the one
  deliberate exception cannot become a hiding place for an accidental one.
- The test-definition guard drives both `CheckableInventory` and the live `TestsHandler` and set-compares
  the resulting **ids**, so it fails if either side's glob drifts from the other. `TestsHandler` enforces
  payload-name == filename on write, so the coupling it relies on is checked at both ends.

Every guard was verified by mutation — removing a producer, narrowing a glob — and confirmed to fail
naming the missing thing. A drift test that has never been seen to fail is not evidence.

## Two robustness fixes from the final review

**A broken inventory no longer takes out unrelated checks.** `CheckableInventory` calls
`CalendarMetadataProvider::create()`, which does not merely index folders — it reads and JSON-parses every
national and diocesan calendar file. One malformed file therefore threw all the way out, and in `Health`
— a long-running process — that broke schema resolution for *every* check, including the Roman temporale,
which has nothing to do with it. The inventory lookups in `Health` now contain that and fall through to
the legacy slug arms.

`/validations` itself still returns 503 in that case, deliberately: degrading to an empty index would
convert a loud failure into exactly the silent absence this endpoint exists to prevent. That inheritance
from `CalendarMetadataProvider` is now documented rather than accidental.

One consequence, stated plainly: containment restores the *process* and restores resolution for
`sourceDataCheck`, but a `universalcalendar` source-path check under a broken inventory degrades to
"Unable to detect schema" rather than validating. That is deliberate — the only richer fallback would be
to reintroduce the duplicated path table #811 deleted.

**A failed `glob()` now raises instead of silently emptying.** `glob()` returns `[]` for a
readable-but-empty directory and `false` only on a system error, so `?: []` was swallowing the sentinel:
an unreadable `jsondata/tests/{rite}/` would have dropped all 11 test definitions with no error anywhere.

## Behaviour change outside `/validations`

Growing `CheckableInventory` grows what `Health::getPathToSchemaFile()` resolves, so a per-calendar source
path passed on a `universalcalendar` message now resolves to a schema where it previously returned `null`.
This is strictly additive — the legacy `sourceDataCheck` arms key on hyphenated *slugs*, not paths, so
nothing is shadowed — but it is a change outside the inventory in a PR otherwise scoped to it.

## Verification

`development` (including #813, which rewrote `Health.php`'s schema resolution) is merged in and the result
verified, since CI tests the branch rather than the merge result:

```text
phpunit_tests/Health*.php + phpunit_tests/Models/ValidationsPath/   154 tests, 1218 assertions, OK
composer test:quick                                                 baseline byte-identical to base
composer lint                                                       clean
composer analyse (PHPStan level 10, 317 files)                      no errors
composer lint:md                                                    clean
```

The two `Health.php` changes were confirmed to compose semantically, not just textually: they touch
disjoint regions, the new `try` wraps only the inventory lookup with #813's widened arms sitting below the
`catch` in the fallback path, and inventory ids are `kind:rite:subject` so `byId()` can never intercept
#813's hyphenated slugs. #813's own test for `national-calendar-USA` proves this directly — that slug is
resolvable only by its widened arm and reachable only by passing through this branch's `byId()` first.
Both sides also derive diocesan paths through the same pre-existing `JsonData::diocesanCalendarFileFor()`
resolvers, so the inventory cannot advertise a path `Health` does not check.

## Docs

The section A design document is amended where this work revisits it — per-calendar items were previously
out of scope, and the "never stats the filesystem" principle is narrowed to what actually mattered: it
never stats *a target* to decide whether to list it. Both are amended in place rather than quietly
deleted.

A diocese id used throughout the planning docs, `roma_lazio_it`, turned out to be fabricated — it exists
nowhere in the source data. Corrected to the real `romamo_it`.

## Summary by CodeRabbit

* **New Features**
  * Expanded the `/validations` inventory to include national, regional, diocesan, and rite-specific test checks.
  * Added comprehensive discovery of all checkable calendar and test resources.
  * Preserved stable validation identifiers while keeping underlying file paths hidden.

* **Bug Fixes**
  * Improved schema and route resolution when inventory data is unavailable or malformed.
  * Excluded calendars without independent source data from the inventory.

* **Documentation**
  * Added implementation and design documentation covering inventory coverage, typed targets, compatibility, and validation behavior.

### DIFF
--- a/docs/superpowers/plans/2026-08-20-inventory-covers-everything.md
+++ b/docs/superpowers/plans/2026-08-20-inventory-covers-everything.md
@@ -1,0 +1,879 @@
+# Inventory Covers Everything Checkable — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or
+> superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Grow `/validations` from 18 static items to every checkable source artifact, so a later change can make the
+inventory id the only address a client ever sends.
+
+**Architecture:** `CheckableInventory` gains a second half: per-calendar source data enumerated from
+`CalendarMetadataProvider::create()`, the same builder `/calendars` uses, plus test definitions read from the
+rite-partitioned tests folders. The static half is untouched. No message shape changes in this plan.
+
+**Tech Stack:** PHP 8.4, PHPUnit 12.
+
+**This is plan 1 of 2 for the section B spec.** It grows the inventory only, and ships useful on its own: `/validations`
+becomes complete, and a client can derive its whole check list from one fetch while still speaking the legacy protocol.
+Plan 2 adds the typed `target`, the tagged calendar identity, and the `Health` rewiring that consumes these ids.
+
+## Global Constraints
+
+- Work in the worktree `/home/johnrdorazio/development/LiturgicalCalendar/LiturgicalCalendarAPI-target` on branch
+  `feat/806-typed-target` (PR base: `development`). **Never commit in the main checkout**
+  `/home/johnrdorazio/development/LiturgicalCalendar/LiturgicalCalendarAPI` — shared with other agents.
+- Never use `git commit --no-verify`. Commits are GPG-signed; if signing fails, stop and ask.
+- PSR-12 per `phpcs.xml`; short array syntax; 4-space indent; single quotes unless interpolating.
+- PHPStan level 10 over `src` only.
+- **Paths come from `JsonData` enum cases, never string literals.**
+- **`CheckableItem` must never serialize its `path`.**
+- **The endpoint never stats a target to decide whether to list it.** An item appears because a calendar is *registered*,
+  not because a file was found. Presence stays the `exists` step's job.
+- Id vocabulary is `kind:rite[:qualifier][:i18n]`, fully qualified. Ids are opaque to clients.
+- **No message-shape changes in this plan.** `executeValidation`, `validateCalendar`, `executeUnitTest` are untouched.
+- New schema files must declare `https://json-schema.org/draft-07/schema#` and carry no `$id` — `SchemaConventionsTest`
+  enforces both.
+- Spec: `docs/superpowers/specs/2026-08-20-typed-target-design.md`.
+
+---
+
+## File Structure
+
+| File                                                              | Responsibility                                                    |
+|-------------------------------------------------------------------|-------------------------------------------------------------------|
+| `src/Models/ValidationsPath/CheckableInventory.php`               | Gains per-calendar and test enumeration alongside the static half |
+| `phpunit_tests/Models/ValidationsPath/CheckableInventoryTest.php` | Extended: the new kinds, their regions, their ids                 |
+| `phpunit_tests/Models/ValidationsPath/InventoryDriftTest.php`     | Extended: every registered calendar has an inventory entry        |
+| `phpunit_tests/Handlers/ValidationsHandlerTest.php`               | Count assertion updated; no path may still leak                   |
+| `docs/superpowers/specs/2026-08-19-checkable-inventory-design.md` | Amended: the filesystem principle, and the inventory's new reach  |
+
+### The four new kinds
+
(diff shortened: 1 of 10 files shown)


JohnRDOrazio:
@coderabbitai review

Issue #812, opened 2026-08-19
Health is not rite-aware: schema lookup and diocesan source paths assume Roman


JohnRDOrazio:
`Health` was never made rite-aware when the rest of the API was. Four call sites resolve
schemas and source paths as though Roman were the only rite. Two are live today; two are
latent and become live the moment the third is fixed, which is why they need fixing together.

Found while unblocking [UnitTestInterface#48](https://github.com/Liturgical-Calendar/UnitTestInterface/issues/48)
(the missing `RiteSelect`), which commits the test interface to sending rite-qualified
resource URLs. It cannot proceed until `Health` can resolve them.

## 1. `resourceDataCheck` schema lookup ignores the rite segment — LIVE

`src/Health.php:2146,2150`, in `retrieveSchemaForCategory()`:

```php
'/\/events\/(?:nation\/[A-Z]{2}|diocese\/[a-z]{6}_[a-z]{2})(?:\?locale=[a-zA-Z0-9_]+)?$/'
'/\/data\/(?:(nation)\/[A-Z]{2}|(diocese)\/[a-z]{6}_[a-z]{2}|(widerregion)\/[A-Z][a-z]+)(?:\?locale=[a-zA-Z0-9_]+)?$/'
```

Neither admits a rite segment. `Router::extractRiteSegment()` accepts one on `calendar`,
`events` and `data`, so `/events/ambrosian/diocese/milano_it` is a valid canonical URL — but
here `ambrosian` sits where the pattern demands `diocese`, nothing matches, the call falls
through to `getPathToSchemaFile()`, and that returns `null`.

The result is the `Unable to detect schema for dataPath …` failure. Note the rite segment is
*optional* and defaults to Roman, so the un-prefixed form still works — this only breaks the
explicit form, which is the **canonical** one per `Router::canonicalRiteUrl()`.

**Fix:** admit an optional `(?:roman|ambrosian)/` segment after `/events/` and `/data/`.

## 2. Diocesan i18n folders resolve under the Roman tree — LIVE

`src/Health.php:640`, the `sourceDataCheck` i18n branch:

```php
$dataPath = strtr(JsonData::DIOCESAN_CALENDAR_I18N_FOLDER->path(), [...]);
```

`DIOCESAN_CALENDAR_I18N_FOLDER` is the Roman constant. Ambrosian diocesan data lives under
`jsondata/sourcedata/rite/ambrosian/calendars/dioceses/`, reached via
[...]

PR #816, opened 2026-08-20
fix(health): resolve schemas for bare rite-qualified collection paths

### LINKED ISSUE #814: Health: bare rite-qualified collection paths (/events/roman, /tests/ambrosian) resolve no schema
Follow-up to #812 / #813, which made `Health`'s schema lookup rite-aware for *item* routes but missed
the **bare rite-qualified collection** form.

`Health::getPathToSchemaFile()` ends in a `match` on **exact** string equality
(`src/Health.php:2092-2102`), with `default => null`:

```php
return match ($dataFile) {
    Route::EVENTS->path()  => LitSchema::EVENTS->path(),   // '/events'
    Route::TESTS->path()   => LitSchema::TESTS->path(),    // '/tests'
    Route::DATA->path()    => LitSchema::DATA->path(),     // '/data'
    ...
    default                => null
};
```

and the `resourceDataCheck` regex arm added by #813 (`src/Health.php:2176,2182`) requires a
`nation/` or `diocese/` (or `widerregion/`) segment *after* the optional rite:

```php
'/\/events\/(?:(?:roman|ambrosian)\/)?(?:nation\/[A-Z]{2}|diocese\/[a-z]{6}_[a-z]{2})…$/'
```

So the collection form with a rite falls between the two and resolves to `null`:

| Path                        | Exact `match` | Regex arm | Result   |
|-----------------------------|---------------|-----------|----------|
| `/events`                   | ✅ hit        | —         | resolved |
| `/events/roman/nation/IT`   | —             | ✅ hit    | resolved |
| `/events/roman`             | ❌ miss       | ❌ miss   | **null** |
| `/events/ambrosian`         | ❌ miss       | ❌ miss   | **null** |
| `/data/ambrosian`           | ❌ miss       | ❌ miss   | **null** |
| `/tests/ambrosian`          | ❌ miss       | ❌ miss   | **null** |

All three are valid routes. `Router::extractRiteSegment()` accepts a rite on `calendar`, `events` and
`data`; `/tests/{rite}` is handled separately at `Router.php:620-624` and "behaves exactly like
`/tests` (collection)". The `/events/{rite}` collection case is already exercised as the *comune* case
in `phpunit_tests/Handlers/SettingsRiteEchoTest.php:143`.

`/calendar/{rite}` is **not** affected — `Route::CALENDAR` is absent from that `match` and calendar
paths resolve elsewhere.

## Why it matters

This is the form a `RiteSelect` produces when no nation or diocese is selected — the "whole rite"
calendar. UnitTestInterface#48 will emit it as soon as it ships, and it is the canonical explicit form
per `Router::canonicalRiteUrl()`, so it is not an edge case.

## Fix

Resolve the rite-qualified collection form to the same schema as its bare equivalent, for `events`,
`data` and `tests`. Whatever shape the fix takes, the invariant to encode in a test is that **for every
route carrying a rite segment, the bare and rite-qualified collection forms resolve to the same
schema** — so a future route added to the `match` cannot reintroduce the asymmetry silently.

## Provenance

Missed when scoping #812: that issue was written outward from the resolution arms and never asked what
the bare rite-qualified form does, even though the canonical-form argument it makes applies equally.
Caught on review of the merged result.

JohnRDOrazio:
Closes #814. Follow-up to #813, which made `Health`'s schema lookup rite-aware for *item* routes and
missed the bare rite-qualified **collection** form.

## The gap

`Health::getPathToSchemaFile()` ends in a `match` on **exact** string equality with `default => null`,
and #813's `resourceDataCheck` regex arm requires a `nation/`, `diocese/` or `widerregion/` segment
*after* the optional rite. The collection form with a rite fell between the two:

| Path                       | Exact `match` | Regex arm | Before   | After    |
|----------------------------|---------------|-----------|----------|----------|
| `/events`                  | ✅ hit        | —         | resolved | resolved |
| `/events/roman/nation/IT`  | —             | ✅ hit    | resolved | resolved |
| `/events/roman`            | ❌ miss       | ❌ miss   | **null** | resolved |
| `/events/ambrosian`        | ❌ miss       | ❌ miss   | **null** | resolved |
| `/data/ambrosian`          | ❌ miss       | ❌ miss   | **null** | resolved |
| `/tests/ambrosian`         | ❌ miss       | ❌ miss   | **null** | resolved |

All three are valid routes — `Router::extractRiteSegment()` accepts a rite on `calendar`, `events` and
`data`, and `/tests/{rite}` is handled at `Router.php:620-624` and "behaves exactly like `/tests`". The
`/events/{rite}` collection case is already exercised as the *comune* case in `SettingsRiteEchoTest`.

This is the form a `RiteSelect` produces when no nation or diocese is selected — the whole-rite calendar
— so UnitTestInterface#48 hits it immediately. It is also the canonical explicit form per
`Router::canonicalRiteUrl()`.

## The fix

A trailing rite segment is stripped before the `match`, so the rite-qualified collection form resolves to
the same schema as its bare equivalent.

**The rite alternation is derived from the `Rite` enum**, and both of #813's `resourceDataCheck` regex
arms now interpolate the same helper instead of a hardcoded `roman|ambrosian`. That literal is what would
have let a third rite silently reopen this hole and #812's; fixing the symptom with another hardcoded
alternation would only have reset the trap.

### One deliberate consequence

The strip is **uniform**, not gated on the three rite-carrying routes. So `/missals/roman`,
`/decrees/ambrosian` and `/calendars/roman` now also resolve to their bare form's schema, even though the
`Router` 404s them.

This is intentional, and is recorded in the code and pinned by a test rather than left to be rediscovered:

- `getPathToSchemaFile()` answers *"which schema would validate a response from this path"*, not *"is this
  path routable"*. Those are different questions and the `Router` remains the sole authority on the second.
  A bogus path still fails loudly — the fetch 404s — arguably with a clearer message than "Unable to
  detect schema".
- The alternative, gating on a list of rite-carrying routes, would reintroduce exactly the staleness that
  produced this bug. That knowledge currently lives in an inline condition inside
  `Router::extractRiteSegment()` with `/tests` handled separately, so there is no registry to derive from.
  Trading a structural invariant for a hand-maintained list is the wrong direction.

`testARiteOnANonRiteCarryingRouteStillResolves()` pins the behaviour so a later reader cannot mistake it
for an oversight and "fix" it back into a stale list. The lookup consults nothing about routability.

## `/calendar/{rite}` is unaffected, and here is why

Not because of anything to do with rite handling: `Route::CALENDAR` simply has no arm in that `match` and
never had one. Calendar responses are never validated through `getPathToSchemaFile()` — they go down the
`validateCalendar` action, which fetches `Route::CALENDAR->path() . $req` and validates against
`LitSchema::LITCAL->path()` directly. So `/calendar/roman` strips to `/calendar` and falls to
`default => null` exactly as before.

The `Route::CALENDAR` × `Rite` pairs are in the invariant test's cross product and passed *before* the fix
as well as after — both sides null — which is what makes them a guard rather than a coincidence.

## The test encodes the invariant, not the cases

For every route carrying a rite segment, the bare and rite-qualified collection forms must resolve to the
same schema — driven from the enums as a cross product rather than a hand-written list, so a route added
to the `match` later cannot reintroduce the asymmetry silently. A test enumerating exactly `/events/roman`,
`/data/ambrosian` and `/tests/roman` would pass today and guard nothing tomorrow.

## Verification

```text
vendor/bin/phpunit phpunit_tests/Health*.php   222 tests, 305 assertions, OK (3 pre-existing skips)
composer lint                                  clean
composer analyse (PHPStan level 10, 317 files) no errors
```

Confirmed discriminating: against pre-fix source the new tests fail **38 data sets** (16 + 16 + 6).

## How this was missed

#813 was scoped outward from the resolution arms, and its issue text framed the problem as "the regexes
lack a rite segment". Five reviews and CodeRabbit all checked that framing and none asked the prior
question — which rite-qualified paths exist at all. A precise issue description narrows reviewers'
attention as effectively as it directs it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rite-qualified collection URLs now resolve to the same schemas as their equivalent bare URLs.
  * Improved schema validation for events, data, and tests collections when a rite suffix is included.
  * Added support for all recognized rites, ensuring new rite values are handled consistently.
  * Rite suffixes on routes that do not require them no longer prevent schema resolution.
  * Unknown rite values continue to be rejected appropriately.

### DIFF
--- a/phpunit_tests/HealthSchemaCategoryTest.php
+++ b/phpunit_tests/HealthSchemaCategoryTest.php
@@ -6,6 +6,7 @@ namespace LiturgicalCalendar\Tests;
 
 use LiturgicalCalendar\Api\Enum\JsonData;
 use LiturgicalCalendar\Api\Enum\LitSchema;
+use LiturgicalCalendar\Api\Enum\Rite;
 use LiturgicalCalendar\Api\Enum\Route;
 use LiturgicalCalendar\Api\Health;
 use LiturgicalCalendar\Api\Router;
@@ -54,6 +55,15 @@ final class HealthSchemaCategoryTest extends TestCase
         $method = new \ReflectionMethod(Health::class, 'retrieveSchemaForCategory');
         /** @var string|null $result */
         $result = $method->invoke(null, $category, $dataPath);
+
+        return $result;
+    }
+
+    private static function getPathToSchemaFile(string $dataFile): ?string
+    {
+        $method = new \ReflectionMethod(Health::class, 'getPathToSchemaFile');
+        /** @var string|null $result */
+        $result = $method->invoke(null, $dataFile);
 
         return $result;
     }
@@ -225,9 +235,12 @@ final class HealthSchemaCategoryTest extends TestCase
         self::initRouter();
 
         return [
-            'events, invented rite' => [Route::EVENTS->path() . '/byzantine/diocese/milano_it'],
-            'data, invented rite'   => [Route::DATA->path() . '/byzantine/diocese/milano_it'],
-            'data, doubled rite'    => [Route::DATA->path() . '/roman/ambrosian/diocese/milano_it'],
+            'events, invented rite'            => [Route::EVENTS->path() . '/byzantine/diocese/milano_it'],
+            'data, invented rite'              => [Route::DATA->path() . '/byzantine/diocese/milano_it'],
+            'data, doubled rite'               => [Route::DATA->path() . '/roman/ambrosian/diocese/milano_it'],
+            'events, invented rite collection' => [Route::EVENTS->path() . '/byzantine'],
+            'data, invented rite collection'   => [Route::DATA->path() . '/byzantine'],
+            'data, doubled rite collection'    => [Route::DATA->path() . '/roman/ambrosian'],
         ];
     }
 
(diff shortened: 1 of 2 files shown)


JohnRDOrazio:
@coderabbitai review

Issue #823, opened 2026-08-20
A throw inside a fulfil handler skips the terminal frame and wedges a v2 client


JohnRDOrazio:
The terminal `{"step":"complete"}` frame added for #806 section C lets a client stop on an explicit signal
instead of counting frames. Its guarantee is that **every path which starts work terminates**.

There is one structural exception, and it should be closed or the guarantee should be qualified.

## The gap

React does not invoke a `then()`'s sibling `onRejected` when its own `onFulfilled` throws. So an exception
raised *inside* a fulfilment handler skips the `sendComplete()` call that follows it in the same handler.

Four sites are shaped this way in `src/Health.php` — the URL and filesystem fulfil handlers in
`runValidationSteps()`, the calendar fulfil handler in `validateCalendar()`, and the test-run fulfil handler
in `executeUnitTest()`.

A v2 client that stops on `complete` therefore waits forever for that request. That is precisely the wedge
the frame exists to prevent, arrived at from inside rather than outside.

## Why it is not urgent

The exposure is narrow. `validateDataAgainstSchema()` catches `\Throwable` internally and returns an error
object rather than throwing, and it is the principal throw source in those handlers. The shape is also
pre-existing: before the terminal frame, a client counting `checks * 3` wedged on the same paths for the same
reason. So this is an **unclosed corner, not a regression** — the new frame simply makes the corner nameable.

It was deliberately not fixed in the section C branch. Wrapping error paths in `catch` blocks late in the
last code task, without its own review, is how a late change to error handling becomes the thing that breaks.

## Fix

Wrap the four fulfil-handler bodies so a throw still terminates the request — emit `complete` (already gated
on `requestId`, so a v1 client is unaffected) and then rethrow or log, whichever matches how the surrounding
handler is expected to fail.

Worth a test per site driving a throw from inside the fulfil handler and asserting the terminal frame still
arrives. The existing terminal-frame tests all drive *rejections*, which take the sibling handler and are
therefore the path that already works.

## Related

- #806 section C is what introduced the guarantee this qualifies.
- [#822](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/822) is a different hole in the
  same area — a missing file reporting `file-exists = success` — and does not overlap with this one.

JohnRDOrazio:
**Correction: there are five sites, not four.** The original report undercounted, and since this list is the
fix-list, an undercount means a silent under-fix.

[...]

Issue #822, opened 2026-08-20
A missing source file reports file-exists = success


JohnRDOrazio:
A source-file check whose file does not exist reports **`file-exists` = success**, then fails the two steps after
it. The one step named after detecting absence is the one that cannot.

## Reproduction

```php
$fs = React\Filesystem\Factory::create();
$fs->file('/path/that/does/not/exist.json')->getContents()->then(
    fn ($data) => print('RESOLVED with ' . strlen((string) $data) . " bytes\n"),
    fn (\Throwable $e) => print("REJECTED: {$e->getMessage()}\n")
);
```

Output:

```text
PHP Warning:  stat(): stat failed for /path/that/does/not/exist.json in vendor/react/filesystem/src/Fallback/File.php on line 32
PHP Warning:  Trying to access array offset on false in .../Fallback/File.php on line 32
PHP Warning:  file_get_contents(...): Failed to open stream: No such file or directory in .../Fallback/File.php on line 32
RESOLVED with 0 bytes
```

`react/filesystem`'s **Fallback** adapter resolves rather than rejects. That adapter is what this project uses:
there is no `ext-eio` or `ext-uv` in `composer.json` or the `Dockerfile`, so `Factory::create()` always returns it.

## Consequence

`Health::cachedFileGetContents()`'s rejection handler — and therefore `handleValidationDataError()`, which emits
`file-exists = error` — is **never reached** for a missing file. Instead the success path runs with empty data:

| step          | reported | truth                        |
|---------------|----------|------------------------------|
| `file-exists` | success  | the file is absent           |
| `json-valid`  | error    | correct, but for the wrong reason (empty string) |
| `schema-valid`| error    | correct, but for the wrong reason |

A user reading the UnitTestInterface sees a green "exists" box above two red ones, which reads as "the file is there
but its contents are broken" — the opposite of what happened. That is worse than a plain failure, because it sends
whoever is debugging to the wrong file.

The three PHP warnings also go to stderr on every such check, which is the `react/filesystem` noise that
`#[WithoutErrorHandler]` was papering over in a couple of test classes.

## Scope

Pre-existing, and identical for v1 `executeValidation` and v2 `validateSource` — the read path is shared, so this is
not a regression from [#820](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/pull/820) or from the
section C work in progress.

It is closely related to but distinct from
[...]

Issue #821, opened 2026-08-20
A file check that fails JSON decoding emits two frames, not count(steps)


JohnRDOrazio:
`GET /validations` publishes a `steps` array per item, and a client is meant to size a check's progress as
`count($item['steps'])` — that is exactly what replaces
[UnitTestInterface#42](https://github.com/Liturgical-Calendar/UnitTestInterface/issues/42)'s four hardcoded
`* 3` "responses per check" constants.

That contract holds everywhere except one path. `Health::processValidationData()` emits **two** frames when a
file's JSON fails to decode — `file-exists` and `json-valid` — and never sends `schema-valid`. A client sizing
the phase as three waits forever for a frame that is not coming, and the run totals drift from the rendered
cards.

The sibling paths are correct:

- `handleValidationDataError()` emits all three.
- The **folder** branch was explicitly fixed for exactly this, in `ea29b678` ("emit one frame per step on a
  sourceFolder check") and `38fdd054` ("keep the three-frame contract on the short paths too").

The file branch was not covered by that work.

## Why it matters more now

Pre-existing, and equally true of the v1 `executeValidation` shape — so this is **not** a regression from
[#820](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/pull/820). But #820 routes roughly 60
per-calendar `validateSource` file checks through this path, addressed by an inventory id whose published
sizing contract is `steps`. The contract and the emission disagree on more targets than before.

## Deliberately not fixed in #820

Adding the third frame changes v1 output, and #820's central guarantee is that the legacy shapes behave
byte-for-byte as before. Fixing it there would have broken the one property the whole branch was built to
preserve. The design spec in that PR instead documents the exception honestly, under
`### The one place count(steps) is wrong today`.

## Distinct from #819

[#819](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/819) is about step **names** —
`/validations` publishes `exists`/`parses`/`validates` while the frames are classed
`file-exists`/`json-valid`/`schema-valid`. This issue is about the **count**. A client can work around #819 by
ignoring the values; it cannot work around this one, because the missing frame is simply absent.

Both should be closed before UnitTestInterface#42 relies on `steps`.

## Fix

Emit `schema-valid` on the JSON-decode failure path with a failing result, the way `handleValidationDataError()`
already does, so every file check emits one frame per published step regardless of where it stops.

[...]

PR #824, opened 2026-08-20
feat(health): structured, correlated responses with a terminal frame

### LINKED ISSUE #819: /validations publishes step names that no emitted frame uses
`GET /validations` publishes a `steps` array per item. The WebSocket runner emits a frame per step. The
two use **different words**, and nothing in the API relates them.

| Source                                   | Vocabulary                                     |
|------------------------------------------|------------------------------------------------|
| `CheckableInventory::STEPS` (`:43`)      | `['exists', 'parses', 'validates']`            |
| `Health`'s emitted frame classes         | `file-exists`, `json-valid`, `schema-valid`    |

`CheckableItem::jsonSerialize()` puts the first list on the wire. `Health` emits the second as
`.<label>.file-exists` and friends. A client that takes `steps` literally — which is what publishing it
invites — waits for a `.<label>.exists` frame that never arrives.

Present since #811 shipped `/validations`; unchanged by #815. Not currently biting because no client
consumes `steps` yet: [UnitTestInterface#42](https://github.com/Liturgical-Calendar/UnitTestInterface/issues/42)
is the first, and its four hardcoded `* 3` "responses per check" constants are exactly what `steps` was
published to replace. So this lands the moment that work starts.

Found while implementing #806 section B, when an attempt to assert emitted frames against the published
`steps` turned out to be impossible: there is no derivation, only a coincidence of length.

## Which way to converge — a recommendation, not a decision

**Keep the published `exists` / `parses` / `validates` and fix the correspondence.** Those names are
transport-agnostic and describe what the step *means*. The frame classes are CSS fragments inherited from
the DOM-coupled response design that [#806](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/806)
section C exists to replace — renaming the published vocabulary down to match them would be a regression
that section C then has to undo.

Two ways to close it, and they are not exclusive:

1. **Now, cheaply:** document the mapping as part of the `/validations` contract so a client can build it
   without guessing.
2. **Properly, in section C:** have the structured response carry the step name itself, at which point the
   CSS fragment becomes a client-side presentation detail and the divergence disappears.

Section B already carries a documented `FRAME_CLASS_FOR_STEP` map internally, with an
`assertArrayHasKey` guard so that a newly published step fails the suite loudly rather than silently
emitting nothing. That is a stopgap inside the server, not a published contract.

## Not just cosmetic

`steps` is how a client sizes a phase — it is the whole reason UnitTestInterface#42 can drop its `* 3`
constants. A published array whose values match no emitted frame is worse than not publishing one,
because it reads as a contract.

JohnRDOrazio:
Refs #806. **Sections C and E**, plus D's terminal frame. Closes #819. Follows #815 and #820, which made an
inventory id the address a client *sends*; this makes the response say what it is *about*.

## The problem

A response carried three usable fields: `type`, `text`, and `classes` — a CSS selector the server built and
the browser matched with `querySelectorAll`. So attribution was string matching, and the server knew Bootstrap
existed. Nothing said which request a frame answered, and nothing said a target was finished, which is why both
clients hardcode "3 responses per check" in four places.

## What ships

Every response keeps `type`, `text` and `classes` and **gains** structured fields:

```jsonc
{ "type": "success",                          // legacy projection
  "text": "...",
  "classes": ".temporale-roman.json-valid",   // legacy projection
  "runToken": "run-a",                        // legacy, at its historical position
  "runId": "run-a",                           // published name for the same value
  "requestId": "req-17",                      // echoed, only when the client sent one
  "target": { "id": "temporale:roman" },
  "step": "parses",                           // the vocabulary /validations publishes
  "status": "pass",
  "details": [ "..." ] }
```

Plus a terminal `{"step":"complete"}` per target, so a client stops on a signal instead of counting.

**`type` and `classes` are now derived from the structured fields**, in one place. That is the substance of the
change: 27 sites across three clusters each built their own selector, which is how #820's label-as-selector bug
happened. `FrameFamily::frameClasses()` is now the only `.`-joined selector composer in the repository, shared by
`Health` and `LitTestRunner`.

`step` carrying the published vocabulary is what **closes #819** — the two vocabularies stop diverging because
one is now a projection of the other rather than an unrelated third word.

## The terminal frame is gated on `requestId`

A client that stops on `complete` **must send a `requestId`**; a request without one receives no terminal frame.

This is not fussiness. A new frame changes the *stream*, and `resources.js` computes `expectedResponses = checks
* 3` and compares with `>=` — so a v1 client would hit its threshold on the three real frames, advance, and then
have the late `complete` frames increment whichever counter was then active, finishing the *following* phase
early too. A cascading miscount with nothing failing visibly, which is worse than a crash.

`requestId` is already the v2 opt-in signal, and a client adopting `complete` is adopting correlation anyway.

**#821 becomes moot rather than fixed.** A JSON-decode failure still emits two step frames rather than three;
nothing waits for the difference any more.

## `requestId` is threaded, never stored per connection

`runToken` lives in `$this->runTokens[$resourceId]` and is injected in `sendMessage()`. Mirroring that for
`requestId` is the obvious implementation and would be self-defeating: `Health` is async, frames are emitted from
promise closures, and several requests can be in flight on one connection — so a per-connection value would stamp
late frames with whichever request arrived most recently. That is the exact misattribution correlation exists to
prevent, and it would pass every sequential test.

It is captured in the closure and passed as a value. Proven by three interleaved reverse-order tests across three
emitters; a per-connection implementation fails all of them while the 23 sequential ones still pass.

## Additive — with two qualifications stated plainly

Every legacy frame is byte-identical: same `type`, same `text`, same `classes`, same key order, `responsetype` on
exactly the five sites that carried it. Verified by hand-diffing all four clusters against the merge-base, and
during development by throwaway drivers comparing every frame against `git show <base>:src/Health.php` — 224
frames on the calendar cluster, 60 on the test-run cluster, 262 on a re-verification after shared code moved, all
with zero differences.

Two things are **not** pure appends, and the claim should not be read as unqualified:

- `runId` is appended to every frame including v1 ones. Byte-identity holds for the legacy keys *and their order*,
  not for the frame as a whole.
- `CORRELATION_ID_PATTERN` gained a `/D` modifier, which tightens `runToken` too: a token ending in `\n` is no
  longer accepted. Without it, `$` matches before a final newline, so a 64-character id plus `\n` passed a
  `{1,64}` bound. Narrow and deliberate, but it is the branch's one genuine v1 behaviour change.

## Verification

```text
Health*.php + LitTestRunnerTest + Models/ValidationsPath   420 tests, 3391 assertions, OK
composer lint / analyse (PHPStan 10) / lint:md / lint:openapi   clean
```

Guards were verified by mutation rather than assumed. Deleting the terminal frame's gate fails 19 tests across six
pre-existing suites. Storing `requestId` per connection fails exactly the four interleaved tests. Dropping the
class qualifier fails only the test that names it. An unmapped `Step` throws rather than emitting `.<fragment>.`,
proven with a temporary enum case.

## Known limits, documented rather than discovered later

- **#823** — a throw *inside* a fulfil handler skips `sendComplete()` at five sites, because React does not invoke
  the sibling `onRejected`. Pre-existing in shape; the spec states it as the one exception to "every path that
  starts work terminates", by shape rather than by a list that can go stale.
- **#822** — a missing source file reports `file-exists = success`, because `react/filesystem`'s Fallback adapter
  resolves with 0 bytes. Pre-existing; fixing it would change v1 output.
- `.test-valid` is deliberately unchanged. It *addresses* the validity box, which the client colours by `status`;
  a class encoding the outcome would be worse, because the client would need a different selector depending on a
  result it has not parsed yet. The rename is reserved for legacy removal.
- `target.calendar` reads `'VA'` from Health's failure arms and `'roman'` from the runner for a rite-level
  calendar; each is true from its own vantage, and `requestId` is the real correlation key.
- `details` entries are HTML-escaped on the XML path and not on the ICS path, each faithful to its own `text`.
  Reconcilable only when `text` stops being a contract.

## Reviews

Fifteen review passes. They found the label-vs-id class of defect had a sibling in the plan itself, a fifth
unfalsifiable test — inherited by copying one that already could not fail — and a documentation gap where the
terminal frame's gate was stated nowhere a client reads while five published sentences said the opposite.

## Summary by CodeRabbit

- **New Features**
  - Health WebSocket responses now include structured targets, validation steps, pass/fail status, details, and run identifiers.
  - Requests can include a correlation ID, echoed in related responses for reliable tracking.
  - Correlated validations now conclude with an explicit `complete` frame.
  - Test-run responses identify their target and outcome more consistently.
  - Invalid or malformed requests receive clearer correlated or uncorrelated error responses.

- **Documentation**
  - Updated API and schema documentation to describe structured response fields, completion behavior, and validation-step vocabulary.


JohnRDOrazio:
Two CodeRabbit threads are left unresolved deliberately. Recording why, so they read as decisions rather than oversights.

**"Fix the silent request paths before publishing `complete` as the stop signal"** — declined for this PR, tracked as [#823](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/823).

The finding is correct: a throw *inside* a fulfil handler skips `sendComplete()`, because React does not invoke the sibling `onRejected`. But the fix is `catch` blocks around five error paths, and adding those late in a branch whose central guarantee is byte-identical legacy output — without their own review — is how a late change to error handling becomes the breakage. The shape is also pre-existing: before the terminal frame, a client counting `checks * 3` wedged on exactly the same paths. This branch does not create the corner, it makes it nameable.

The spec states it as one of two known holes in "every path that starts work terminates", by *shape* rather than by a list that can go stale — so anyone fixing #823 finds the rule rather than a stale enumeration. The second hole, `dropSupersededQueuedRequests()` discarding entries without terminating them, was found while checking this one.

**`#[Group('slow')]` on four test classes** — declined, measured rather than assumed: 20 ms, 34 ms, 36 ms, 77 ms.

This project reserves `slow` for genuinely measurable cost — multi-year calendar generation, rate-limit window waits. Applying it here would exclude these suites from `composer test:quick`, dropping the interleaved-correlation tests and every terminal-frame arm out of the suite people run before committing. Those are the two guarantees this PR exists to establish.

The attribute-not-docblock half of that advice is right and is recorded for whenever something here does warrant the group: a `@group slow` docblock silently selects zero tests under `--group slow` and exits successfully.

Everything else CodeRabbit raised is fixed — including a real error of mine it caught, the frame-assignment count, which was 27 in the spec and is 32 in the code. The corrected text states the counting rule and the command that reproduces it, since a bare number had by then been wrong three times.

Issue #827, opened 2026-08-20
Health WebSocket frames leak the server's absolute filesystem path to unauthenticated clients


JohnRDOrazio:
`Health` response frames interpolate `LitSchema::*->path()` — an **absolute** filesystem path — into text sent to any client on the WebSocket endpoint. No authentication stands in front of it.

Measured on `feat/806-strict-validation`:

```text
The $category of $calendar for the year $year was successfully validated against the Schema
<absolute path to the deployment>/jsondata/schemas/LitCal.json
```

*(The real path is elided here deliberately — see the note at the end.)*

Six sites interpolate a schema path into frame text (`grep -c 'against the Schema' src/Health.php`), and `LitSchema::*->path()` appears 26 times in the file. `src/Health.php:1914` leaks **two** absolute paths in a single error frame:

```php
'… (path for schema: $pathForSchema, Route::CALENDARS->path(): ' . Route::CALENDARS->path()
    . ', LitSchema::METADATA->path(): ' . LitSchema::METADATA->path() . ')'
```

## Why it is filed separately

Found while reviewing #806 section G. That work fixed the same class of leak on the **rejection** path — a `swaggest/json-schema` error embedded the schema's absolute path in every `$ref[…]` segment — by scoping validation to the claimed arm and adding a `sanitize()` backstop, pinned by a test asserting no rejection text contains `Router::$apiFilePath`.

These sites are on the **success and result** paths, are unchanged since before that branch, and are outside its remit. Fixing them there would have mixed an unrelated change into a security fix.

## Severity

Low on its own — it discloses the deployment's directory layout and the OS account name, not credentials or data. It is worth fixing because it is free to fix, because the layout is a useful input to anything else an attacker tries, and because the codebase now has a test asserting the rejection path does not do this, so the inconsistency will read as an oversight to the next person.

## Suggested fix

Send the schema's **name**, not its path — `LitCal.json` rather than the absolute path. `LitSchema` already has the name as its case value (`'/LitCal.json'`), so this is a `basename()` or a small accessor, not a redesign. `$schema` variables in `runValidationSteps()`/`processValidationData()` carry a resolved path and need the same treatment at the point of interpolation.

Then widen the existing guard: the section G test asserts no *rejection* text contains `Router::$apiFilePath`; make it assert that of **every** frame `Health` emits, so success frames are covered by the same rule. That test is the reason this will not come back.

## Related

[...]

Issue #826, opened 2026-08-20
Evaluate replacing swaggest/json-schema with a modern JSON Schema validator


JohnRDOrazio:
`swaggest/json-schema` v0.12.43 implements **draft-07**. Every schema in `jsondata/schemas/` declares draft-07, so the pairing is currently coherent — but the library is on a `0.x` line, and the repo publishes its schemas as a public contract, so it is worth deciding deliberately rather than by inertia.

Raised while designing #806 section G, where the draft-07 ceiling first had a practical consequence.

## What prompted it

Section G wants "reject unknown properties, but only for messages that opted into v2". The natural spelling is:

```jsonc
{ "allOf": [
    { "$ref": "#/definitions/executeValidation" },
    { "if": { "required": ["requestId"] }, "then": { "unevaluatedProperties": false } }
] }
```

`unevaluatedProperties` is draft 2019-09. In draft-07, `additionalProperties: false` sees only the properties declared in the **same** schema object, so inside a `then` it rejects everything. The only schema-only alternative is two arms per action with every property list written twice — the duplication #806 exists to remove.

Section G therefore applies that one gate in PHP while reading the allowed names from the schema, so the vocabulary still lives in one place. That is a fine outcome; it is not a reason on its own to migrate.

## Candidates

- **`opis/json-schema`** — draft-2020-12 / 2019-09 / 07 / 06 with all keywords, plus `$filters`, `$data`, `$map`, casting, custom formats. Actively maintained (2.6.0, October 2025). PHP 7.4+. Apache-2.0. The usual default for a validation pipeline.
- **`ropi/json-schema-evaluator`** — draft 2020-12 only, but passes the entire official JSON Schema test suite except `optional/refOfUnknownKeyword.json` and `optional/ecmascript-regex.json` (PCRE rather than ECMAScript regexes). PHP ≥ 8.1 plus bcmath, mbstring, fileinfo. Smaller community; the strongest conformance claim for `$dynamicRef` / `unevaluated*`.

## Blast radius, measured

**17 files** reference `Swaggest`: 5 in `src/` (`Health`, `RegionalDataHandler`, `TestsHandler`, `DecreesHandler`, `Test/LitTestRunner`) and 12 test classes. This is the API's entire source-data integrity pipeline, not a corner of it.

[...]

PR #828, opened 2026-08-21
feat(health): strict inbound validation and typed protocol errors (#806 section G)

### LINKED ISSUE #806: Proposal: a versioned, self-describing WebSocket contract to replace the empirically-grown Health handshake

JohnRDOrazio:
Before this branch, several malformed WebSocket messages **terminated the `Health` process for every connected client** — on an endpoint with no authentication in front of it.

`validateMessageProperties()` checked that required keys *existed* and never what they were, and the dispatch unpacked them straight into typed parameters:

```php
$this->validateCalendar($m->calendar, $m->year, $m->category, $m->responsetype, ...);
//                       string        int       string        string
```

Ratchet's `IoServer::handleData` catches `\Exception`. `TypeError`, `ValueError` and `Error` are not `\Exception`, so they escaped and killed the daemon. Measured against `c38dbb97`:

| message                                                       | before                                     |
|---------------------------------------------------------------|--------------------------------------------|
| `validateCalendar` with `"year": "not-a-year"`                | `TypeError` — process dies                 |
| `validateCalendar` with `"category": []`                      | `TypeError` — process dies                 |
| `validateCalendar` with `"responsetype": "NOT_A_FORMAT"`      | `ValueError` — process dies                |
| `executeUnitTest` with `"test": {…}`                          | `TypeError` — process dies                 |
| `executeValidation` with `"category": {…}`                    | `Error` (the `(string)` cast) — dies       |
| `validateCalendar` with `"year": 1e30`                        | `TypeError` — process dies                 |
| `executeValidation` with a NUL byte in `sourceFile`           | `ValueError` — process dies                |

The split was not accidental: the three v1 actions crashed and the three v2 actions did not, because the v2 handlers were written after the hazard was understood. This branch applies that lesson everywhere instead of door by door.

## What it does

- **`jsondata/schemas/WebSocketMessage.json`** is now the published contract, transcribed from the `@phpstan-type` annotations PHPStan already enforces against the implementation — so annotation, code and contract agree by construction. `ACTION_PROPERTIES` and `validateMessageProperties()` are deleted rather than kept beside it.
- **Every inbound message is validated** against the single definition it claims to be, resolved by the discriminator the dispatch already used.
- **Refusals are typed:** `{"type":"protocolError","errorCode":"…","text":"…"}`. Nine codes, each existing only where a client would *act* differently — `INVALID_MESSAGE` deliberately covers every schema violation, because "wrong type", "unknown enum value" and "undeclared property" all mean *fix the message*.
- **Unknown-property rejection is gated on `requestId`**, the same v2 opt-in the terminal `complete` frame uses. The shipped client sends `runToken` on every message and spreads `rite` onto an `executeValidation` that never reads it — a uniform rule would have taken the test interface down on day one.
- **A `\Throwable` backstop** around the dispatch, because a process-wide crash must not depend on a schema file being correct.

## Compatibility

**No message the shipped UnitTestInterface sends stops working**, verified against that client's source at `9126ad9` rather than against this branch's own fixtures. Run tokens are `crypto.randomUUID()` and match the correlation pattern; every `category` and `responsetype` it sends is inside the enums; all ~20 `executeValidation` sites pair `category` with `sourceFile`/`sourceFolder` the way the schema requires. The client sends no `requestId` anywhere, so the strict gate is entirely dormant for it.

Two deliberate tightenings, both refusing things that already failed: `year` must be a JSON number (the client sends numbers), and `category` must be a known value. The one `executeValidation` combination now refused — `universalcalendar` + `sourceFolder` — is the one `wsProtocol.js`'s own docblock records as closing the connection today.

## Evidence

The final review fuzzed **4000 mutated messages** and 75 targeted vectors through `onMessage()`: **0 escapes**. Then it neutered the schema gate and re-ran 800 of them — **92 hit the `\Throwable` backstop**, proving both layers are load-bearing rather than one guard wearing two hats.

Every guard was mutation-checked: remove it, watch the named tests fail, restore.

Three defects were found by *probing the running code* rather than reading diffs, and all are fixed here:

- validation against the whole `oneOf` reported an **irrelevant arm** — a `runTest` was told it was missing `sourceFolder`;
- rejection text **leaked the server's absolute filesystem path** to unauthenticated clients;
- `humanize()` silently dropped the entire reason on messages over ~13 KB (PCRE JIT stack limit).

Rejection text is now `<action>.<property.path>: <expectation>`, built from the action the client sent rather than the internal definition name, with no path, no stack trace and no internal vocabulary. A test asserts no rejection text ever contains the project root.

## Not in scope

- **#823** — throws inside promise callbacks, after `onMessage()` has returned, which no `try` around the dispatch can see. Documented in the spec.
- **#827** — `Health`'s *success* frames still quote absolute schema paths. Same class of leak, byte-identical to before this branch, filed separately rather than mixed into a security fix.
- **#826** — a draft-2019-09+ validator would let the unknown-property gate move into the schema; the spec records why it is in PHP.

## Verification

`composer analyse` (PHPStan level 10) clean, `composer lint`, `composer lint:md`, `composer lint:openapi` clean. The eight directly affected suites are 203/203. Remaining `composer test:quick` failures in the worktree are the known environmental ones (`ApiTestCase` resolves the production host without `.env.local`; the WS suite talks to a container serving the main checkout) — CI runs `composer ws:start` from the branch, so those assertions meet their own server there.

Closes #806 (section G)

## Summary by CodeRabbit

* **New Features**
  * Added strict validation for incoming WebSocket messages, including required fields, supported actions, value types, and message structure.
  * Added structured `protocolError` responses with machine-readable error codes and explanatory messages.
  * Preserved request and run identifiers in relevant responses.
  * Added clearer handling for invalid JSON, retired properties, unknown targets, malformed requests, and unexpected server errors.

* **Bug Fixes**
  * Prevented malformed or oversized messages from terminating processing.
  * Improved protection against exposing internal file paths or schema details in error responses.

### DIFF
--- a/docs/superpowers/plans/2026-08-20-strict-validation.md
+++ b/docs/superpowers/plans/2026-08-20-strict-validation.md
@@ -1,0 +1,1307 @@
+# Strict Validation and Typed Protocol Errors Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task.
+Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make a malformed WebSocket message impossible to crash the `Health` process, and answer it with a typed, machine-readable protocol error.
+
+**Architecture:** A published draft-07 schema, `jsondata/schemas/WebSocketMessage.json`, becomes the authority for what an inbound message may look like. A new
+`WebSocketMessageValidator` service validates every message against it before dispatch and — only for messages carrying a `requestId` — refuses properties the schema does not
+declare, reading the allowed names out of the schema rather than from a list of its own. Rejections become `type: "protocolError"` frames carrying a `ProtocolErrorCode`. A
+`\Throwable` backstop around the dispatch `switch` guarantees the daemon survives anything a handler does.
+
+**Tech Stack:** PHP 8.4, `swaggest/json-schema` v0.12.43 (draft-07), Ratchet/ReactPHP, PHPUnit 12, PHPStan level 10, PSR-12 via phpcs.
+
+**Spec:** `docs/superpowers/specs/2026-08-20-strict-validation-design.md` — read it before Task 1. Every "why" below is short because the spec carries the reasoning.
+
+## Global Constraints
+
+- **No v1 message that works today may stop working**, except ones that currently kill the process. This is the single most important constraint; Task 1's compatibility test
+is what enforces it.
+- Unknown-property rejection applies **only** when the message carries a `requestId`. Type and enum validation applies to **all** messages.
+- The schema declares `"$schema": "https://json-schema.org/draft-07/schema#"` and puts shapes under `definitions`, matching the other 21 schemas in `jsondata/schemas/`.
+- `swaggest/json-schema` v0.12.43 implements draft-07 only. **Do not use `unevaluatedProperties`, `$defs`, `$dynamicRef`, or any 2019-09+ keyword** — see spec §4 and issue #826.
+- The new error frame uses **`text`**, never `message`. `errorCode` is the new field.
+- Retired-property checks stay in `Health` and keep running **before** schema validation.
+- Tests drive `Health::onMessage()` with a real JSON message. Do not test the private validator by invoking it directly except where a step explicitly says to.
+- Every test class that builds a `Health` must `use HealthQueueIsolationTrait` and build via `$this->newHealth()`. A hand-written `new Health()` leaks queued HTTP requests
+that fire for real at PHPUnit shutdown.
+- Commits are GPG-signed and pre-commit hooks run `composer lint` / `lint:md`. Never pass `--no-verify`.
+- Run `composer test:quick`, never a bare `phpunit --exclude-group`.
+
+## File Structure
+
+| File                                                         | Responsibility                                                   |
+|--------------------------------------------------------------|------------------------------------------------------------------|
+| `jsondata/schemas/WebSocketMessage.json` (new)               | The authority: every inbound message shape, types, enums         |
+| `src/Enum/LitSchema.php` (modify)                            | One new case so the schema resolves by path like every other     |
+| `src/Enum/ProtocolErrorCode.php` (new)                       | The error-code vocabulary                                        |
+| `src/Services/WebSocketMessageValidator.php` (new)           | Schema validation + the `requestId`-gated unknown-property check |
+| `src/Health.php` (modify)                                    | Wire the validator in, emit typed errors, add the backstop       |
+| `phpunit_tests/Schemas/WebSocketMessageSchemaTest.php` (new) | The schema is correct and accepts what the shipped client sends  |
+| `phpunit_tests/HealthProtocolValidationTest.php` (new)       | The crash vectors are refused; codes are right; gate works       |
+
+`WebSocketMessageValidator` is a separate class rather than more methods on `Health` because `src/Health.php` is already 4108 lines, and validating an inbound message is a
+distinct responsibility with a clean interface. `src/Services/` is where `JwtService` lives.
+
+---
+
+### Task 1: The published schema and its compatibility test
+
+**Files:**
+
+- Create: `jsondata/schemas/WebSocketMessage.json`
+- Modify: `src/Enum/LitSchema.php` (add a case; extend `error()` and `fromURL()`)
+- Test: `phpunit_tests/Schemas/WebSocketMessageSchemaTest.php`
+
+**Interfaces:**
+
+- Consumes: nothing from earlier tasks.
+- Produces: `LitSchema::WEBSOCKET_MESSAGE` (path `/WebSocketMessage.json`), and a schema whose `definitions` keys are exactly `executeValidation`, `validateCalendarLegacy`,
+`validateCalendarTyped`, `executeUnitTest`, `runTest`, `cancelRun`, `validateSource`, `calendarIdentity`. Task 3 reads `definitions.<shape>.properties` by those names.
+
+**Background you need:** the shapes are already specified as `@phpstan-type` annotations in the `Health` class docblock (`src/Health.php:55-65`). PHPStan level 10 enforces
+them against the implementation, so they are the closest thing to a written contract that exists. Transcribe them; do not invent constraints.
+
+Two deliberate decisions, both of which you must not "improve":
+
+1. **`year` is `"type": "integer"`.** PHP's coercive typing means a client sending `"2024"` works today. Requiring an integer is a deliberate, narrow tightening; the shipped
+client sends numbers (`assets/js/index.js:686-693`), which Step 3's test proves.
+2. **`category` enums are closed.** A misspelled category currently survives to "Unable to detect schema" much later; closing the enum is the point of #806 §7. It is still an
+error either way — just earlier and legible.
+
+- [ ] **Step 1: Write the schema file**
+
+Create `jsondata/schemas/WebSocketMessage.json`:
+
+```json
+{
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "$id": "https://litcal.johnromanodorazio.com/api/dev/jsondata/schemas/WebSocketMessage.json",
+    "title": "Health WebSocket inbound message",
+    "description": "Every message the Health WebSocket endpoint accepts. Shapes are discriminated by `action`, and for `validateCalendar` additionally by whether `calendar` is a string (v1) or an object (v2). Properties a shape does not declare are refused only for messages carrying a `requestId`; see API issue #806 section G.",
+    "oneOf": [
+        { "$ref": "#/definitions/executeValidation" },
+        { "$ref": "#/definitions/validateCalendarLegacy" },
+        { "$ref": "#/definitions/validateCalendarTyped" },
+        { "$ref": "#/definitions/executeUnitTest" },
+        { "$ref": "#/definitions/runTest" },
+        { "$ref": "#/definitions/cancelRun" },
+        { "$ref": "#/definitions/validateSource" }
+    ],
+    "definitions": {
+        "correlationId": {
+            "type": "string",
+            "pattern": "^[A-Za-z0-9_-]{1,64}$",
+            "description": "An opaque client-minted handle the server echoes back. One alphabet for runToken and requestId alike."
+        },
+        "calendarIdentity": {
+            "type": "object",
+            "required": ["kind", "rite"],
+            "properties": {
+                "kind": { "enum": ["general", "national", "diocesan", "rite"] },
+                "id": { "type": "string" },
+                "rite": { "type": "string" }
+            }
+        },
+        "executeValidation": {
+            "type": "object",
+            "required": ["action", "category", "validate"],
+            "oneOf": [
+                { "required": ["sourceFile"] },
+                { "required": ["sourceFolder"] }
+            ],
+            "properties": {
+                "action": { "const": "executeValidation" },
+                "category": { "enum": ["universalcalendar", "sourceDataCheck", "resourceDataCheck"] },
+                "validate": { "type": "string" },
+                "sourceFile": { "type": "string" },
+                "sourceFolder": { "type": "string" },
+                "responsetype": { "type": "string" },
+                "rite": { "type": "string" },
+                "runToken": { "$ref": "#/definitions/correlationId" },
+                "requestId": { "$ref": "#/definitions/correlationId" }
+            }
+        },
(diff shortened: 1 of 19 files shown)


JohnRDOrazio:
All four items verified against the code. Two were real and are fixed; two are declined with evidence.

## Fixed — coverage was misattributed, not missing (618b896b)

The 1.01% on `WebSocketMessageValidator` is not untested code. `#[CoversClass(Health::class)]` **restricts** a test class's coverage attribution to `Health`, so every line the validator executes while being driven through `Health::onMessage()` went uncredited.

Measured, changing no tests at all — only adding the attribute:

| | lines | methods |
|---|---|---|
| before | 1.01% | — |
| after (from `HealthProtocolValidationTest` alone) | **87.64%** (78/89) | **61.54%** (8/13) |

Writing new tests to move that number would have been treating the symptom. An audit for the same mistake found it in **nine** classes in total, all now carrying the second attribute: `HealthProtocolValidationTest`, `HealthHelpersTest`, `HealthCancelRunTest`, `HealthRiteRequestPathTest`, `HealthTypedCalendarTest`, `HealthValidateSourceTest`, `HealthSourceFileFailureArmsTest`, `HealthTerminalFrameTest`, `HealthCorrelationTest`.

## Fixed — the unknown-property gate did not reach nested objects (ad7448bf)

Correct finding, and reproduced before fixing: with a `requestId` present, `{"action":"validateSource","target":{"id":"temporale:roman","extra":"x"},"requestId":"r1"}` was **accepted** and ran the check. The gate iterates root-level keys only, and the nested objects declared `properties` without `additionalProperties`.

Fixed in the schema rather than by recursing in PHP: `"additionalProperties": false` on the `calendarIdentity` definition and on `validateSource`'s inline `target`. The draft-07 limitation that forced the *root* gate into PHP does not apply here — those subschemas declare their own `properties` in the same schema object, so `additionalProperties` sees them.

One asymmetry, deliberate and recorded in a schema `description` so nobody "fixes" it into consistency: nested strictness is unconditional while root strictness stays gated on `requestId`. That is safe because `calendarIdentity` and `target` appear only in v2 shapes, and the shipped UnitTestInterface sends none of them — no `validateSource`, `runTest`, `responseFormat` or `requestId` anywhere in `assets/js/*.js` at `9126ad9`.

## Declined — requiring `id` on `calendarIdentity` in the schema

The handler already refuses this, safely, with a better message. Probed on the current branch:

```json
{"action":"runTest","test":"T","calendar":{"kind":"national","rite":"roman"},"year":2024,"requestId":"r4"}
→ {"errorCode":"invalid_message","text":"calendar.id is required for kind national."}
```

Moving the check into the schema would replace that sentence with a generic schema complaint for no safety gain — `$id` is validated as string-or-null before use, so there is no crash path. This branch already spent two fix rounds undoing exactly that regression when schema validation pre-empted hand-written diagnostics.

The companion suggestion — that a `general` id must be `roman` — is not a rule anywhere in `resolveCalendarIdentity()`. Adding it would make the published schema stricter than the server, which is the one thing a contract must never be.

## Declined — `#[Group('slow')]`

Measured on this branch:

| class | time |
|---|---|
| `WebSocketMessageSchemaTest` | 228 ms |
| `HealthProtocolValidationTest` | 212 ms |
| `HealthHelpersTest` | 189 ms |

`CLAUDE.md` reserves that group for tests with *measurable* runtime cost — multi-year calendar calculations, rate-limit window waits, full-schema-corpus validation. Marking these would drop them from `composer test:quick`, which is the suite most likely to catch a regression in this exact code.

## Verification

`composer analyse` (PHPStan level 10) clean, `composer lint` clean, 458 tests / 2606 assertions across the Health and schema suites. The client-fixture regression net (`WebSocketMessageSchemaTest`, 28/28) passes unchanged.

PR #829, opened 2026-08-21
fix(health): the WebSocket server cannot start on development

JohnRDOrazio:
**`development` is broken right now: `public/LitCalTestServer.php` dies on startup, every time.**

```text
PHP Fatal error: Uncaught Error: Typed static property
LiturgicalCalendar\Api\Router::$apiFilePath must not be accessed before initialization
  in src/Enum/JsonData.php:476
#0 src/Enum/LitSchema.php(33):        JsonData->path()
#1 src/Services/WebSocketMessageValidator.php(82): LitSchema->path()
#2 src/Health.php(301):               WebSocketMessageValidator->__construct()
#3 public/LitCalTestServer.php(139):  Health->__construct()
```

`Health::__construct()` now builds a `WebSocketMessageValidator`, whose constructor resolves the message schema's path. `Router::$apiFilePath` is a **typed** static, so reading it before assignment is a fatal `Error` rather than a null. The entry point calls `new Health()` with nothing initialised — `Router::getApiPaths()` ran in `onOpen()`, on the first client *connection*, which is far too late.

Introduced by #828, which merged green.

## Why the whole suite missed it

`phpunit_tests/bootstrap.php` initialises the paths before any test runs. So all 447 in-process tests construct a `Health` in an order production never uses, and **no test written inside that process could have failed** — however carefully written. The failure lives in the gap between the harness's setup and the entry point's.

That is what `HealthConstructionOrderTest` addresses: it runs a fixture in a **subprocess** that requires nothing but the autoloader, exactly as the entry point does, and asserts it survives. Verified to reproduce the fatal with this fix reverted.

## The guard is load-bearing

The obvious fix — call `Router::getApiPaths()` in the constructor — **turned a green suite red.** `getApiPaths()` *overwrites* the path, and several tests deliberately point it at a fixture tree before building a `Health` (`BrokenInventoryTrait` is the documented case). `HealthValidateSourceTest` went from 18/18 to a failure.

So it is guarded:

```php
if (false === isset(Router::$apiFilePath)) {
    Router::getApiPaths();
}
```

`isset()` on an uninitialised typed static is `false` and does not throw — precisely the "initialise only if nobody has" test needed, and it leaves deliberate overrides alone. I found this by running the suite after the first attempt, not by reading the diff.

## Verification

- `HealthConstructionOrderTest` passes with the fix, **fails with it reverted** (the exact production fatal).
- `HealthValidateSourceTest` back to 18/18.
- `composer test:quick`: 2904 tests, **0 failures**. The 16 remaining errors are all `WebSocket/*` suites, which need a live server on `:8082` — currently down, because this bug is what prevents it starting. They pass in CI, which starts the server from the branch.
- `composer analyse` (PHPStan level 10) and `composer lint` clean.

## How it was found

By restarting the API container after merging #828, to pick up the new code — the WebSocket server is a long-running ReactPHP process, so unlike the `php -S` workers it does not re-read source on its own. It never came back up. That restart is the only reason this surfaced today rather than at the next deploy.

Refs #806, #828

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved health-check initialization when the application starts without prior setup.
  * Preserved existing configuration overrides while ensuring required routing information is available.
* **Tests**
  * Added coverage confirming health checks can be constructed in a clean process.
  * Added validation that startup completes successfully without initialization errors.

### DIFF
--- a/phpunit_tests/HealthConstructionOrderTest.php
+++ b/phpunit_tests/HealthConstructionOrderTest.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests;
+
+use LiturgicalCalendar\Api\Health;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * `new Health()` must work with nothing initialised first, because that is what the entry point does.
+ *
+ * **This test exists because the suite could not see the bug it guards.** `Health::__construct()`
+ * builds a `WebSocketMessageValidator`, whose constructor resolves the message schema's path through
+ * `Router::$apiFilePath` — a *typed* static, so reading it before assignment is a fatal `Error`, not
+ * a null. `public/LitCalTestServer.php` calls `new Health()` without calling `Router::getApiPaths()`
+ * first; `Health::onOpen()` was where paths got initialised, on the first client connection. So the
+ * WebSocket server died at startup, every time, on a code path 447 in-process tests never touched.
+ *
+ * They never touched it because `phpunit_tests/bootstrap.php` initialises the paths before any test
+ * runs. Every test therefore constructs a `Health` in an order production never uses. A test written
+ * inside this process — however carefully — would assert against that initialised state and pass.
+ *
+ * Hence the subprocess. The fixture requires nothing but the autoloader, exactly as the entry point
+ * does, and this asserts it survives. It is the cheapest thing that could have caught the failure.
+ */
+#[CoversClass(Health::class)]
+final class HealthConstructionOrderTest extends TestCase
+{
+    public function testAHealthCanBeConstructedWithNothingInitialisedFirst(): void
+    {
+        $projectRoot = dirname(__DIR__);
+        $fixture     = $projectRoot . '/phpunit_tests/fixtures/construct-health-bare.php';
+        self::assertFileExists($fixture);
+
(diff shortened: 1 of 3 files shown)


JohnRDOrazio:
Verified against the code and measured. **Declining**, and this one is worth the detail because the suggestion is more plausible here than usual — this test spawns a subprocess, which is a genuinely different cost profile from the in-process suites.

**Measured on this branch, three consecutive runs:**

| test | time |
|-------------------------------|-----------|
| `HealthConstructionOrderTest` (subprocess) | 224 ms / 220 ms / 248 ms |
| `HealthProtocolValidationTest` (in-process, for comparison) | 256 ms |

The subprocess is *faster* than a typical in-process Health suite — spawning PHP costs almost nothing next to PHPUnit's own boot. `CLAUDE.md` reserves `slow` for tests with measurable runtime cost, and the suites that actually carry it are multi-year Ambrosian calendar computations and rate-limit window waits, which are orders of magnitude away from this.

**The specific reason it matters here.** `#[Group('slow')]` removes a test from `composer test:quick`. This is the **only** test in the suite that can catch entry-point initialisation-order bugs — every other test runs inside a PHPUnit process whose bootstrap has already initialised `Router::$apiFilePath`, which is exactly why 447 of them passed while the WebSocket server could not start at all. Excluding it from the routine suite would remove the single guard against the regression this PR exists to fix, four days before someone reintroduces it.

If it ever does become slow — say the fixture grows into starting a real server and binding a port — that is the moment to reconsider, and the docblock says what it is for so the next reader can make that call on evidence.

PR #830, opened 2026-08-21
ops: capture the deploy sentinel units, and restart the WebSocket server on deploy

JohnRDOrazio:
Two things: the deploy mechanism on `catholicdigitalcommons.org` now lives in the repo, and it stops being half a deploy.

**Nothing here is applied to the server.** Installing is a documented manual step.

## The mechanism existed only on that host

Two systemd units, a root-owned shell script and a path watcher — reconstructible only by SSH archaeology. The deploy user is chrooted and cannot `sudo`, so a deploy drops `<approot>/tmp/restart.txt` and a root-owned path unit does the privileged part. That is a good design and it was written down nowhere.

Captured verbatim into `deploy/`, with header comments saying what each file is for, plus `docs/ops/deploy-sentinel-runbook.md`.

## It was only reloading php-fpm

The script did one thing: `systemctl reload plesk-php84-fpm`. Nothing restarted `litcal-websocket.service` — not the deploy, not anything else — so it kept serving whatever it loaded at boot.

Measured on the server, 2026-08-21:

| | |
|---------------------------------------|-----------------------------------------------|
| deployed `api/dev/src/Health.php` | modified **07:33 UTC** that morning |
| `litcal-websocket` process | started **2026-08-20 06:31 UTC**, `NRestarts=0` |

The REST API was current and the WebSocket endpoint was **25 hours stale**. All of API#806 section G — a new message contract, schema validation, typed error frames — was on disk and inert.

`public/LitCalTestServer.php` is a long-running ReactPHP process: it loads `src/` into memory once and never re-reads it. So `git pull` genuinely deploys only half of any change touching `src/Health.php`. The same shape bites locally — `composer start` re-reads source per request, `composer ws:start` does not.

## The crash-loop check is the part worth reviewing

The obvious change is "also restart the WebSocket unit". That alone would be a trap.

The unit is `Restart=always, RestartSec=5`, so a process that dies at startup **never reports failure** — it respawns every five seconds while the deploy reports success and `systemctl is-active` looks fine. PR #828 shipped exactly that fatal: `Health::__construct()` could not run at all. The service survived only because nothing ever restarted it, which is the same reason it was serving stale code.

So after restarting, the script waits past one restart interval and compares `NRestarts`. If it moved, the deploy fails loudly with the journal command to run:

```text
litcal-websocket.service is CRASH-LOOPING after deploy (NRestarts 0->2);
check: journalctl -u litcal-websocket.service -n 50
```

Two further deliberate choices:

- **The WebSocket restart is skipped if the php-fpm reload failed.** A broken deploy should not also drop live connections.
- **Whether the api/dev sentinel fired is recorded before the reload**, since the reload clears sentinels on success and this must not depend on that ordering.

## Verification

`sh -n` clean (POSIX `sh`, matching the shebang). `composer lint:md` clean. The script is not executed by CI — it needs root, systemd and that host — so the runbook carries an explicit end-to-end check instead: `touch` the sentinel, then read `journalctl -t litcal-fpm-reload`. A healthy api/dev deploy logs `sentinel seen … (api/dev=1)` → `restarting litcal-websocket.service` → `restarted and stable` → `reload OK; sentinels cleared`.

## After merging

The live endpoint is still stale until the unit is restarted once by hand — the new script only takes effect from the *next* deploy, and only after it is installed:

```bash
sudo install -m 0755 -o root -g root deploy/sbin/litcal-fpm-reload.sh /usr/local/sbin/litcal-fpm-reload.sh
sudo systemctl restart litcal-websocket
```

Refs #828, #829

## Summary by CodeRabbit

* **New Features**
  * Added automated service reloads following deployments.
  * Added WebSocket service management with automatic restart and crash-loop detection.
  * Deployment signals are cleared only after successful service recovery; failures remain eligible for retry.

* **Documentation**
  * Added operational guidance for configuring, verifying, and troubleshooting deployment-triggered reloads and WebSocket restarts.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+# Render the systemd templates against /etc/litcal-deploy.env and install everything.
+# Run as root on the deployment host, from a checkout of this repository:
+#
+#     sudo deploy/install.sh
+#
+# The templates in deploy/systemd/*.in carry placeholders, never real values — this
+# repository is public. The real paths, accounts and unit names live only in the
+# config file, which is not committed.
+set -eu
+
+CONFIG="${LITCAL_DEPLOY_ENV:-/etc/litcal-deploy.env}"
+SRC_DIR="$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd)"
+UNIT_DIR=/etc/systemd/system
+
+if [ ! -r "$CONFIG" ]; then
+  echo "error: $CONFIG missing. Copy deploy/litcal-deploy.env.example there and fill it in." >&2
+  exit 1
+fi
+# shellcheck source=/dev/null
+. "$CONFIG"
+
+for var in API_ROOT PHP_BIN RUN_USER RUN_GROUP FPM_UNIT WS_UNIT; do
+  eval "value=\${$var:-}"
+  if [ -z "$value" ]; then
+    echo "error: $var not set in $CONFIG" >&2
+    exit 1
+  fi
+done
+
+# One PathChanged= per app that is actually deployed here, so a host watching a
+# subset does not get a unit referring to directories it does not have.
+path_changed_lines=""
+for root in "$API_ROOT" "${FRONTEND_ROOT:-}" "${FRONTEND_STAGING_ROOT:-}" "${TESTS_ROOT:-}"; do
+  [ -n "$root" ] || continue
+  path_changed_lines="${path_changed_lines}PathChanged=${root}/tmp/restart.txt
+"
+done
+
+render() {
+  sed \
+    -e "s|@API_ROOT@|${API_ROOT}|g" \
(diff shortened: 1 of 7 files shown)


JohnRDOrazio:
Closing this and reopening from a clean branch.

This PR committed the server's absolute paths, the deploy account and its group into a **public** repository — the same class of disclosure #827 is removing from WebSocket frames, except permanent and indexed. Stripping the filesystem root off every frame so an unauthenticated client cannot learn the deployment layout is pointless if the layout is committed to a public repo instead.

Rather than force-pushing over it, the branch is being deleted and the work reopened as a single commit whose history never contains those values. Replacement PR linked below.

JohnRDOrazio:
Superseded by #831, which carries the same work as templates. Keeping this closed rather than deleting the branch again: the diff above now renders the sanitised commit, which is the point of the force-push.

PR #831, opened 2026-08-21
ops: capture the deploy sentinel units as templates, restart the WebSocket server

JohnRDOrazio:
Replaces #830, which committed the server's real paths, deploy account and group into this **public** repository. That is the same class of disclosure #827 is removing from WebSocket frames — except permanent and indexed. #830 was closed and its branch deleted rather than force-pushed; this branch is a single commit whose history never contains those values.

## Everything host-specific is a template

- `deploy/systemd/*.in` carry `@PLACEHOLDER@` tokens — no real path, account or unit name.
- `deploy/sbin/litcal-fpm-reload.sh` reads every host value from `/etc/litcal-deploy.env`, which is not committed. It **refuses to run** if that file is missing rather than guessing.
- `deploy/litcal-deploy.env.example` documents the variables with obviously-illustrative values (`example.org`, `www-data`, `/usr/bin/php`).
- `deploy/install.sh` renders the templates and installs everything.

The `PathChanged=` list is generated per host: any app root left empty in the config is dropped from the watch list, so a host that deploys a subset does not get a unit referencing directories it lacks.

## Why capture this at all

The deploy mechanism existed only on the production host — two systemd units, a root-owned script, a path watcher — reconstructible only by SSH archaeology. The deploy user is chrooted and cannot `sudo`, so a deploy drops `<approot>/tmp/restart.txt` and a root-owned path unit does the privileged part. Good design, written down nowhere.

## It was only reloading php-fpm

Nothing restarted the WebSocket unit — not the deploy, not anything else. Measured on the host: the deployed `src/Health.php` was **~25 hours newer** than the running process (`NRestarts=0`), so an entire protocol change (API#806 section G) was on disk and inert.

`public/LitCalTestServer.php` is a long-running ReactPHP process: it loads `src/` into memory once and never re-reads it. So `git pull` genuinely deploys only half of any change touching it. The same shape bites locally — `composer start` re-reads source per request, `composer ws:start` does not.

## The crash-loop check is the part worth reviewing

"Also restart the WebSocket unit" alone would be a trap. The unit is `Restart=always`, so a process that dies at startup **never reports failure** — it respawns every `RestartSec` while the deploy reports success and `systemctl is-active` looks fine. PR #828 shipped exactly that fatal; the service survived only because nothing ever restarted it, which is the same reason it was serving stale code.

So after restarting, the script waits past one restart interval and compares `NRestarts`, failing the deploy loudly with the journal command to run.

Two further deliberate choices: the WebSocket restart is **skipped if the php-fpm reload failed** (a broken deploy should not also drop live connections), and whether the API sentinel fired is recorded **before** the reload, since the reload clears sentinels on success.

## Verification

`sh -n` clean on both scripts (POSIX `sh`, matching the shebangs). `composer lint:md` clean. Template rendering smoke-tested against a fake config, producing correct `User=`, `Group=`, `WorkingDirectory=` and `ExecStart=` lines.

Grepped the working tree for every real value — domain, account, group, PHP prefix, FPM unit name, vhost root: **zero occurrences** under `deploy/` and `docs/ops/`.

CI cannot exercise this — it needs root, systemd and that host — so the runbook carries a manual end-to-end check instead: `touch` the sentinel, then read `journalctl -t litcal-fpm-reload`.

## After merging

The live endpoint stays stale until the unit is restarted once by hand; the new script only takes effect from the *next* deploy, and only after `sudo deploy/install.sh` has been run with `/etc/litcal-deploy.env` in place.

Refs #828, #829, #827

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -57,7 +57,13 @@ chmod 0644 "${UNIT_DIR}/litcal-fpm-reload.path" "${UNIT_DIR}/litcal-fpm-reload.s
 install -m 0755 -o root -g root "${SRC_DIR}/sbin/litcal-fpm-reload.sh" /usr/local/sbin/litcal-fpm-reload.sh
 
 systemctl daemon-reload
-systemctl enable --now litcal-fpm-reload.path "$WS_UNIT"
+systemctl enable litcal-fpm-reload.path "$WS_UNIT"
+
+# Restart, not `enable --now`. `--now` starts a unit that is stopped and does nothing at
+# all to one already running, so re-running this installer with updated unit content
+# would leave the old definition live and report success — the same silent no-op this
+# script exists to remove from the deploy path.
+systemctl restart litcal-fpm-reload.path "$WS_UNIT"
 
 echo "installed. verify with:"
 echo "  systemctl status litcal-fpm-reload.path"
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+# Render the systemd templates against /etc/litcal-deploy.env and install everything.
+# Run as root on the deployment host, from a checkout of this repository:
+#
+#     sudo deploy/install.sh
+#
+# The templates in deploy/systemd/*.in carry placeholders, never real values — this
+# repository is public. The real paths, accounts and unit names live only in the
+# config file, which is not committed.
+set -eu
+
+CONFIG="${LITCAL_DEPLOY_ENV:-/etc/litcal-deploy.env}"
+SRC_DIR="$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd)"
+UNIT_DIR=/etc/systemd/system
+
+if [ ! -r "$CONFIG" ]; then
+  echo "error: $CONFIG missing. Copy deploy/litcal-deploy.env.example there and fill it in." >&2
+  exit 1
+fi
+# shellcheck source=/dev/null
+. "$CONFIG"
+
+for var in API_ROOT PHP_BIN RUN_USER RUN_GROUP FPM_UNIT WS_UNIT; do
+  eval "value=\${$var:-}"
+  if [ -z "$value" ]; then
+    echo "error: $var not set in $CONFIG" >&2
+    exit 1
+  fi
+done
+
+# One PathChanged= per app that is actually deployed here, so a host watching a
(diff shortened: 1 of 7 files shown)


JohnRDOrazio:
All five findings verified against the code and **all five fixed** (`a93c7d29`). Every one was the same disease this script exists to cure: a deploy that silently does not take effect.

## A deploy landing mid-run lost its sentinel

Correct, and the sharpest of the five. The sentinel was tested at the start and deleted at the end — 8+ seconds later, past a service restart. A second deploy inside that window dropped a fresh sentinel that this run then removed, so its changes sat on disk and were never reloaded.

Sentinels are now **claimed by rename** before any work. `mv` within a filesystem is atomic, a concurrent deploy writes a new `restart.txt` this run cannot see, and systemd queues another run for it. Failure paths rename the claim back, so "sentinels kept for retry" still means what it says.

## The crash-loop check failed open

Also correct, and it defeated the check's entire purpose. Both `NRestarts` reads ended `|| echo 0`, so a failed query made *both* sides 0, the comparison said "no change", and a crash-looping service was reported stable. Now validated as numeric; anything else logs, keeps the sentinels and exits 1.

## Cleanup was unquoted and unchecked

`rm -f $SENTINELS` split a path containing a space into two that do not exist, and success was logged whatever `rm` returned. Now removed one at a time, quoted, with the result checked — claiming a deploy is done while a sentinel survives would leave the trigger armed.

## `enable --now` does not restart a running unit

Correct and, I think, the most consequential for day-to-day use: re-running the installer with **updated** unit content would have left the old definition live and reported success. It now enables *and* restarts, with a comment saying why the restart is not redundant.

## The runbook used `$API_ROOT` undefined

Fixed — the verification block now sources `/etc/litcal-deploy.env` first, so the commands work as written rather than needing mental substitution.

## Verification

CI cannot exercise this script (root, systemd, that host), so I ran it against stubbed `systemctl`/`logger` with fake sentinels:

| scenario | result |
|-------------------------------------|-------------------------------------------------|
| happy path | 4 log lines, exit 0, sentinels cleared |
| crash loop (`NRestarts` increments) | detected, exit 1, **sentinel restored for retry** |
| `NRestarts` unreadable | fails closed, exit 1, sentinel kept |
| deploy lands mid-run | second sentinel **survives** |

The last one was also checked against a reproduction of the old shape, which **eats it** — so the fix is doing the work, not the test setup.

PR #832, opened 2026-08-21
fix(health): stop WebSocket frames leaking the server's filesystem path

### LINKED ISSUE #827: Health WebSocket frames leak the server's absolute filesystem path to unauthenticated clients

JohnRDOrazio:
Closes #827.

`Health`'s frames sent the server's absolute filesystem path to any client on an unauthenticated WebSocket:

```text
… was successfully validated against the Schema /home/…/jsondata/schemas/LitCal.json
```

## Two layers, because there were two sources

**1. Schema paths → schema names** (`d3e579be`). Eight interpolation sites, one of them found empirically while building the required schema-failure test: `validateDataAgainstSchema()`'s `swaggest/json-schema` error text embeds the schema's absolute import path in its `$ref[...]` trace. That is the same leak API#806 section G already closed on the *rejection* path — it was still open on the success path.

**2. Data paths → project-relative** (`f709b734`). The first layer was not enough, and the residual is the interesting part. A v2 `validateSource` sends **only an opaque id** and got back three frames carrying the absolute root, because the *data* path was resolved by the server from the inventory:

```text
{"action":"validateSource","target":{"id":"temporale:roman"}}
→ The Data file /home/…/LCAPI-827/jsondata/sourcedata/rite/roman/…/propriumdetempore.json exists
```

The rule that separates a leak from a useful diagnostic: **echo what the client gave you, not what you resolved.** A v1 `executeValidation` that sent `sourceFile: "jsondata/x.json"` should still see it. So rather than tracking provenance, `stripProjectRoot()` removes `Router::$apiFilePath` from `text` and from every string in `details` — correct for both cases with one transform. The result is *better* than a redaction:

```text
The Data file jsondata/sourcedata/rite/roman/missals/propriumdetempore/propriumdetempore.json exists
```

Stable across hosts, meaningful to a client, disclosing nothing.

## One choke point, not eight patches

`stripProjectRoot()` lives in `Health::sendMessage()`. Review verified that claim rather than accepting it: `grep -n '\->send(' src/Health.php` returns exactly one hit, inside `sendMessage()` itself, and every emitter funnels through it. This leak has now surfaced twice in different frame families, so the fix belongs where a future frame type cannot bypass it.

## Three tests were asserting the leak as correct behaviour

That is why section G's narrower guard did not catch this. `HealthFolderStepResultTest` and two `HealthRiteSourcePathTest` cases each asserted an absolute path was present verbatim. All three now assert the project-relative form **and** add `assertStringNotContainsString(rtrim(Router::$apiFilePath, '/'), …)` — strictly more assertions over strictly more behaviour, not a failing test made to pass.

The section G guard is widened accordingly: from "no *rejection* text contains the root" to **no frame `Health` emits, of any kind, contains it** — renamed `testNoFrameEverLeaksTheServerFilesystemPath`, with the `temporale:roman` regression fixture pinned explicitly.

## Verification

Independent review probed 6 end-to-end scenarios and 7 direct edge cases: client-supplied relative paths echoed verbatim, `resourceDataCheck` URLs byte-identical, non-path `details` untouched, terminal frames and malformed shapes handled without error. `Route::*->path()` was confirmed to be a **URL**, never a filesystem path, so those interpolations are correctly untouched — which also corrects #827's text, where I wrote that one site leaked two absolute paths; it leaks one, alongside a URL.

Cost measured at **1.37 µs per frame** (200k calls in 274 ms) — negligible against WebSocket I/O.

459 tests / 2654 assertions in the Health and schema suites, PHPStan level 10 clean, phpcs clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed server filesystem paths from health and validation messages.
  * Validation results now display safe, project-relative filenames instead of absolute paths.
  * Improved protection across source, calendar, folder, file, and schema validation responses.

* **Tests**
  * Added coverage confirming outgoing frames never expose the server’s filesystem location.
  * Updated path assertions to match the safer relative-path format.

### DIFF
--- a/phpunit_tests/HealthFolderStepResultTest.php
+++ b/phpunit_tests/HealthFolderStepResultTest.php
@@ -248,7 +248,11 @@ final class HealthFolderStepResultTest extends TestCase
     {
         Router::getApiPaths();
 
-        $derived = strtr(JsonData::NATIONAL_CALENDAR_I18N_FOLDER->path(), ['{nation}' => 'ZZ']);
+        $derivedAbsolute = strtr(JsonData::NATIONAL_CALENDAR_I18N_FOLDER->path(), ['{nation}' => 'ZZ']);
+        // #827: a frame quotes the server-derived folder project-relative, never with the server's
+        // absolute filesystem root, so the assertion below has to match what actually reaches the
+        // wire rather than the raw JsonData::*->path() this is derived from.
+        $derived = substr($derivedAbsolute, strlen(Router::$apiFilePath));
 
         $conn   = self::createStubConnection(4);
         $health = $this->newHealth();
@@ -277,6 +281,11 @@ final class HealthFolderStepResultTest extends TestCase
                 $derived,
                 (string) $frame->text,
                 'and must report on the folder the server derived from the slug and actually read'
+            );
+            self::assertStringNotContainsString(
+                rtrim(Router::$apiFilePath, '/'),
+                (string) $frame->text,
+                'the frame must not leak the server filesystem root while reporting the derived folder (#827)'
             );
         }
     }
--- a/phpunit_tests/HealthProtocolValidationTest.php
+++ b/phpunit_tests/HealthProtocolValidationTest.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace LiturgicalCalendar\Tests;
 
-use LiturgicalCalendar\Api\Enum\JsonData;
 use LiturgicalCalendar\Api\Enum\ProtocolErrorCode;
 use LiturgicalCalendar\Api\Health;
 use LiturgicalCalendar\Api\Models\ValidationsPath\CheckableInventory;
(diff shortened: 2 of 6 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
## Fixed — the leak assertion only inspected `text` (`7be08ae6`)

Correct finding, and it went further than it looked. `stripProjectRoot()` sanitises **both** `text` and `details`, so the closure covering only `text` left the details half of the fix unfalsifiable — remove it from production and the test stayed green.

**Extending the closure was not enough on its own.** I checked by mutation before claiming a fix: with the production `details` branch deleted, the extended test *still passed*, because no scenario in it put a path in `details`. An assertion that cannot fail is not coverage, so this would have shipped as decoration.

The test now drives the one arm that genuinely does. `runValidationSteps()`'s folder branch collects `"unreadable i18n json file {name}: "` followed by the read's own `"Unable to read file: {absolute path}"`, and that lands in `details`. The fixture is a **directory** named like an i18n file: the `*.json` glob matches it and the `is_file()` check added for #822 refuses it — so it reproduces deterministically without depending on permissions, which matters because `chmod 000` is a no-op when the suite runs as root, as it does in CI. It is created under the project root so the reported path actually begins with `Router::$apiFilePath`, and is removed in a `finally`.

Verified both ways:

```text
production details-stripping removed → FAILS:
  folder check with an unreadable file (details carries the path) frame #0 leaked the
  server filesystem path in details[0]: unreadable i18n json file it.json:
  Unable to read file: /…/jsondata/sourcedata/.leak-probe/it.json (no such file)

restored → OK (1 test, 42 assertions)   [was 33]
```

## Declined — `#[Group('slow')]` on the four classes

Measured on this branch:

| class | time |
|--------------------------------|--------|
| `HealthProtocolValidationTest` | 255 ms |
| `HealthFolderStepResultTest`   | 193 ms |
| `HealthRiteSourcePathTest`     | 244 ms |
| `HealthValidateSourceTest`     | 952 ms |

`CLAUDE.md` reserves that group for tests with *measurable* runtime cost; the suites that carry it are multi-year Ambrosian calendar computations and rate-limit window waits, orders of magnitude beyond these. Marking them would drop them from `composer test:quick` — including the leak assertion above, which is the only guard against this class of regression, and which this PR exists to install.

PHPStan level 10 and phpcs clean; 459 tests in the Health and schema suites.

PR #834, opened 2026-08-21
fix(health): a URL check must not report `exists` for a 4xx or 5xx

### LINKED ISSUE #833: A URL-based check reports `exists = success` for any HTTP status, including 4xx and 5xx
Observed in the live test interface: Ambrosian rite years **1970–1975** show a green **"data exists"**, then fail later steps. The data does not exist — the API correctly refuses those years — so the check is reporting something untrue, and pointing the reader at the wrong step.

## What the API actually says

```console
$ curl -s -o /dev/null -w '%{http_code}\n' '…/calendar/ambrosian/1970'
400
```

```json
{
  "type": "…rfc9110#name-400-bad-request",
  "title": "Bad Request",
  "status": 400,
  "detail": "The Ambrosian rite is only available from 1976 onward (the first reformed Ambrosian Missal); requested year 1970."
}
```

The boundary **is** enforced, with an exemplary problem document that says precisely what is wrong. None of it reaches the client.

## Why the check reports success

Two facts compose into the bug:

1. `Health`'s Guzzle client is built with **`'http_errors' => false`** (`src/Health.php:330`), so a 400 resolves the promise as a success rather than rejecting.
2. `cachedGet()` resolves with `array{data: string, fromCache: bool}` — **the status code is not propagated at all.** `validateCalendar()`'s fulfil handler therefore cannot inspect it even in principle, and unconditionally emits:

```php
$this->sendCalendarStepResult($to, $calendar, $year, Step::EXISTS, Status::PASS,
    "The $category of $calendar for the year $year exists", …);
```

What the user then sees is worse than a plain failure: `exists` passes; `json-valid` **also** passes, because a problem document is perfectly valid JSON; and `schema-valid` fails. The reader is told "the schema is wrong" when the truth is "the endpoint refused the request".

## Scope: every URL-based check, not just calendars

The same shape applies to `executeValidation` with `category: "resourceDataCheck"`, whose `sourceFile` is an endpoint URL and which reaches `processValidationData()` — that also emits `Step::EXISTS, Status::PASS` unconditionally. Any resource endpoint returning 404 or 500 is currently reported as existing.

## Relationship to #822

This is the same defect — a check reporting existence it never established — in the half that #822's fix structurally cannot reach. #822 was about `react/filesystem` resolving a missing file with zero bytes, and it was fixed by stat-ing before the read in `cachedFileGetContents()`. That is the **filesystem** path. Calendar years and resource endpoints are **URLs**: there is nothing to stat, and the HTTP status is the only signal that the thing exists. It was discarded.

## Relationship to UnitTestInterface#52

UTI#52 will stop the client requesting years below a rite's lower bound, which removes this particular symptom from the interface. It does not fix this: any 4xx or 5xx from any URL-based check would still be reported as `exists = success`. The two are worth doing independently — one stops asking a bad question, the other stops giving a false answer.

## Suggested fix

1. Propagate the status code out of `cachedGet()` — e.g. resolve with `array{data: string, status: int, fromCache: bool}`. Everything else follows from having the information.
2. Gate the `exists` frame on a 2xx. On anything else, emit `exists` as **failed**, quoting the status and the problem document's `detail`, which already explains the cause better than any text we would write:

   ```text
   ambrosian 1970 was refused with HTTP 400: The Ambrosian rite is only available
   from 1976 onward (the first reformed Ambrosian Missal); requested year 1970.
   ```

3. Fail the remaining steps rather than skipping them, per the one-frame-per-published-step contract established in #821.

A test should drive a URL that really returns 4xx and assert `exists` fails — the current suite has no such case, which is why this survived alongside #822's fix.

JohnRDOrazio:
Closes #833.

Reported from the live test interface: Ambrosian years **1970–1975** showed a green **"data exists"** while the API was correctly returning 400 with a problem document explaining the rite's lower bound.

## Two facts composed into it

1. `Health`'s Guzzle client is built with **`'http_errors' => false`**, so a 400 resolves as a *success*.
2. `cachedGet()` resolved with `{data, fromCache}` — **the status was never propagated**, so no consumer could check it even in principle.

`validateCalendar()`'s fulfil handler therefore emitted `Step::EXISTS, Status::PASS` unconditionally.

What a user saw was worse than a plain failure: `exists` passed, `json-valid` **also** passed — a problem document is perfectly valid JSON — and `schema-valid` failed. So the interface said *"the schema is wrong"* when the truth was *"the endpoint refused the request"*.

## The fix

`cachedGet()` now resolves with `{data, status, fromCache}`; the `exists` frame is gated on a 2xx at each consumer; and a non-2xx **fails the remaining steps** rather than skipping them, per the one-frame-per-published-step contract from #821. The failure quotes the problem document's own `detail`, which explains the cause better than anything we would write:

```text
ambrosian 1970 was refused with HTTP 400: The Ambrosian rite is only available
from 1976 onward (the first reformed Ambrosian Missal); requested year 1970.
```

## The cache was the trap

A cached 4xx replaying without its status would reopen the bug on the **second** run of the same check — exactly what a user hits re-running the interface. Entries now carry `{status, body}`.

Beyond that: entries written by the *old* code are bare bodies with no status. They decode to `null` and are treated as a miss, so the URL is re-fetched, rather than being assumed 200 — which would have silently reopened the bug for precisely the checks already cached.

## Relationship to #822

This is the same defect — a check asserting existence it never established — in the half #822's fix structurally could not reach. #822 stat-ed before reading, which works for a **file**. For a **URL** there is nothing to stat: the HTTP status is the only existence signal there is, and it was being discarded.

## Deliberately unchanged

The internal `/calendars` metadata fetch. It emits no step frame to any client, and its existing shape guards already treat a refused or malformed body as "not loaded yet", so there is no false `exists` there. A comment records that rather than leaving the omission to look accidental.

## Verification

Six new tests drive `onMessage()` through the reported scenario and its siblings — Ambrosian 1970 via `validateCalendar`, an unknown nation, a `resourceDataCheck` URL, and `runTest` — with no real network I/O.

Mutation-checked: replacing the 2xx gate with `return true` fails **4 of 6**; restored, all pass. The suite had no case where a URL check receives a 4xx, which is why this survived alongside #822's fix.

469 tests / 2648 assertions in the Health and schema suites, PHPStan level 10 clean, phpcs clean.

Refs #822, #821, UnitTestInterface#52

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now consistently reports HTTP status codes for calendar, resource, and test requests.
  * Failure responses include problem details and preserve `Retry-After` guidance when available.
  * Rate-limited and server-error responses are handled consistently without losing response information.
  * Cached responses retain validated status data, while malformed or out-of-range entries are discarded and refreshed automatically.
  * Unknown calendars now report clear 400 validation errors with valid-value guidance.

### DIFF
--- a/phpunit_tests/HealthHelpersTest.php
+++ b/phpunit_tests/HealthHelpersTest.php
@@ -14,13 +14,15 @@ use Psr\Http\Message\ResponseInterface;
 use React\Promise\Deferred;
 
 /**
- * Targeted coverage for two pure static helpers on the WebSocket Health component:
+ * Targeted coverage for pure static helpers on the WebSocket Health component:
  *  - isInternalApiUrl(): the host check that keeps the first-party WS_API_KEY from leaking to an
  *    arbitrary absolute URL validated via executeValidation.
- *  - isUpstreamFailureStatus(): which upstream HTTP statuses (429, 5xx) are rejected versus flowed
- *    through to per-format validation (e.g. a 404 for an unknown calendar).
+ *  - isUpstreamFailureStatus(): which upstream HTTP statuses (429, 5xx) are kept out of the cache
+ *    (#834) versus eligible for it (e.g. a 400 or 404, cached like any other response). All statuses
+ *    resolve and are reported through the same `exists` gate — this helper no longer decides who
+ *    rejects, only who gets cached.
  *
- * Both are dependency-free of the Ratchet/Guzzle machinery, so they are exercised directly via
+ * All are dependency-free of the Ratchet/Guzzle machinery, so they are exercised directly via
  * reflection without standing up the WebSocket server.
  */
 #[CoversClass(Health::class)]
@@ -135,7 +137,7 @@ final class HealthHelpersTest extends TestCase
      */
     private static function runHandleHttpResponse(ResponseInterface $response): array
     {
-        /** @var Deferred<array{data: string, status: int, fromCache: bool}> $deferred */
+        /** @var Deferred<array{data: string, status: int, fromCache: bool, retryAfter: ?string}> $deferred */
         $deferred = new Deferred();
         $resolved = null;
         $rejected = null;
@@ -159,31 +161,40 @@ final class HealthHelpersTest extends TestCase
         return ['resolved' => $resolved, 'rejected' => $rejected];
     }
 
-    public function testHandleHttpResponseRejectsRateLimited(): void
+    /**
+     * #834 revised this: a 429 used to be rejected on its own, separately from every other non-2xx.
+     * That carve-out existed because a 429's problem+json body "decodes fine but lacks the calendar
+     * keys," which the JSON-shape branch would otherwise mislabel as "response data was perhaps
+     * truncated?" — but every consumer now gates on {@see Health::isHttpSuccessStatus()} *before* any
+     * JSON-shape check runs, so that mislabelling can no longer happen regardless of how this method
+     * settles. A 429 now resolves like any other refusal, with its status on the resolved value and
+     * its `Retry-After` hint carried alongside it rather than only visible on a rejection's message.
+     */
+    public function testHandleHttpResponseResolvesRateLimitedWithStatusAndRetryAfter(): void
     {
         $response = new Response(429, ['Retry-After' => '30'], '{"status":429}', '1.1', 'Too Many Requests');
         $result   = self::runHandleHttpResponse($response);
 
-        $this->assertNull($result['resolved']);
-        $rejected = $result['rejected'];
-        if (!$rejected instanceof \RuntimeException) {
-            self::fail('Expected a RuntimeException rejection.');
-        }
-        $this->assertStringContainsString('HTTP 429', $rejected->getMessage());
-        $this->assertStringContainsString('Retry-After: 30', $rejected->getMessage());
+        $this->assertNull($result['rejected'], 'a 429 must resolve, not reject, like every other non-2xx');
+        $this->assertSame(
+            ['data' => '{"status":429}', 'status' => 429, 'fromCache' => false, 'retryAfter' => '30'],
+            $result['resolved']
+        );
     }
 
-    public function testHandleHttpResponseRejectsServerError(): void
+    /**
+     * A 5xx resolves the same way, with no `Retry-After` when the response carried none.
+     */
+    public function testHandleHttpResponseResolvesServerErrorWithStatus(): void
     {
         $response = new Response(503, [], 'service unavailable', '1.1', 'Service Unavailable');
         $result   = self::runHandleHttpResponse($response);
 
-        $this->assertNull($result['resolved']);
-        $rejected = $result['rejected'];
-        if (!$rejected instanceof \RuntimeException) {
-            self::fail('Expected a RuntimeException rejection.');
-        }
-        $this->assertStringContainsString('HTTP 503', $rejected->getMessage());
+        $this->assertNull($result['rejected']);
+        $this->assertSame(
+            ['data' => 'service unavailable', 'status' => 503, 'fromCache' => false, 'retryAfter' => null],
+            $result['resolved']
+        );
     }
 
     public function testHandleHttpResponseResolvesSuccessBody(): void
@@ -192,7 +203,7 @@ final class HealthHelpersTest extends TestCase
         $result = self::runHandleHttpResponse(new Response(200, [], $body));
 
         $this->assertNull($result['rejected']);
-        $this->assertSame(['data' => $body, 'status' => 200, 'fromCache' => false], $result['resolved']);
+        $this->assertSame(['data' => $body, 'status' => 200, 'fromCache' => false, 'retryAfter' => null], $result['resolved']);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


JohnRDOrazio:
Both code findings verified against the code and fixed (`775e1959`); the third is declined with measurements.

## Fixed — a malformed cache entry hard-failed the check instead of re-fetching

Correct, and this one corrected me as well as the code. I had described the legacy-entry handling as *"decode to `null` → reject and re-fetch"*. It rejected and **returned**:

```php
} else {
    $deferred->reject(new \RuntimeException("Cache fetch for URL $url failed or returned malformed cache payload"));
}
return $deferred->promise();   // never reaches the live-request path
```

So on deploy, every URL whose entry was written by the *old* code would fail hard until its TTL expired, and `handleValidationDataError()` would report all three steps failed for a check that is perfectly fine — a **false failure introduced by the fix for false successes**, landing exactly at upgrade time.

A malformed or undecodable entry is now treated as a miss: the stale key is deleted and execution falls through into the same live-request path a plain miss uses.

**Tested rather than reasoned about**, since that distinction is what produced the bug: the new test drives the real `cachedGet()` against a minimal APCu-shaped backend, seeds a **legacy bare-body entry**, and asserts there is no rejection, that a live request is actually queued, that the stale entry is deleted, and that fulfilling the queued request resolves the original promise.

## Fixed — 429/5xx now resolve uniformly, and the stale rationale is gone

On its own this was an inconsistency rather than a defect: those statuses rejected before reaching the new gate, but all steps still failed and the status still appeared in the text.

What decided it was the comment justifying the split — that a 429 body "decodes fine but lacks the calendar keys, which the JSON branch would otherwise mislabel as *response data was perhaps truncated?*". **This PR's gate makes that impossible**: every non-2xx is now caught before `json-valid` runs. The reasoning was obsolete, and stale reasoning is how the next reader concludes a split is load-bearing when it is not.

Every status now resolves with its status and body, and the gate reports them uniformly. Two parts of the original decision were deliberately preserved:

- **429/5xx are still excluded from the cache** — that half stood on its own merits.
- **`Retry-After` survives.** It is carried through as a new `retryAfter` field to all three gate call sites and folded into the failure text, so a 429 still tells the client when to come back.

The docblocks on `handleHttpResponse()`, `isUpstreamFailureStatus()` and `isHttpSuccessStatus()` now describe the current design rather than the superseded one.

## Declined — `#[Group('slow')]`

`HealthHttpStatusExistsTest` runs in **298 ms**. `CLAUDE.md` reserves that group for tests with measurable runtime cost — the suites that carry it are multi-year calendar computations and rate-limit window waits. Marking this would remove it from `composer test:quick`, which is the suite most likely to catch a regression in exactly this code.

## Verification

471 tests / 2661 assertions in the Health and schema suites (2 new, 0 new failures; the single error is the pre-existing environment-only `Routes/Readonly/HealthTest`). PHPStan level 10 and phpcs clean.

JohnRDOrazio:
Verified and fixed (`256daf19`).

`decodeHttpCacheEntry()` validated the status with `is_int()` alone, which accepts `0`, `-1` and `99999`. Any of those decodes as a *valid* envelope, reaches the `exists` gate as "not a 2xx", and fails a check that may be perfectly healthy.

That is the same false failure the malformed-entry fall-through was added to prevent one round ago — arriving through a **weaker type check rather than a missing one**, which is why it slipped past that guard. Range-validating (100–599) turns a corrupt envelope back into a cache miss, so the entry is deleted and the URL refetched like any other unusable one.

Mutation-checked: with the range check removed, the new test fails on the fall-through assertion; restored, it passes.

Worth recording: the first attempt at that mutation check restored the file with `git checkout -- src/Health.php`, which discarded the **uncommitted** fix along with the mutation. The test then failed for real, which is how I noticed — the assertion did exactly what it was written to do, on its author. The second attempt restored from a copy instead.

472 tests / 2667 assertions in the Health and schema suites; PHPStan level 10 and phpcs clean.

JohnRDOrazio:
Both fixed (`04d00f73`), and both were mine.

## The docblock still said 404

Correct. I corrected the assertion to 400 and left the prose above it saying *"An unknown calendar 404s"* and *"the 404 page's body"* — a comment stating the wrong status directly above a test asserting the right one. That is precisely the stale-reasoning problem this branch has spent its time removing, reintroduced while fixing an instance of it.

The docblock now leads with what actually happens and why: the API validates the nation against a known list and refuses the **value** with 400, rather than routing to a missing resource.

## `valid values are` was too generic

Also correct, and it mattered more than a nitpick. That phrase would match any parameter-validation message from any route, so the assertion proved only that *some* explanation came back — not that the one quoted belonged to this refusal. Since the point of #833 is that the client is told the real reason, an assertion that cannot tell one reason from another is not testing that.

Now pinned on `Invalid value` and `national_calendar`. The list of nations the message goes on to name is deliberately **not** asserted: that grows as calendars are added, and pinning it would make an unrelated addition fail here for no reason.

Mutation-checked — corrupting the endpoint substring fails the test.

## Verification

The WebSocket suite cannot run in a worktree against `:8082`, which is the shared main checkout's server — that is why the 404/400 guess shipped in the first place. So it was run against a WebSocket server started **from this branch** on a spare port (`WS_PORT` set in `.env.local`, since `Dotenv::createMutable` overrides the shell environment):

```text
OK (489 tests, 2892 assertions)   — Health, WebSocketMessage and WebSocket suites
```

phpcs clean.

Issue #839, opened 2026-08-21
The `locale` parameter is undocumented across the whole /data family


JohnRDOrazio:
`RegionalDataParams` accepts, canonicalises and validates a `locale` parameter across the whole `/data` family,
and `openapi.json` documents it on **none** of the paths.

## Evidence

`src/Params/RegionalDataParams.php:120-129`:

```php
if (array_key_exists('locale', $params)) {
    $locale = \Locale::canonicalize($params['locale']);
    if (null === $locale) {
        $description = "Invalid value {$params['locale']} for param `locale`";
        …
    }
    if (LitLocale::isValid($locale)) {
        $this->locale = $locale;
    } else {
        $description = "Invalid value {$params['locale']} for param `locale`, valid values are: la, la_VA, " . …
    }
}
```

It is live on GET, not merely declared. `GET /data/diocese/romamo_it?locale=zz_ZZ` returns **HTTP 400**:

```json
{
  "type": "https://datatracker.ietf.org/doc/html/rfc9110#name-400-bad-request",
  "title": "Bad Request",
  "status": 400,
  "detail": "Invalid value zz_ZZ for param `locale`, valid values are: la, la_VA, af, af_NA, af_ZA, …"
}
```

Every documented `/data` path declares exactly one parameter, `key`:

| path | documented parameters |
|---|---|
| `/data/nation/{key}` | `key` |
| `/data/diocese/{key}` | `key` |
| `/data/widerregion/{key}` | `key` |
| `/data/ambrosian/diocese/{key}` | `key` |

So a client cannot discover the parameter, cannot learn its accepted values, and — because the 400 is also
undocumented on these paths — cannot learn that supplying a bad one is a refusal rather than a fallback.

## Fix

Document `locale` on the `/data` family, `$ref`ing whatever shared locale definition the other paths already use
rather than inlining an enum (the `/calendar` and `/events` paths document a locale surface; follow that
precedent and check whether the accepted value set is genuinely the same before reusing its `$ref`).

Establish and document **where** the parameter is read from per method — query string on GET, request body on
PUT/PATCH, or both — rather than assuming it is uniform; the guard above reads it out of an already-assembled
`$params` array, so the binding differs by method upstream of it.

Declare the `400` response on the paths that can produce it. #818 found and fixed the same class of omission for
the rite-mismatch `422`, which was undeclared on every diocesan read.

[...]

Issue #838, opened 2026-08-21
openapi.json does not document the /data i18n sub-resource /data/{category}/{key}/{locale}


JohnRDOrazio:
`/data/{category}/{key}/{locale}` is a real, working sub-resource that returns a **different payload shape** from
the path it hangs off — and `openapi.json` does not document it at all.

## What exists

`RegionalDataHandler` binds a third path segment for GET and POST:

```php
if (in_array($method, [RequestMethod::GET, RequestMethod::POST], true)) {
    if (isset($this->requestPathParams[2])) {
        $params['i18n']        = $this->requestPathParams[2];
        $params['i18nRequest'] = $params['i18n'];
    }
}
```

— `src/Handlers/RegionalDataHandler.php:1855-1858`.

Confirmed live against `localhost:8000`, all HTTP 200:

| request | returns |
|---|---|
| `/data/diocese/romamo_it` | the diocesan calendar (`{"litcal":[…]}`) |
| `/data/diocese/romamo_it/it_IT` | a **translation map**: `{"BeatoGregorioX":"Beato Gregorio X, papa", …}` |
| `/data/nation/IT/it_IT` | translation map |
| `/data/ambrosian/diocese/milano_it/it_IT` | translation map |

So this is not a variant of the documented response — it is a distinct resource with an `event_key` → translated
name shape, available on all three tiers and on the rite-qualified path added in #818. A client reading the schema
cannot discover it, and cannot learn what it returns.

## Note the naming

The segment is bound internally to **`i18n`** / `i18nRequest`, not to `locale`. The `/data` family separately
accepts a `locale` *parameter* which is a different thing (see the companion issue). Documenting both without
distinguishing them would be worse than documenting neither, so whatever spelling the spec settles on should make
the distinction explicit.

## Fix

Document the sub-resource for the tiers and methods that actually accept it — GET and POST only, per the guard
above; establish and state what PUT/PATCH/DELETE do with a third segment rather than assuming they reject it.
A response schema for the translation-map shape is needed; check whether one already exists among the source-data
schemas (`LitCalTranslation.json`) before writing a new one.

Follow the precedent #818 set for the rite segment: mirror the existing `/data` path style, `$ref` the shared
`CommonDef.json` definitions rather than inlining, and add the rite-qualified variant alongside the bare one only
where it genuinely exists.

[...]

Issue #835, opened 2026-08-21
APCu is misconfigured in CI and its availability check would select a cache that silently stores nothing


JohnRDOrazio:
Two related APCu problems, filed together because the same fact settles both: **production uses Redis, and APCu is deliberately off under CLI.**

## Evidence

Measured on the deployment host, in the PHP CLI binary the WebSocket server runs under:

| | |
|---------------------------------------------|--------------------------------|
| `extension_loaded('apcu')`                  | `true`                         |
| `apc.enabled`                               | `1`                            |
| `apc.enable_cli`                            | **`0`**                        |
| `extension_loaded('redis')`                 | `true`                         |
| `Health` backend selection                  | tries **Redis first**; APCu only if Redis is absent |

So the live WebSocket server caches through Redis and never touches APCu.

---

## 1. The CI ini setting is a no-op (typo)

`.github/workflows/phpunit.yml` sets:

```yaml
ini-values: apcu.enable_cli=1, opcache.enable=1, opcache.enable_cli=1
```

**APCu's setting is `apc.enable_cli`, not `apcu.`** — it inherited the prefix from APC. So the line does nothing: the extension loads (making `function_exists('apcu_store')` true) while CLI support stays off, so `apcu_store()` silently no-ops and `apcu_exists()` returns false.

That combination is unusually good at hiding: any test guarding on `function_exists()` concludes APCu is present, then finds it does not work. It surfaced in #834, where a shim guarded with `!function_exists('apcu_store')` correctly stood down in CI and left the tests talking to a dead extension.

### Fix: correct the name, and set it to `0`

Not `1`. Enabling it would:

- exercise a caching path **production never takes**, since production selects Redis;
- introduce **process-wide state across 3116 tests** — APCu persists for the life of the PHP process, so one test caching a response changes what a later test sees, which is a classic source of order-dependent flakiness;
- buy no coverage that corresponds to any real configuration.

The cache *logic* is better covered deterministically by a shim (the approach #834 takes) than by a real backend whose state leaks between tests.

The current behaviour is therefore already correct — but correct **by accident**. Setting `apc.enable_cli=0` under the right name turns a typo into a recorded decision, with a comment noting that production uses Redis and APCu is deliberately off under CLI.

---

## 2. The availability check would select a cache that stores nothing

`src/Health.php` decides the backend with:

```php
$apcuAvailable = extension_loaded('apcu')
    && function_exists('apcu_exists')
[...]

PR #843, opened 2026-08-21
docs(openapi): document the rite segment on /data (#818)

### LINKED ISSUE #818: openapi.json understates the `/data` route — the rite segment is undocumented
`Router::extractRiteSegment()` accepts an optional leading rite segment on `data`, exactly as it does on
`calendar` and `events`:

```php
if ($route === 'calendar' || $route === '' || $route === 'events' || $route === 'data') {
```

`openapi.json` documents only `/data/nation/{key}`, `/data/diocese/{key}` and `/data/widerregion/{key}`, with
no rite-qualified variants — while `/calendar` and `/events` both document theirs. A client reading the schema
cannot discover `/data/ambrosian/diocese/{key}`.

Note the asymmetry is real and should be documented as such: only the diocesan tier exists under more than one
rite, so `/data/ambrosian/nation/{key}` correctly does not exist.

Found while implementing #48 (PR #49).

JohnRDOrazio:
Closes #818.

`Router::extractRiteSegment()` accepts a leading rite segment on `data` exactly as it does on `calendar` and `events`, but `openapi.json` documented no rite-qualified `/data` variants. A client reading the schema could not discover `/data/ambrosian/diocese/{key}`.

## The issue's framing was wrong in one important way

The issue calls the rite segment "optional". It is optional **only for Roman dioceses**, where it is an exact synonym. For the four Ambrosian dioceses it is **required** — the bare path is a 422, not a synonym. Documenting it as an optional decoration would have misdescribed the API, so the spec says "required here, not decorative" in bold.

Grammar established from source and confirmed live:

| request | result | proof |
|---|---|---|
| `/data/{cat}/{key}` | rite resolves to `roman` | `Router::extractRiteSegment()` returns `Rite::default()`; `'data'` is in the accepting list (`src/Router.php:133-144`) |
| `/data/roman/{cat}/{key}` | exact synonym, all three tiers | byte-identical responses (md5 match) vs the bare form |
| `/data/ambrosian/diocese/{key}` | 200 | `JsonData::diocesanCalendarFileFor($this->rite)` |
| `/data/ambrosian/diocese/{roman diocese}` | **422** | `RegionalDataHandler::checkDiocesanCalendarConditions()` |
| `/data/diocese/{ambrosian diocese}` | **422** | same check, both directions |
| `/data/ambrosian/nation/{key}` | **400** | `RegionalDataParams::validateRiteCompatibility()` |
| `/data/ambrosian/widerregion/{key}` | **400** | same |
| `/data/latin/…` | 400 "Unexpected path param latin" | unknown segment is not a rite; falls through as a category |

PUT is the one exception to the requirement — no stored rite to disagree with, so the comparison is skipped. The test's data provider excludes it and says why.

## What was added

- **`/data/ambrosian/diocese/{key}`**, full GET/POST/PUT/PATCH/DELETE, cloned from `/data/diocese/{key}` so tags, `security` blocks, parameter and response `$ref`s and operation ordering are identical. operationIds follow the `/data` family convention.
- **A Milan request-body example** (`DiocesanCalendarMilanRequestBody`) rather than reusing the Boston/Rome ones — those are Roman dioceses, and copying them onto an Ambrosian path would document a payload the path cannot legitimately take. Built from the real source file plus its `it_IT`/`la_VA` i18n entries; Redocly's `no-invalid-media-type-examples` validates it against `DiocesanCalendar.json`.
- **The `422` declared** on `/data/diocese/{key}` GET/POST/DELETE as well as the new path. This was a **pre-existing hole beyond the issue's scope**: the rite-mismatch refusal was undeclared on every diocesan read.
- **The asymmetry documented where a reader will find it** — the new path's descriptions state it in full, and `/data/nation/{key}` and `/data/widerregion/{key}` GET each gained a one-sentence note quoting the actual 400 text.

## No `/data/roman/…` duplicate paths — a judgement call worth reviewing

Two precedents exist in this file: `/calendar` enumerates both rites at every tier, `/events` documents only the Ambrosian variant. This follows `/events`, reasoning that `/calendar`'s roman paths exist to name the **canonical** form, and `Router::canonicalRiteUrl()` (`src/Router.php:199`) explicitly excludes `data` — `/data` emits no `Link: rel="canonical"` header, so there is no canonical form for such duplicates to name. The synonym is documented in prose on the bare paths instead.

(A templated `{rite}` parameter, as `/tests` uses, would over-promise here: the rite × tier matrix is not full.)

Happy to switch to the `/calendar` style if you'd prefer `/data` mirrored it.

## Tests

New `phpunit_tests/Schemas/OpenApiDataRiteSegmentTest.php` (15 tests, 37 assertions), pure-logic per the CLAUDE.md layering table. It asserts the documented set **against `validateRiteCompatibility()` itself** rather than a second hardcoded list — matching the house style of `OpenApiAmbrosianLocaleParamTest`, which asserts against `riteSupportsLocale()`. Negative control: renaming the path key produces 6 failures, so the assertions bite.

No existing test asserted the set of documented paths.

## Verification

- `composer lint:openapi`: *"Woohoo! Your API description is valid. 🎉"* — 11 problems explicitly ignored, **identical to the pre-change baseline**; no new suppressions in `.redocly.lint-ignore.yaml`
- `composer lint` (phpcs): clean
- `composer analyse` (PHPStan level 10): `[OK] No errors`
- `composer test:quick`: exit 0 — `Tests: 2933, Assertions: 177784, Skipped: 11`; delta is exactly the 15 new tests

## Follow-ups filed

Two further gaps on the same family, deliberately out of scope: #838 (the `/data/{category}/{key}/{locale}` i18n sub-resource is undocumented entirely) and #839 (the `locale` parameter is undocumented across the whole family).

## Summary by CodeRabbit

- **New Features**
  - Added documented Ambrosian diocesan data operations for retrieving, creating, updating, and deleting records.
  - Added validation and authorisation details for Ambrosian diocesan requests.
  - Added an example request for Milan’s Ambrosian diocesan data.

- **Bug Fixes**
  - Clarified Roman-rite resolution for national, regional, and diocesan data paths.
  - Documented `422` responses when diocesan requests use an incompatible rite.

### DIFF
--- a/jsondata/schemas/openapi.json
+++ b/jsondata/schemas/openapi.json
@@ -5471,7 +5471,7 @@
           {}
         ],
         "summary": "Retrieve a national calendar data resource",
-        "description": "There is no rite-qualified variant of this path other than the explicit synonym `/data/roman/nation/{key}`: only the diocesan tier exists under more than one rite, so `/data/ambrosian/nation/{key}` returns `400 Bad Request` (\"The ambrosian rite has no national calendars; request a diocesan calendar of that rite instead.\").",
+        "description": "`/data/roman/nation/{key}` states the same rite explicitly and is the canonical form of this path. There is no other rite-qualified variant: only the diocesan tier exists under more than one rite, so `/data/ambrosian/nation/{key}` returns `400 Bad Request` (\"The ambrosian rite has no national calendars; request a diocesan calendar of that rite instead.\").",
         "operationId": "nationalDataGET",
         "parameters": [
           {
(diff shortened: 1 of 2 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
One applied, one declined.

## Applied: compare method *sets*, not ordered path-item keys (`f1c60772`)

Valid, and it was asserting two things it did not mean to.

**Ordering** inside an OpenAPI path item carries no semantics, so a cosmetic reshuffle of operations would have failed the test with nothing actually wrong.

**Non-method keys** are the subtler half. A path item may legally hold `parameters`, `summary`, `description` or `servers` alongside its operations. None of the `/data` path items has one today — I checked all four — but hoisting the shared `key` parameter to the path level is an ordinary refactor, and it would have entered the security loop as though it were an operation, where `$operation['security'] ?? null` compares `null` to `null` and **passes for the wrong reason**. A test that silently stops checking is worse than one that fails.

Both sides are now filtered to HTTP methods and sorted before comparison, and the loop iterates only those methods.

Mutation-checked that the relaxed assertions still bite:

- dropping `DELETE` from the rite-qualified path → **2 failures**
- stripping `security` from its `PUT` → **1 failure**

## Declined: replace the POST description with a pointer to GET

The observation is accurate — `ambrosianDiocesanDataPOST`'s description is byte-identical to `ambrosianDiocesanDataGET`'s, 1536 characters of it — but the duplication is deliberate, and the pattern it is being compared against is the wrong one.

**On `/data`, `POST` is a read verb.** It is the "retrieve with parameters in a request body" form of `GET`, not a write. That is why the two share a description: they describe the same operation. The `PUT`/`PATCH`/`DELETE` descriptions on this path are short and point at `GET` because those *are* writes, with their own auth semantics to state.

The decisive evidence is the path this one was cloned from. On the bare `/data/diocese/{key}`:

| method | description |
|---|---|
| `GET` | 290 chars |
| `POST` | **290 chars — byte-identical to GET** |
| `PUT` | 308 chars, distinct |
| `PATCH` | 230 chars, distinct |
| `DELETE` | 196 chars, distinct |

So GET/POST already share one description on the bare path, and the rite-qualified path mirrors that exactly. Applying the suggestion would make the new path **inconsistent with the path it was cloned from**, which is the property this PR set out to preserve and which the accompanying test asserts.

That said, the drift risk you are pointing at is real and is worse here than on the bare path: 1536 duplicated characters drift more easily than 290. If it is worth addressing, the coherent fix is to collapse GET/POST to a pointer on **both** diocesan paths — and ideally on `/data/nation/{key}` and `/data/widerregion/{key}` too, which have the same GET/POST pairing — rather than on the new path alone. That is a separate change across the whole `/data` family; happy to open it as a follow-up if wanted.

## Also in this push

`development` merged in (`f6e0bea0`), so CI runs against the merge result. Verified before pushing:

- `composer lint` (phpcs) → clean (one alignment warning in my new code, auto-fixed with `phpcbf`)
- `composer analyse` (PHPStan level 10) → `[OK] No errors`
- `composer lint:openapi` → `Woohoo! Your API description is valid. 🎉`
- `composer test:quick` → `OK`, `Tests: 2973, Assertions: 177910, Skipped: 11`

JohnRDOrazio:
Applied (`4e3ba726`). Good catch, and it exposes a gap of exactly the kind this PR is otherwise about.

The exclusion was already *described* — this method's docblock says the Roman rite "is deliberately excluded: it is the rite an absent segment resolves to, and the contract expresses it through the bare path rather than through a `/data/roman/...` duplicate of every operation." But nothing **enforced** it. Prose does not fail. A duplicate could have been added without anything noticing that the contract had grown a second spelling of a path it already documents, and the stated reasoning for the exclusion — `Router::canonicalRiteUrl()` excludes `data`, so `/data` emits no canonical `Link` header and there is no canonical form for such a duplicate to name — would have been quietly contradicted by the very file it describes.

That is the same shape as the defect family this repo has been closing all week (#822, #833, #834, #835): a claim asserted where nothing establishes it.

The default-rite branch now asserts the absence alongside the bare form's presence. Deliberately framed as a tripwire rather than a prohibition — if enumerating both rites at every tier ever becomes the intent, mirroring how `/calendar` does it, this assertion is what has to be revisited. That is the point: the change becomes a conscious one rather than a drift.

Mutation-checked: adding a `/data/roman/diocese/{key}` duplicate fails with

```
Failed asserting that an array does not have the key '/data/roman/diocese/{key}'.
```

and `openapi.json` was restored byte-clean afterwards (`git diff` empty).

`OpenApiDataRiteSegmentTest` now `OK (15 tests, 40 assertions)`, up from 37. `composer lint` clean.

JohnRDOrazio:
Reversing the judgement this PR rested on, per the maintainer: enumerating both rites at every tier **is** the intent, mirroring `/calendar`, with the bare paths pointing at the explicit ones.

The reasoning I gave earlier was inverted. I argued that because `Router::canonicalRiteUrl()` excludes `data`, there is no canonical form for a `/data/roman/...` duplicate to name — treating that exclusion as a justification when it is itself the gap.

## What changed here (`c87d8436`)

**Added `/data/roman/{nation,diocese,widerregion}/{key}`**, cloned from their bare counterparts so tags, `security` blocks, parameter and response `$ref`s and the method surface stay identical. operationIds follow the family's existing convention — `romanNationalDataGET`, alongside `ambrosianDiocesanDataGET`. All three verified live: HTTP 200, payload byte-identical to the bare form.

**The bare descriptions** now name the explicit form as canonical rather than as "an accepted explicit synonym".

**The assertion from `4e3ba726` is inverted.** It asserted the *absence* of exactly these paths. It was written as a tripwire for the case where enumerating both rites became the intent — so the tripwire worked, just sooner than anyone expected. It now asserts presence, and is mutation-checked: renaming `/data/roman/nation/{key}` fails with `Failed asserting that an array has the key`.

### One note on how the edit was made

I spliced the new blocks in at text level rather than re-serialising the document. `openapi.json` is **inconsistently escaped** — some em-dashes as `—`, some literal — so no single `json.dumps` setting round-trips it. My first attempt re-serialised and produced a **24,151-line reformat**; this diff is 904 insertions and 43 deletions. Worth knowing before anyone else edits this file programmatically.

## Verification

- `composer lint:openapi` → `Woohoo! Your API description is valid. 🎉`, 11 problems ignored — identical to baseline, no new suppressions
- `composer lint` (phpcs) → clean
- `composer analyse` (PHPStan level 10) → `[OK] No errors`
- `composer test:quick` → `OK`, `Tests: 2973, Assertions: 177928, Skipped: 11`

## Deliberately **not** in this PR

**The runtime half**, filed as #848: `canonicalRiteUrl()` still excludes `data`, so `/data` emits no `Link: rel="canonical"`. The recorded rationale ("a `rel="canonical"` on a PUT is noise") is right but too broad — the resolution is to emit it on `/data`'s **read** methods, `GET` and `POST`, and not on writes. #848 also covers the two missing `/events/roman/...` paths (`/events` *emits* canonical links pointing at URLs its own contract does not document) and the fact that the `Link` header is declared on **zero** responses across all three families.

**Deprecating the bare paths.** `/calendar` marks its bare operations `deprecated: true` wherever a roman variant exists. Applying that rule to `/data` would deprecate its entire write surface, which is a louder contract statement than this PR should make unilaterally — generated clients surface it as warnings. Left for an explicit decision in #848.

Issue #849, opened 2026-08-21
ICS metadata stamps: DTSTAMP is local time wearing a Z, and the three stamps should not share one value


JohnRDOrazio:
`GET /calendar/{year}?return_type=ICS` returns a different `ETag` on every request, so an ICS client never
receives a `304` and re-downloads the full calendar (~372 KB) every time. JSON and XML revalidate correctly.

This is not the intended design — the ICS timestamps are *meant* to come from the GitHub release's
`published_at`, precisely so they are stable. Three separate defects defeat that.

Surfaced by `phpunit_tests/Routes/Readonly/CalendarEtagStabilityTest::testTheIcsRepresentationAlsoHasAStableEtag`,
which fails **deterministically** (3/3) against a local server on current `development`.

## Reproduction

```
$ curl -s "http://localhost:8000/calendar/2026?return_type=ICS" > a.ics
$ curl -s "http://localhost:8000/calendar/2026?return_type=ICS" > b.ics   # ~1s later
$ diff a.ics b.ics
13c13
< DTSTAMP:20260821T223658Z
---
> DTSTAMP:20260821T223659Z
15c15
< CREATED:20260821T203658Z
---
> CREATED:20260821T203659Z
```

Same byte length, different content, therefore different `ETag`. For contrast, two `GET /calendar/2026`
(JSON) requests return an identical `ETag: W/"4cd90754bcbc494cd344fa2e7830ac89"`.

---

## 1. `DTSTAMP` never uses the release date

`src/Handlers/CalendarHandler.php:5385-5403`:

```php
$publishDate = $GitHubReleasesObj->published_at;              // :5385  resolved…
…
$ical .= 'DTSTAMP:' . date('Ymd') . 'T' . date('His') . "Z\r\n";   // :5390  …and ignored
…
$ical .= 'CREATED:' . str_replace(…, $publishDate) . "\r\n";       // :5399  uses it
$ical .= 'LAST-MODIFIED:' . str_replace(…, $publishDate) . "\r\n"; // :5403  uses it
```

`CREATED` and `LAST-MODIFIED` consume the resolved publish date. `DTSTAMP` is hardcoded to the current time.

**This alone makes the ETag unstable, independently of everything below** — fixing the GitHub lookup would not
fix it. Any fix that stops at "make the release lookup reliable" leaves the bug in place.

## 2. `DTSTAMP` labels local time as UTC

[...]

PR #850, opened 2026-08-21
fix(cache): answer the APCu question with a round trip at the six surviving sites (#836)

### LINKED ISSUE #836: The APCu availability predicate #835 fixed in Health survives unfixed in six other places
The APCu availability predicate that #835 replaced in `Health` is not unique to `Health` — the identical
expression survives in **six** other places, and #835's fix does not transfer to any of them.

## The shape

#835 established that this predicate answers the wrong question:

```php
extension_loaded('apcu')
    && function_exists('apcu_exists')
    && function_exists('apcu_store')
    && function_exists('apcu_fetch')
```

All four operands are true on the deployment host while `apc.enable_cli=0`, so under the CLI SAPI it selects an
APCu that **silently stores nothing**: every store reports success, every read misses, and the only symptom is
quiet slowness. `Health` now answers it with a store→fetch round trip (`Health::apcuUsable()`), which also catches
a full or broken shared-memory segment.

## Where it survives

Enumerated against `b98f9951`:

| site                                        | enclosing                                 | form    |
|---------------------------------------------|-------------------------------------------|---------|
| `src/Utilities.php:573`                     | `jsonFileToArray()`                       | full    |
| `src/Utilities.php:607`                     | `jsonFileToObject()`                      | full    |
| `src/Utilities.php:642`                     | `invalidateJsonFileCache()`               | narrow  |
| `src/Utilities.php:665`                     | `jsonFileToObjectArray()`                 | full    |
| `src/Utilities.php:708`                     | `jsonUrlToObject()`                       | full    |
| `src/Models/MissalsPath/MissalMetadataMap.php:32` | `__construct()`                     | full    |

"Full" is the four-operand expression above; "narrow" is `!extension_loaded('apcu') || !function_exists('apcu_delete')`.

## Exposure is lower here than in `Health`, but the shape is still wrong

The callers of these helpers are all on the HTTP request path — `CalendarHandler`, `EventsHandler`,
`MissalsHandler`, `DecreesHandler`, `RegionalDataHandler`, `TemporaleHandler`, `CalendarMetadataProvider`,
`CalendarParams`, `LocaleConfigurator`, `AmbrosianSanctoraleLoader` — rather than in the WebSocket server's CLI
process, and under FPM `apc.enabled=1` makes them work. So this is **not** a live production defect the way
#835's `Health` fallback was.

It is still worth closing, on two grounds:

1. **The predicate reports a conclusion it never established.** That is the same family as #822, #833, #834 and
   #835, and the reason each of those survived so long is that the wrong answer looks exactly like the right one.
2. **The SAPI dependence is undocumented and unasserted.** Nothing records which entry points execute this code
   under which SAPI, and nothing fails if that changes. Any CLI-SAPI execution of these helpers gets the inert
   cache — silently.

**Part of the work is establishing that, and it should not be assumed either way:** enumerate which entry points
reach `Utilities::jsonFileTo*()` / `jsonUrlToObject()` under which SAPI (in particular whether the dev server
started by `composer start`, which runs under `cli-server`, is exposed — APCu's `apc.enable_cli` governs the `cli`
SAPI and its behaviour under `cli-server` was **not** verified while filing this).

## Fix

Give these sites the same round-trip verification `Health::apcuUsable()` uses, rather than six more copies of a
check that cannot fail honestly. A shared helper is the obvious shape, but note `Health` deliberately resolves
`apcu_*` **namespace-first** so `phpunit_tests/Support/ApcuShim.php` can stand in for the backend — any shared
helper has to keep that property or it becomes untestable.

Worth a test that an enabled-but-non-functional APCu is not selected at whichever sites end up sharing the helper,
mirroring `phpunit_tests/HealthApcuDetectionTest.php`.

## Related

- #835 is where the shape was diagnosed and fixed for `Health`'s two sites.
- #822, #833, #834 — the same "a check reporting success it never established" family.

JohnRDOrazio:
Closes #836.

The four-operand predicate `extension_loaded('apcu') && function_exists('apcu_exists'|'apcu_store'|'apcu_fetch')` answers the wrong question: all four are true on a host where `apc.enable_cli=0`, so under the CLI SAPI it selects an APCu that silently stores nothing. #835 fixed the two copies in `Health`; six survived.

## The issue's open question, answered — and the answer is reassuring

The issue explicitly recorded that `php -S` (`cli-server` SAPI) had **not** been checked. Measured in the project's own `litcal-api` image, which has real ext-apcu:

| SAPI | `apc.enable_cli` | `apcu_enabled()` | old predicate | round trip |
|---|---|---|---|---|
| `cli` | 0 | **false** | true | **fails** |
| `cli` | 1 | true | true | ok |
| `cli-server` (`php -S`) | 0 | **true** | true | **ok** |

APCu disables itself by comparing `sapi_module.name` against the literal `"cli"`; `php -S` reports `"cli-server"`, so `apc.enable_cli=0` never reaches it. **`composer start` gets a genuinely working cache.** All `src/` callers of these helpers are on the HTTP path (php-fpm or cli-server); the WebSocket server reaches this data only via internal HTTP requests, which execute in the HTTP process. That leaves **PHPUnit as the only routine `cli`-SAPI executor** — so these six were latent, but latent only as long as that list holds, and nothing asserted it.

## Placement: a new `src/ApcuCache.php`, not a helper function

A probe only describes the functions reachable *from the namespace its own calls are written in*. `MissalMetadataMap` lives in `…\Models\MissalsPath`, where the same unqualified `apcu_store()` resolves differently — it was deciding usability from one set of functions and writing through another. Rather than thread `__NAMESPACE__` through every call site and dispatch via variable function names, every `apcu_*` call outside `Health` now lives in one class in the namespace the shim already covers. **Decision-point and write-point are the same place by construction**, and `MissalMetadataMap` no longer calls `apcu_*` at all.

Other decisions:

- **Memoised.** These helpers are on the hot path of every handler (per JSON file load), unlike `Health`'s twice-per-connection probe, and `apc.enabled`/`apc.enable_cli` are startup settings. Tests reset the private static by reflection, the repo idiom (cf. `HealthApcuDetectionTest`); no public reset exists in the request path.
- **`delete()` is deliberately weaker** — gated on callability, not on the round trip. A store that cannot happen costs a cache hit; a delete that does not happen serves stale data.
- **`error_log()`, not `echo`.** `Health` is a CLI process; the same line from `Utilities` would be prepended to a PSR-7 body. Asserted by `testTheProbeNeverWritesToOutput`.

## The one surprise, and why `phpunit_tests/bootstrap.php` changed

Making the six sites shim-reachable turned the shim into a live, process-wide cache for the rest of the suite: **328 errors / 53 failures**. The shim stores objects *by reference* where real APCu serializes, so cached `stdClass` instances were mutated across tests.

The fix pins the answer in `bootstrap.php` before any test can load the shim — the same coupling CI's `apc.enable_cli=0` exists to prevent. The pinned value is what this process would genuinely answer with no shim loaded.

**It is deliberately not written as a call to `ApcuCache::isUsable()`**, however much shorter that would be. PHP's `INIT_NS_FCALL_BY_NAME` caches the resolved function in the opline's run-time cache, so executing `ApcuCache`'s unqualified `apcu_store()` while the extension is loaded and the shim is not **permanently binds that call site to the real function**. A shim required later is then ignored at exactly the sites it exists to stand in for. This is not hypothetical — it turned every positive case red under a loaded ext-apcu while staying green on a host without one. Documented in both files.

This is the change in this PR most worth a human eye.

## Tests

`phpunit_tests/ApcuCacheDetectionTest.php`, 14 cases. Pre-fix, against unmodified code in the CI-like container (`apcu` loaded, `apc.enable_cli=0`), the sharpest failure was a throwing backend propagating **straight out of a JSON read that had already succeeded**. Post-fix all 14 pass in three environments: dev host (no apcu), container with `apc.enable_cli=0`, and container with `apc.enable_cli=1` — where the previously-always-skipped `UtilitiesJsonFileCacheTest` actually runs, and stays green.

## Verification

- `composer test:quick`: baseline `Tests: 2973` → `Tests: 2987` (+14), no new skips

One pre-existing failure appears in full runs, `Routes/Readonly/CalendarEtagStabilityTest`. It is **not from this branch** — it fails 3/3 on pristine `development` too, and it is a `Routes/*` test hitting `:8000`, which serves the *main* checkout. Filed as #849.

## Follow-ups, deliberately not done here

- **`phpunit_tests/Support/ApcuShim.php` is not a faithful APCu**: it preserves object identity where real APCu serializes. Harmless while only `Health` (scalars/arrays) uses it, but it is now one call away from biting someone. Recommend `serialize()`/`unserialize()` in the shim.
- `Health::apcuUsable()` left untouched as instructed. It could eventually delegate to `ApcuCache`, but note it additionally requires `apcu_sma_info`, which `ApcuCache` does not.

## Summary by CodeRabbit

- **New Features**
  - Added reliable APCu cache detection with safe handling for unavailable or failing cache backends.
  - Centralised cache operations for JSON data and missal metadata.
  - Cache reads, writes, expiry and invalidation now work consistently across supported data sources.

- **Bug Fixes**
  - Prevented cache failures from interrupting normal data loading.
  - Ensured temporary detection data is cleaned up and cache invalidation reaches the correct backend.

- **Tests**
  - Added comprehensive coverage for cache usability, failures, cleanup, expiry and integration scenarios.

### DIFF
--- a/phpunit_tests/ApcuCacheDetectionTest.php
+++ b/phpunit_tests/ApcuCacheDetectionTest.php
@@ -66,6 +66,8 @@ final class ApcuCacheDetectionTest extends TestCase
         $this->usableBefore = is_bool($usable) ? $usable : null;
         self::forgetProbeResult();
 
+        self::assertShimIsTheBoundBackend();
+
         $this->tmpFile = sys_get_temp_dir() . '/litcal_apcu_cache_test_' . uniqid() . '.json';
     }
 
@@ -97,6 +99,48 @@ final class ApcuCacheDetectionTest extends TestCase
     private static function forgetProbeResult(): void
     {
         ( new \ReflectionProperty(ApcuCache::class, 'usable') )->setValue(null, null);
+    }
+
+    /**
+     * Fail loudly if `ApcuCache`'s calls are not reaching the shim.
+     *
+     * PHP's `INIT_NS_FCALL_BY_NAME` caches the function it resolved in the opline's run-time cache,
+     * so a call site that once executed while the shim was not loaded stays bound to the *global*
+     * `apcu_*` for the life of the process. Nothing in this class can undo that, and the failure is
+     * silent in the worst way: the fault switches below would simply have no effect, and the
+     * positive cases would pass against a real backend while the negative ones passed for the wrong
+     * reason.
+     *
+     * `phpunit_tests/bootstrap.php` is what keeps this from happening — by pinning the memo it stops
+     * `Utilities` reaching an `apcu_*` opline at all on hosts where the answer is `false`. But that
+     * argument does not hold on a host where the probe genuinely answers `true` (a real APCu with
+     * `apc.enable_cli=1`), where an earlier test *can* bind these sites first. Rather than load the
+     * shim from the bootstrap — which would also bind `Health`'s call sites to it and change what
+     * `Health::apcuUsable()` measures on such a host — the coupling is asserted here, where it is
+     * cheap and where a violation is a test failure rather than a wrong answer.
+     */
+    private static function assertShimIsTheBoundBackend(): void
+    {
+        ApcuShimStore::simulateDisabledStore(false);
+        ApcuShimStore::simulateThrowingStore(false);
+
+        $sentinel = self::PROBE_KEY_PREFIX . 'binding_' . uniqid();
+        ApcuShimStore::store($sentinel, 'bound', 10);
+
+        ( new \ReflectionProperty(ApcuCache::class, 'usable') )->setValue(null, true);
+
+        try {
+            $value = ApcuCache::fetch($sentinel, $found);
+            self::assertTrue(
+                $found && 'bound' === $value,
+                'ApcuCache is not reaching phpunit_tests/Support/ApcuShim.php: its apcu_* call sites were '
+                . 'bound to the global functions before the shim was loaded, so the fault switches this '
+                . 'class relies on cannot take effect. See assertShimIsTheBoundBackend().'
+            );
+        } finally {
+            ApcuShimStore::delete($sentinel);
+            self::forgetProbeResult();
+        }
     }
 
     /**
--- a/phpunit_tests/ApcuCacheDetectionTest.php
+++ b/phpunit_tests/ApcuCacheDetectionTest.php
@@ -1,0 +1,372 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests;
+
+use LiturgicalCalendar\Api\ApcuCache;
+use LiturgicalCalendar\Api\ApcuShimStore;
+use LiturgicalCalendar\Api\Models\MissalsPath\MissalMetadataMap;
+use LiturgicalCalendar\Api\Router;
+use LiturgicalCalendar\Api\Utilities;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * #836: an APCu that is present but inert must not be selected as the JSON-file cache either.
+ *
+ * The same predicate #835 replaced in `Health` — `extension_loaded('apcu')` plus `function_exists()`
+ * on three of the `apcu_*` names — survived in six further places: the four `Utilities::jsonFileTo*()`
+ * / `jsonUrlToObject()` readers, `Utilities::invalidateJsonFileCache()` (a narrower variant), and
+ * `MissalMetadataMap::__construct()`. All of them now consult {@see ApcuCache}, whose answer comes
+ * from a store→fetch round trip.
+ *
+ * Both directions are asserted, as in {@see HealthApcuDetectionTest}: a probe that always answered
+ * "no" would satisfy every negative case here while silently disabling caching outright, so the
+ * positive cases are what make the negative ones mean something.
+ *
+ * **Why the sites are tested and not only the probe.** `ApcuCache` exists in the
+ * `LiturgicalCalendar\Api` namespace on purpose. PHP resolves an unqualified `apcu_store()` against
+ * the *calling* namespace before the global one, so a probe run inside `LiturgicalCalendar\Api` only
+ * describes the functions a caller in `LiturgicalCalendar\Api` will really reach.
+ * `MissalMetadataMap` lives in `LiturgicalCalendar\Api\Models\MissalsPath`, where the very same
+ * unqualified call resolves somewhere else entirely — which is why it no longer calls `apcu_*` at all
+ * and goes through `ApcuCache` instead. `testMissalMetadataMapCachesThroughTheSharedHelper()` is what
+ * holds that property down: before #836 it failed, because the map's own unqualified calls bypassed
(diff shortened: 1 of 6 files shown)


JohnRDOrazio:
All three applied, one with a narrower remedy than proposed (`4e3ac348`).

## Applied: contain backend exceptions in `exists()` / `fetch()` / `store()`

Valid, and the strongest argument for it is the class's own docblock: *"an unusable backend is reported as a miss rather than as an error"*. That promise held for an **inert** backend but not for a **throwing** one, which escaped straight out of a JSON read — the harder of the two failures to reason about was the one left uncovered. `delete()` and the probe were already guarded, so the class was total in three places out of five.

`fetch()` sets `$success = false` and returns `false`; `exists()` and `store()` return `false`; each logs through `error_log()` (not `echo` — these run under FPM, where output would be prepended to a PSR-7 body).

## Applied: drop the redundant `exists()` wrappers

All three call sites — `jsonFileToArray()`, `jsonFileToObject()`, `MissalMetadataMap::buildIndex()`. `fetch()` already reports both a miss and an unusable backend through `$success`, so `exists()`-then-`fetch()` bought only a second round trip on a hot path, plus a window for the entry to expire between the two calls. The success and data-shape guards are untouched.

## Applied differently: the shim-binding concern

The request was to load `ApcuShim.php` before any `ApcuCache` call site, **including from the bootstrap**. I have not done that, because loading the shim from `bootstrap.php` would also bind **`Health`'s** `apcu_*` call sites to it, changing what `Health::apcuUsable()` measures on a host with a real CLI-enabled APCu. That is a wider blast radius than the problem being solved, and `Health` belongs to a different branch (#851).

**The concern behind it is real, though, and was not covered.** `INIT_NS_FCALL_BY_NAME` caches its resolution per opline, so a call site that once executed while the shim was not loaded stays bound to the global `apcu_*` for the life of the process. `bootstrap.php` prevents that today by pinning the memo — which stops `Utilities` reaching an `apcu_*` opline at all — but that argument only holds where the probe answers `false`. On a host where it genuinely answers `true` (real APCu, `apc.enable_cli=1`), an earlier test *can* bind these sites first, and the failure is silent in the worst way: the fault switches simply stop working, so the positive cases pass against a real backend and the negative ones pass for the wrong reason.

So `setUp()` now writes a sentinel directly through the shim and requires `ApcuCache` to read it back, failing with an explanatory message if it does not:

```
ApcuCache is not reaching phpunit_tests/Support/ApcuShim.php: its apcu_* call sites were
bound to the global functions before the shim was loaded, so the fault switches this class
relies on cannot take effect. See assertShimIsTheBoundBackend().
```

That converts a silent wrong-answer into a loud failure, at the layer where the coupling actually matters, without changing what the bootstrap binds. Mutation-checked: changing the sentinel's value makes every case fail with exactly that message.

## Verification

- `ApcuCacheDetectionTest` → `OK (14 tests, 46 assertions)` (was 32 — the guard adds one per case)
- `composer test:quick` → exit 0, `Tests: 2987, Assertions: 177974, Skipped: 11`, no failures

JohnRDOrazio:
@coderabbitai review

PR #851, opened 2026-08-21
fix(health): terminate a superseded queued request instead of discarding it (#837)

### LINKED ISSUE #837: dropSupersededQueuedRequests() discards queued requests without rejecting them, so a superseded request never terminates
`Health::dropSupersededQueuedRequests()` filters superseded entries out of the request queue **without calling
their `reject`**, so a dropped request emits neither its remaining step frames nor a terminal `complete` frame.

This is the second of the two known holes in the "every path that starts work terminates" guarantee introduced by
#806 section C. The first — a throw inside a fulfil handler — was closed by #823. This one was recorded alongside
it and deliberately left open, so that the two would not be conflated. It is now the only one left.

## The code

`src/Health.php`:

```php
private function dropSupersededQueuedRequests(): void
{
    if (empty($this->queue)) {
        return;
    }
    $this->queue = array_values(array_filter(
        $this->queue,
        fn (array $item): bool => null === $item['resourceId']
            || null === $item['runToken']
            || ( $this->runTokens[$item['resourceId']] ?? null ) === $item['runToken']
    ));
}
```

Every queue entry carries the deferred's callbacks:

```php
$this->queue[] = [
    'url'        => $url,
    'options'    => $options,
    'resolve'    => $resolve,
    'reject'     => $reject,
    'resourceId' => $queuedResourceId,
    'runToken'   => $queuedRunToken
];
```

`array_filter` drops the entry and with it the only reference to `reject`. The promise handed to the caller is
never settled — not resolved, not rejected — so neither the fulfil handler nor the sibling `onRejected` ever runs,
and the request goes silent after whatever step frames it had already managed to emit.

## How it is reached

Two call sites, and the second is the less obvious one:

| call site | trigger |
|---|---|
| `cancelRun()` | an explicit client `cancelRun` — `unset($this->runTokens[$resourceId])` immediately precedes the drop, so every queued entry for that connection is superseded |
| `processQueue()` | called on **every** queue pass, so an ordinary run-token change supersedes in-flight queued work without any explicit cancellation |

## Why it matters now

Before #806 section C a client counted `checks * 3` frames and wedged here for the same reason, so this is
pre-existing rather than a regression. But the terminal frame makes the silence nameable: a v2 client that stops
on `complete`, exactly as the design tells it to, waits forever for a dropped request.

Note that #823's fix does **not** cover this. That fix attaches a recovery arm to the promise derived from each
`then()`, which terminates the request when a handler *throws*. Here no handler ever runs at all, because the
promise is never settled — there is nothing for a rejection handler to catch.

## Fix

Reject the dropped entries rather than discarding them, so the existing `onRejected` arms run and the request
terminates the way any other failure does. The distinction a fix has to preserve: a **cancelled** run and a
**failed** one should not look identical to a client — `cancelRun` is a request to stop, and terminating with an
error frame that reads like a check failure would be its own untruth. Decide deliberately whether a cancelled
request emits `complete` only, or a distinguishable cancellation signal followed by `complete`.

Worth a test per call site — one driving an explicit `cancelRun` with work queued, one driving a run-token change
— asserting the queued request terminates rather than going silent. `phpunit_tests/HealthFulfilHandlerThrowTest.php`
(from #823) is the closest existing pattern for asserting "the request terminated exactly once".

## Related

- #806 section C introduced the guarantee this qualifies.
- #823 closed the other known hole in it and is where this one was re-confirmed as still open.
- Recorded in `docs/superpowers/specs/2026-08-20-structured-responses-design.md`, section C, at the end of the
  known-limit subsection.

JohnRDOrazio:
Closes #837.

`dropSupersededQueuedRequests()` filtered superseded entries out of the queue with `array_filter`, discarding each entry's callbacks without settling the promise. Neither the fulfil handler nor the sibling `onRejected` ever ran, so the request went silent after whatever step frames it had already emitted, and a v2 client stopping on `complete` waited forever.

This was the last of the two known holes in the "every path that starts work terminates" guarantee; #823 closed the other.

## Both call sites

| call site | line | trigger |
|---|---|---|
| `cancelRun()` | `Health.php:1172` (right after `unset($this->runTokens[$resourceId])`) | explicit client cancellation |
| `processQueue()` | `Health.php:4125` (first statement) | **every** queue pass — an ordinary run-token change supersedes queued work with no cancellation |

Droppable entries come from exactly three client-facing `cachedGet()` callers, each 1:1 with one request and one `sendComplete()`. The fourth caller — the connect-time `/calendars` metadata fetch — is queued before any run token exists, is therefore never dropped, and needs no terminator.

## The issue's literal fix is a trap

#837 said "call their `reject`". Doing so would corrupt the concurrency accounting: `++$this->inFlight` happens at **dispatch** (`Health.php:4162`, inside `processQueue()`'s loop), while `cachedGet()`'s `reject` closure decrements it. A **queued** entry never incremented the counter, so rejecting it drives `inFlight` negative and silently raises the concurrency cap for the life of the process.

This already has a footprint in the existing suite: `HealthTerminalFrameTest` settles queued entries by hand and its drain log prints `inFlight: -1` and `inFlight: -2`. Hence a separate `onSuperseded` terminator on the entry rather than reuse of `reject`.

## Cancellation must not be reported as failure

The second reason not to reject into the existing arms is honesty. Those arms report a *failed check*: `validateCalendar()`'s says the calendar "does not exist at the URL", `executeUnitTest()`'s says it "was not retrieved", the resource arm runs `handleValidationDataError()` and emits three FAIL frames. **None of that happened** — the fetch never left the queue. Reporting it would be the same "reports something that did not happen" family as #822/#833/#834/#835.

**A superseded request now emits its terminal frame and nothing else: `complete` carrying `cancelled: true`.**

- Distinguishability rides on a frame the client was already going to receive, never a new one — a v1 client sizes a phase as `checks * 3`, so an extra frame is a cascading miscount.
- `type` stays `'success'`: since UnitTestInterface#46 an unrecognised `type` is painted as a visible failed check, which is exactly what an abandoned run must not look like.
- `cancelled` is **omitted** rather than sent as `false`, so an ordinary terminal frame stays byte-identical.
- The `requestId` gate is untouched, so a v1 client still sees nothing at all.

**Completes exactly once**: the promise is deliberately left unsettled, so neither `then()` handler nor #823's tail handler can run; a queued entry has by definition not been dispatched, so no frame of that request has been emitted yet; and `$this->queue` is reassigned to the survivors *before* any terminator runs, so a re-entrant pass cannot see a dropped entry twice. Asserted directly — the tests count terminal frames as exactly 1.

## Tests

`phpunit_tests/HealthSupersededRequestTest.php`, 6 cases: both call sites, the test-run site, "no step outcome", the v1 gate, and a throwing terminator not taking the batch down. Pre-fix, 3 failed with exactly the reported silence (`the stream was: []`), plus one risky test with no frames to assert on. Post-fix `OK (6 tests, 29 assertions)`; full `--filter Health` `OK (463 tests, 2796 assertions)`.

## Verification

- `composer lint`, `composer lint:md` clean; `composer analyse` (PHPStan level 10) `[OK] No errors`
- `composer test:quick`: `Tests: 2979, Assertions: 177957, Skipped: 11`, exit 0

## Protocol note

`cancelled` is a **new key on the terminal frame** — a v2 addition behind the existing `requestId` gate. UnitTestInterface#42 will want to know about it when the v2 contract lands.

Section C of `docs/superpowers/specs/2026-08-20-structured-responses-design.md` now records the guarantee as having both holes closed, with an amendment covering the fix, the rejected alternative, the `inFlight` trap and the `cancelled` key.

## Summary by CodeRabbit

* **New Features**
  * Cancelled or superseded queued requests now receive a clear terminal completion frame.
  * Cancelled requests no longer produce misleading step or error frames.
  * HTTP response statuses and retry guidance are handled consistently, including cached responses.

* **Bug Fixes**
  * Improved validation failure reporting for refused responses, malformed data and decoding errors.
  * Fixed dropped queued requests that previously ended without a completion notification.
  * Improved calendar, diocesan path and schema matching.
  * APCu availability checks now verify stored and retrieved values.

### DIFF
--- a/docs/superpowers/specs/2026-08-20-structured-responses-design.md
+++ b/docs/superpowers/specs/2026-08-20-structured-responses-design.md
@@ -358,6 +358,7 @@ happy path        exists(pass) → parses(pass) → validates(pass) → complete
 JSON decode fails exists(pass) → parses(fail) → validates(fail) → complete
 file missing      exists(fail) → parses(fail) → validates(fail) → complete
 test run          validates(pass|fail) → complete
+cancelled run     complete(cancelled) — the queued work was dropped before it ran, so no step frame precedes it
 unknown target    echobot rejection only — no complete, nothing was started
 ```
 
@@ -370,6 +371,14 @@ A client stops on `complete` and never counts frames.
 > `steps` is now an exact frame count and safe to render one card per step against. The stopping rule is unchanged
 > and the terminal frame is not thereby redundant: `complete` stays correct for an arm a later change makes shorter,
 > which is the property that made it worth adding while the counts disagreed.
+>
+> **Amended after #837 landed: the `cancelled run` row, and the one key that goes with it.** A request whose queued
+> work is dropped because its run was cancelled or superseded terminates like any other — but it emits **no step
+> frames at all**, because none of its steps ran, and its terminal frame carries `"cancelled": true` so a client can
+> tell a run it abandoned from one that failed. That key is the only addition to this frame, `type` stays `'success'`,
+> and it is omitted rather than sent as `false` on every other path, so an ordinary terminal frame is unchanged. The
+> full reasoning — including why a dropped request is *not* rejected into the existing failure arms — is in the
+> known-holes subsection below.
 
 ### A closed hole, and the shape that would reopen it: a throw inside a fulfil handler (#823)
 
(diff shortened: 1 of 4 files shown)


JohnRDOrazio:
@coderabbitai review

Issue #817, opened 2026-08-20
Ambrosian diocesan `/events` responses fail the API's own `LitCalEventsPath` schema


JohnRDOrazio:
`jsondata/schemas/LitCalEventsPath.json`'s `settings` property is a `oneOf` with exactly three branches:

- `{national_calendar: string, diocesan_calendar: <diocesan id>}`
- `{national_calendar: string, diocesan_calendar: null}`
- `{national_calendar: null,   diocesan_calendar: null}`

None of the three permits `{national_calendar: null, diocesan_calendar: <diocesan id set>}` — which is exactly
the shape returned by `GET /events/ambrosian/diocese/milano_it?locale=it`:

```json
{"locale":"it_IT","national_calendar":null,"diocesan_calendar":"milano_it","rite":"ambrosian"}
```

Only an Ambrosian diocese produces this shape, because the Ambrosian rite has no national tier — a diocesan
request under Ambrosian has a diocesan calendar but no national calendar to report. The response is correct;
the schema is incomplete.

This surfaces in the UnitTestInterface `resources.php` runner as 8 red `events-path-diocese-*` `schema-valid`
cards on every Ambrosian resources run (2 locales × 4 Ambrosian dioceses), confirmed live: all 8 failures come
back HTTP 200 with the shape above, so the request itself succeeds and only the schema validation fails.

This is a direct instance of what LiturgicalCalendarAPI#800 described. Fix: add a fourth `oneOf` branch (or
otherwise relax the schema) for `{national_calendar: null, diocesan_calendar: <diocesan id>}`.

Found while implementing #48 (PR #49).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #852, opened 2026-08-21
fix(schema): admit Ambrosian diocesan /calendar settings in LitCal.json (#844)

### LINKED ISSUE #844: Ambrosian diocesan /calendar responses fail LitCal.json — national_calendar is omitted, not null
`GET /calendar/ambrosian/diocese/milano_it/2026` returns HTTP 200 with a `settings` object that fails
`LitCal.json`'s own `settings` schema. This is the `/calendar` counterpart of #817 — but it is **not the same
shape**, so #817's fix does not transfer.

## The gap

`LitCal.json`'s `settings` is a `oneOf` over three named definitions:

| definition | requires |
|---|---|
| `NationalCalendarSettings` | … `national_calendar`, `rite` |
| `DiocesanCalendarSettings` | … `national_calendar`, `diocesan_calendar`, `rite` |
| `GeneralRomanCalendarSettings` | … `rite` (no calendar keys at all) |

All three set `additionalProperties: false`, and `DiocesanCalendarSettings.national_calendar` is
`$ref: CommonDef.json#/definitions/Nation` — a string enum.

The Ambrosian diocesan response omits `national_calendar` **entirely**:

```
settings keys: ascension, corpus_christi, diocesan_calendar, epiphany, eternal_high_priest,
               holydays_of_obligation, locale, return_type, rite, year, year_type
```

Validated against `LitCal.json#/properties/settings` it fails all three branches:

```
No valid results for oneOf {
  0: Required property missing: national_calendar
  1: Required property missing: national_calendar
  2: Additional properties not allowed: diocesan_calendar
}
```

The rite-level `/calendar/ambrosian/2026` response **is** valid — it carries no calendar keys, so it matches
`GeneralRomanCalendarSettings`. Only the diocesan tier is affected, for the same underlying reason as #817: the
Ambrosian rite has no national tier, so an Ambrosian diocese has no national calendar to report.

## Why #817's fix does not transfer

#817 fixed the `/events` spelling, where `national_calendar` is present and **`null`**. Here it is **absent**.
Two different fixes:

- `/events` needed a branch admitting `national_calendar: null`.
- `/calendar` needs a branch that does not `require` `national_calendar` at all — and, because
  `additionalProperties: false` is set, one that explicitly declares `diocesan_calendar` so it is not rejected as
  an unexpected property.

**Whether the two endpoints *should* spell it differently is worth deciding as part of this**, rather than
schema-patching around a divergence nobody chose. Making `/calendar` emit `national_calendar: null` like
`/events` would be a response-shape change with client impact; keeping them different means the schemas stay
asymmetric forever. Either is defensible; drifting into one by default is not.

## Fix

Apply the pattern #817 established: add a branch discriminated on `rite`, reusing
`CommonDef.json#/definitions/RiteWithoutNationalCalendars` (added by #817 for exactly this purpose, and
documented there as "keep in sync with `Rite`: whenever a rite is added, decide explicitly whether it has a
national tier").

**Mind the draft-07 trap #817 measured:** `{"$ref": …, "not": {…}}` and `{"$ref": …, "enum": […]}` are both
**silent no-ops** — `$ref` replaces the schema object and sibling keywords are ignored. Use a `$ref` to a narrow
definition, or `allOf`. `phpunit_tests/Schemas/EventsPathSettingsSchemaTest.php` has a committed guard asserting
no `$ref` in those branches carries sibling validation keywords; a `/calendar` equivalent is worth adding.

Also re-run the `oneOf` exclusivity analysis after the change — a payload matching two branches fails `oneOf`
just as hard as one matching none.

## Related

- #817 — the `/events` counterpart, and the source of `RiteWithoutNationalCalendars`.
- #800 — the general observation this is another instance of.
- A separate looseness worth folding in if it is addressed: `NationalCalendarSettings` and
  `DiocesanCalendarSettings` do not constrain `rite`, so they still admit an Ambrosian national calendar, which
  the routing layer refuses with a 400. #817 deliberately left the same looseness in `LitCalEventsPath.json`
  rather than fix half of it; both files would be fixed together.

Found while implementing #817.

JohnRDOrazio:
Closes #844.

`GET /calendar/ambrosian/diocese/milano_it/2026` returns HTTP 200 with a `settings` object that fails `LitCal.json`'s own `settings` schema. It is the `/calendar` counterpart of #817 — but **not the same shape**, so that fix does not transfer.

## Why a new definition rather than a widened one

`/events` presents `national_calendar` as present-and-`null`; `/calendar` **omits the key entirely**:

```
settings keys: ascension, corpus_christi, diocesan_calendar, epiphany, eternal_high_priest,
               holydays_of_obligation, locale, return_type, rite, year, year_type
```

You cannot express "this key must not appear" by loosening a type. And because `additionalProperties: false` is set, you also cannot simply drop `national_calendar` from `DiocesanCalendarSettings`' `required` — that would legalise the omission for **Roman** dioceses too. A separate definition, discriminated on `rite`, is the only shape that admits Ambrosian while keeping Roman closed.

Added `#/definitions/DiocesanCalendarSettingsWithoutNationalCalendar` — `DiocesanCalendarSettings` verbatim with two deltas: `national_calendar` is not declared at all and is dropped from `required` (so `additionalProperties: false` does the rejecting), and `rite` is a bare `$ref` to `CommonDef.json#/definitions/RiteWithoutNationalCalendars`, the definition #817 added for exactly this purpose.

**Never a `$ref` plus a sibling `enum`/`not`** — in draft-07 a `$ref` replaces its containing schema object and siblings are silently ignored, so those spellings validate everything they appear to forbid. The draft-07 guard from `EventsPathSettingsSchemaTest` is extended here; because `LitCal.json` spells each branch as a `$ref` to a local definition, the guard first asserts the branch object carries nothing but `$ref`, resolves it, then inspects the per-property sub-schemas.

## `oneOf` exclusivity, measured

Disjointness is structural: `National`/`Diocesan` both *require* `national_calendar`, which the new branch forbids; `GeneralRoman` forbids `diocesan_calendar`, which the new branch requires.

| payload | valid | branches matched |
|---|---|---|
| General Roman, Roman national, Roman diocesan | yes | 1 each |
| `/calendar/ambrosian/{year}` (rite-level) | yes | 1 |
| Ambrosian diocesan × 4 dioceses × 2 locales | yes | 1 each |
| **synthetic** Roman diocesan, `national_calendar` **omitted** | **no** | **0** |
| **synthetic** Roman diocesan, `national_calendar: null` | no | 0 |
| **synthetic** General Roman + `diocesan_calendar` | no | 0 |

Pre-fix the 8 Ambrosian rows were `valid=no, branches=0`. The three Roman rows matched exactly one branch before and after — the change is purely additive to them.

## Tests

`phpunit_tests/Schemas/CalendarSettingsSchemaTest.php` — 49 tests / 205 assertions, following `EventsPathSettingsSchemaTest`'s `#[DataProvider]` + `Router::getApiPaths()` + `Schema::import()` pattern: 12 emitted shapes asserted valid *and* exactly-one-branch, 11 invalid shapes asserted rejected *and* zero-branch, the dedicated Roman-paradigm test, the draft-07 guard, and a drift guard tying the new definition to `DiocesanCalendarSettings`.

Test-first: with the schema at `HEAD`, `Tests: 49, Errors: 8, Failures: 9`.

## Verification

- `composer lint:openapi` — valid, 11 problems ignored, unchanged (openapi.json `$ref`s `./LitCal.json`)
- `composer test:quick`: baseline `Tests: 2973` → `Tests: 3022` (+49), green

## A decision surfaced, not taken

**The response shape was deliberately not changed.** Should `/calendar` emit `national_calendar: null` like `/events` instead of omitting it?

The divergence looks like an accident of two independent code paths — `/events` builds a fixed-key DTO and leaves the field null, `/calendar` builds incrementally and never sets it — rather than a design. `null` is arguably the better spelling: it states the field was considered and is inapplicable, where omission is indistinguishable from a serialisation bug, and it keeps the key set stable across tiers for typed clients.

Against that, it is a breaking-ish response change: a client doing `'national_calendar' in settings` flips from false to true. So it belongs in its own change with its own note. The schema must describe what the API emits *today* regardless; if the shape later converges, this definition becomes a one-line edit and the rite discriminator survives intact.

## Known looseness left alone, deliberately

`NationalCalendarSettings` and `DiocesanCalendarSettings` do not constrain `rite`, so a synthetic *Ambrosian national* payload still validates (the routing layer refuses it with 400, so it is unreachable). Not fixed here because #817 left the identical hole in `LitCalEventsPath.json`, and fixing one file would recreate exactly the asymmetry #844 objects to. The clean fix is a `RiteWithNationalCalendars` complement in `CommonDef.json` referenced from both files.

## Summary by CodeRabbit

* **New Features**
  * Added support for diocesan calendars that do not have a national calendar.
  * Added validation for the required diocesan settings and supported rites.

* **Bug Fixes**
  * Improved settings validation to reject unsupported, missing, null, or unexpected calendar properties.

* **Tests**
  * Added comprehensive coverage for Roman, national, diocesan, and Ambrosian calendar settings, including valid and invalid configurations.

### DIFF
--- a/jsondata/schemas/LitCal.json
+++ b/jsondata/schemas/LitCal.json
@@ -13,6 +13,9 @@
                     "$ref": "#/definitions/DiocesanCalendarSettings"
                 },
                 {
+                    "$ref": "#/definitions/DiocesanCalendarSettingsWithoutNationalCalendar"
+                },
+                {
                     "$ref": "#/definitions/GeneralRomanCalendarSettings"
                 }
             ]
@@ -1044,6 +1047,70 @@
                 "holydays_of_obligation",
                 "rite",
                 "national_calendar",
+                "diocesan_calendar"
+            ]
+        },
+        "DiocesanCalendarSettingsWithoutNationalCalendar": {
+            "type": "object",
+            "title": "Diocesan Calendar Settings without a National Calendar",
+            "description": "the settings that were utilized (and possibly requested) to generate a diocesan calendar belonging to a rite that has no national tier. Identical to `DiocesanCalendarSettings` except that `national_calendar` is not merely nullable but absent: `CalendarHandler` never sets one for such a diocese, and `additionalProperties: false` combined with the property not being declared here keeps it that way. `rite` is narrowed to `RiteWithoutNationalCalendars` so that a *Roman* diocesan response missing its `national_calendar` — an impossible shape, since under the Roman rite a diocese always inherits from a national calendar — keeps being rejected by this branch and by every other one. See issue #844 (and #817 for the `/events` counterpart, which spells the same situation as an explicit `national_calendar: null`).",
+            "additionalProperties": false,
+            "properties": {
+                "year": {
+                    "$ref": "./CommonDef.json#/definitions/Year"
+                },
+                "year_type": {
+                    "type": "string",
+                    "enum": [
+                        "CIVIL",
+                        "LITURGICAL"
+                    ]
+                },
+                "epiphany": {
+                    "$ref": "./CommonDef.json#/definitions/Epiphany"
+                },
+                "ascension": {
+                    "$ref": "./CommonDef.json#/definitions/Ascension"
+                },
+                "corpus_christi": {
+                    "$ref": "./CommonDef.json#/definitions/CorpusChristi"
+                },
+                "locale": {
+                    "$ref": "./CommonDef.json#/definitions/Locale"
+                },
+                "return_type": {
+                    "type": "string",
+                    "enum": [
+                        "ICS",
+                        "JSON",
+                        "XML",
+                        "YML"
+                    ]
+                },
+                "eternal_high_priest": {
+                    "type": "boolean"
+                },
+                "holydays_of_obligation": {
+                    "$ref": "./CommonDef.json#/definitions/HolyDaysOfObligation"
+                },
+                "rite": {
+                    "$ref": "./CommonDef.json#/definitions/RiteWithoutNationalCalendars"
+                },
+                "diocesan_calendar": {
+                    "$ref": "./CommonDef.json#/definitions/DiocesanCalendarId"
+                }
+            },
+            "required": [
+                "year",
+                "epiphany",
+                "ascension",
+                "corpus_christi",
+                "locale",
+                "return_type",
+                "year_type",
+                "eternal_high_priest",
+                "holydays_of_obligation",
+                "rite",
                 "diocesan_calendar"
             ]
         }
--- a/phpunit_tests/Schemas/CalendarSettingsSchemaTest.php
+++ b/phpunit_tests/Schemas/CalendarSettingsSchemaTest.php
@@ -1,0 +1,344 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests\Schemas;
+
+use LiturgicalCalendar\Api\Enum\LitSchema;
+use LiturgicalCalendar\Api\Router;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+use Swaggest\JsonSchema\Schema;
+
+/**
+ * Unit tests for the `settings` property of the `LitCal.json` response schema (issue #844).
+ *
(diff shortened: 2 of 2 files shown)


JohnRDOrazio:
@coderabbitai review

Issue #761, opened 2026-08-07
Ambrosian routes accept locales the rite has no liturgical books for


JohnRDOrazio:
## Summary

`/calendars` correctly declares that the Ambrosian rite supports only Italian and Latin, but the
`/calendar/ambrosian` endpoint does not enforce it: any locale the API supports globally is accepted and
served.

**The metadata is already right** — this is about the endpoint honouring what the metadata declares.

## What the metadata declares

```jsonc
"ambrosian_calendars": [
  { "calendar_id": "ambrosian", "rite": "ambrosian", "locales": ["it", "la"] }
]
```

and every Ambrosian diocese agrees:

```text
lugano_ch    locales=['it_IT', 'la_VA']
bergam_it    locales=['it_IT', 'la_VA']
milano_it    locales=['it_IT', 'la_VA']
novara_it    locales=['it_IT', 'la_VA']
```

## What the endpoint does

```console
$ for l in it la nl en; do
    curl -s -o /tmp/r.json -w "locale=$l -> HTTP %{http_code}, " "http://localhost:8000/calendar/ambrosian?locale=$l"
    jq -r '.settings.locale' /tmp/r.json
  done
locale=it -> HTTP 200, it_IT
locale=la -> HTTP 200, la
locale=nl -> HTTP 200, nl_NL
locale=en -> HTTP 200, en_US
```

Dutch and English are accepted and echoed back as the applied locale, for a rite that has no liturgical
books in either.

## Why `it` and `la` are the correct and complete set

The Ambrosian Rite has exactly two official liturgical languages, and Latin is not vestigial — it is the
language of the *editio typica*:

- *Messale Ambrosiano* (Italian), promulgated by Cardinal Colombo, 1976
- *Missale Ambrosianum* (Latin), 1981
- Liturgy of the Hours 1984, various rituals, renewed Lectionary 2008
- Second edition of the *Messale Ambrosiano* promulgated by Archbishop Delpini "in latino e in italiano",
  in force from the First Sunday of Advent 2024 (17 November), following the *recognitio* granted by
  decree *Mediolanensis Ecclesia* of 7 December 2023

No other vernacular edition exists, and the reason is structural rather than accidental. The Ambrosian
books are promulgated by the Archbishop of Milan as *Capo Rito* with confirmation from the Dicastery, and
the Italian text is calibrated to CEI norms — the 2008 CEI Bible and the 2020 *Messale Romano* were the
[...]

PR #853, opened 2026-08-21
fix(locale): negotiate Accept-Language against the calendar's own locales (#845)

### LINKED ISSUE #845: Accept-Language should be negotiated against the requested calendar's own locale list, not filtered by exact match afterwards
`Accept-Language: fr` on `GET /events/nation/CA` returns **`en_CA`**, although Canada declares
`["en_CA", "fr_CA"]`. Under RFC 4647 §3.3.1 basic filtering the range `fr` matches the tag `fr-CA`, so a
conformant negotiator offered that pair selects `fr_CA`.

**Scope note (see the maintainer comment below):** this is about the **`Accept-Language` header**, not the
`?locale=` query parameter. Regional calendars store their data under specific regional variants, so `fr` on its
own does not name a dataset, and an explicit `?locale=fr` demanding an exact identifier is intended behaviour.
A language *range* in a header is a different thing from an exact selector, and standard matching applies to it.

## What actually happens

The negotiation and the calendar's locale list are applied at **two different points, in the wrong order**.

**Step 1 — `EventsHandler::handle()`, line ~830.** The header is negotiated against the *rite's* candidate set:

```php
$locale = Negotiator::pickLanguage($request, CalendarMetadataProvider::negotiableLocalesForRite($this->rite), LitLocale::LATIN);
```

For the Roman rite `negotiableLocalesForRite()` returns `[]` — documented as "empty for an unrestricted rite,
which is `pickLanguage()`'s own signal to use its full default list". So `fr` is negotiated against every ICU
locale, where `fr` exists as a tag in its own right. `pickLanguage()` prefers an exact match over a prefix match,
so it correctly returns **`fr`**.

At this point the requested calendar is not yet known, so nothing has consulted `["en_CA", "fr_CA"]`.

**Step 2 — after the calendar JSON is loaded**, at `EventsHandler.php:161-164` (diocesan) and `:255-258`
(national), plus `:217` and `:744`:

```php
if (!in_array($this->EventsParams->Locale, self::$NationalData->metadata->locales)) {
    $this->EventsParams->Locale = self::$NationalData->metadata->locales[0];
}
```

`'fr'` is not literally in `["en_CA", "fr_CA"]`, so it silently becomes `locales[0]` — `en_CA`.

So the right matching runs against the wrong candidate set, and the right candidate set is then applied with the
wrong matching. (The four sites are also inconsistent about strictness — `:161` and `:255` omit the third
`in_array()` argument, `:217` and `:744` pass `true`.)

## This has already bitten once, and the workaround is load-bearing

`CalendarMetadataProvider::negotiableLocalesForRite()`'s docblock records an earlier instance:

> The rite-level metadata declares `it`/`la`, but the Ambrosian diocesan layer matches the negotiated locale
> against its own full identifiers (`it_IT`/`la_VA`) with a strict `in_array()`, falling back to its first
> locale on a miss. Handing the negotiator only bare languages made it answer `la` where it used to answer
> `la_VA`, which failed that membership test — so `Accept-Language: la-VA` on
> `/events/ambrosian/diocese/milano_it` silently came back in Italian. Narrowing the *set* of acceptable
> languages must not narrow the *shape* of the tag returned for one that is acceptable.

That is this bug, diagnosed correctly, and worked around by carefully shaping the *rite* candidate set so the
negotiator happens to emit full identifiers the downstream `in_array()` will accept. The workaround is fragile
by construction: it makes a function that answers "which languages does this rite have books for" also
responsible for the *shape* of tags a completely different layer will string-compare.

## Proposed focus: negotiate against the calendar's own locale list

Replace the exact-match-plus-silent-default filter with a real negotiation, performed at the point the
calendar's declared locales are known:

```php
// once $NationalData / $DiocesanData is loaded
$locale = Negotiator::pickLanguage($request, $data->metadata->locales, $data->metadata->locales[0]);
```

**The machinery already exists** — `Negotiator::pickLanguage()` implements exactly the required semantics, per
its own docblock: exact match wins over prefix match; a requested `en` matches `en_US` and `en`, preferring the
more specific supported tag; `*` matches anything; and both hyphen and underscore spellings are accepted. It is
already called this way with a non-empty `$supported` at `EventsHandler:830` and `CalendarHandler:5499`. Nothing
new needs to be written; the existing negotiator needs to be handed the right candidate set at the right moment.

Points the implementation has to settle:

1. **The header must still be reachable at that point.** Negotiation currently happens in `handle()` while the
   filter happens after the calendar loads, so either the `ServerRequestInterface` or the raw header has to be
   available there. Prefer re-negotiating the original header over trying to widen the already-negotiated tag —
   `fr` → `fr_CA` is recoverable, but information has been discarded by then in the general case.
2. **Header and explicit parameter must stay distinguishable.** The comment at `EventsHandler:826-829` already
   records that they deliberately differ (#761): an unsupported *header* preference degrades, while an explicit
   *param* trips a rejection. Re-negotiating must not turn an explicit `?locale=fr` into `fr_CA`.
3. **`/calendar` needs the same treatment.** `GET /calendar/nation/CA/2026` behaves identically; `CalendarHandler`
   has the parallel structure.
4. **Simplify `negotiableLocalesForRite()` afterwards.** Once the calendar layer negotiates properly, that
   function no longer has to guarantee tag *shape* for a downstream string comparison, and its docblock's warning
   can be retired rather than preserved. Check `CalendarMetadataProvider::localesForRite()` and the rite-scoped
   rejection in `EventsParams::validateRiteCompatibility()` (#761) at the same time.
5. **Ambiguity needs a stated rule.** If a range matches more than one declared locale — `fr` against a
   hypothetical `["fr_CA", "fr_FR"]` — which wins? `pickLanguage()` has a rule; whatever it is should be
   asserted by a test rather than left as an emergent property.

## Tests worth having

- `Accept-Language: fr` on `/events/nation/CA` and `/calendar/nation/CA/2026` resolves to `fr_CA`. This is the
  sharpest case, because a wrong answer is a different *language*, not a different region of the same one.
- `Accept-Language: la` on `/events/ambrosian/diocese/milano_it` resolves to `la_VA` — the case the
  `negotiableLocalesForRite()` workaround exists for, asserted directly instead of via the workaround.
- `Accept-Language: fr-CA` keeps resolving to `fr_CA` (it already does — guard against regression).
- An explicit `?locale=fr` does **not** silently become `fr_CA`, per the scope note above.
- A range matching nothing the calendar declares still lands on the documented default.

## Separate question, deliberately not folded in

A **valid but unavailable** explicit locale is handled inconsistently: `?locale=zz_ZZ` is a **400**, while
`?locale=de_DE` and `?locale=fr_FR` are **200 with `en_CA`**. All three are locales the calendar has no data
for; the difference is only that `zz_ZZ` fails `LitLocale::isValid()` at the params layer while `de_DE` passes
it and is discarded at the handler layer. Whether the second should refuse or keep substituting is a policy
question independent of header negotiation. If substitution stays, it should be documented on the paths and in
the `locale` parameter description so a client knows to read `settings.locale` back. (In the API's favour:
`settings.locale` does report the locale actually used, so the response is not untruthful.)

## Related

- #761 — rite-scoped locale rejection, and the source of the deliberate header-vs-param asymmetry.
- CLAUDE.md's standing rule: always use `Negotiator::pickLanguage()`, never `\Locale::acceptFromHttp()`, because
  ICU does not know Latin.

Found while implementing #817.

JohnRDOrazio:
Closes #845.

`Accept-Language: fr` on `GET /events/nation/CA` returned **`en_CA`**, although Canada declares `["en_CA", "fr_CA"]`. Under RFC 4647 §3.3.1 basic filtering the range `fr` matches the tag `fr-CA`, so a French-speaking client was getting English with a 200 and no indication anything had been refused. `/calendar` behaved identically.

**Scope:** this is the `Accept-Language` **header**, not the `?locale=` query parameter. An explicit `?locale=fr` demanding an exact dataset identifier remains intended behaviour and is unchanged.

## No new negotiator was needed

`Negotiator::pickLanguage()` already implements the required semantics. The bug was **ordering and candidate set**:

1. `EventsHandler::handle():830` negotiated the header against the *rite's* candidate set. For the Roman rite that is `[]`, which `pickLanguage()` reads as "use my full default list" — so `fr` matched against every ICU locale, where `fr` is a tag in its own right, and exact-beats-prefix correctly returned `fr`. The requested calendar was not yet known.
2. After the calendar JSON loaded, `EventsHandler:161-164`/`:217`/`:255-258` and `CalendarHandler:457-458`/`:3494-3495` did `in_array($locale, $data->metadata->locales)` and fell back to `locales[0]`.

So the right matching ran against the wrong candidate set, and the right candidate set was applied with the wrong matching. `CalendarHandler:5494` carried a TODO asking for exactly this fix; it is now removed.

Both handlers retain the `ServerRequestInterface` and call a new `resolveCalendarLocale(array $declaredLocales)` at each filter site, **re-negotiating the original header** against the calendar's declared locales — re-negotiated rather than widened, because the first pass has already discarded the quality ordering and the alternatives.

## One deviation from the issue's proposed one-liner, and it matters

The issue suggested `pickLanguage($request, $locales, $locales[0])`. But `pickLanguage()` returns **`null`, not `$fallback`**, when a *non-empty* header matches nothing supported — the fallback argument only covers an *absent* header. Applying that literally would have turned `Accept-Language: nl` on an Ambrosian diocese from `la_VA` into `it_IT`, silently reversing #761's "an unacceptable header degrades to Latin".

So the helper only fires when the locale in hand is **not** declared — exactly where the old code discarded it anyway — and coalesces `?? $declaredLocales[0]`. A locale the calendar declares is left untouched. This is pinned as a test data row.

## Header vs. explicit parameter

`localeExplicitlyRequested` is set from `array_key_exists('locale', $requestParams)` (query for GET, parsed body for POST) before the merge. Explicit params take the unchanged exact-match + `locales[0]` branch. Asserted by `testExplicitLocaleParamIsNotWidenedToARegionalVariant` (`?locale=fr` → `en_CA` on both endpoints) and `testExplicitLocaleParamOverridesTheHeader`.

## `negotiableLocalesForRite()` kept, its warning retired

The dependency its docblock recorded is genuinely gone — the diocesan layer no longer string-compares an already-negotiated tag, and `Accept-Language: la` now reaches `la_VA` even though the rite-level pass answers bare `la`.

But its region-qualified shape remains load-bearing for an **independent** reason, verified empirically: that list is the *entire* candidate set on the rite-level routes, so returning bare `['it','la']` makes `pickLanguage` answer `it` for `it-IT` and `la` for `la-VA`, changing `settings.locale` on `/events/ambrosian` and breaking `AmbrosianLocaleEnforcementTest`. Simplifying it would be an observable contract change, so it was not simplified. Stale justifications in `CalendarMetadataProviderTest` and `AmbrosianLocaleEnforcementTest` are updated.

## Ambiguity rule, established empirically and now asserted

`fr` against `['fr_CA','fr_FR']` → `fr_CA`; against `['fr_FR','fr_CA']` → `fr_FR`. Every candidate scores identically and `pickLanguage` replaces its running best only on a *strictly* greater score, so **the calendar's declaration order wins** — the same order that supplies the `locales[0]` default. Exact still beats prefix either way: `fr` against `['fr_CA','fr']` → `fr`.

## Tests

`phpunit_tests/Handlers/CalendarScopedLocaleNegotiationTest.php`, 34 cases. Pre-fix **7 failures**: `fr` → `en_CA` instead of `fr_CA` on both endpoints (including `fr-FR,fr;q=0.9`), and `la` → `it_IT` instead of `la_VA` on `/events/ambrosian/diocese/milano_it`. Post-fix `OK (34 tests, 108 assertions)`; full `--filter Locale` `OK (205 tests, 2649 assertions)`.

In-process handler tests deliberately, not `Routes/*` — those hit `:8000`, which serves the main checkout, so a green there would say nothing about this branch.

## Verification

- `composer test:quick`: baseline `Tests: 2973` → `Tests: 3007`

One unrelated failure appears in full runs (`Handlers/Auth/NotificationsHandlerTest`). Confirmed environmental: it fails **3/3 on pristine `development`** and passes intermittently on this branch — shared-Postgres contention while several branches were being tested concurrently. Nothing here touches DB, auth or notifications.

## Out of scope

Refuse-vs-substitute for a valid-but-unavailable explicit locale (`?locale=de_DE` → 200 `en_CA` while `?locale=zz_ZZ` → 400) is unchanged: the explicit-parameter branch is behaviourally identical to before, so both outcomes stand as #845 describes.

## Summary by CodeRabbit

* **New Features**
  * Calendar and event locales now negotiate `Accept-Language` headers against the locales available for the selected national or diocesan calendar.
  * Supports language-range matching, quality preferences, wildcards and sensible fallback behaviour.
  * Explicit `locale` parameters take precedence over request headers and fall back to the calendar’s first declared locale when unmatched.
  * Ambiguous language ranges resolve consistently using the first declared matching locale.

* **Tests**
  * Added comprehensive coverage for national, diocesan and Ambrosian locale negotiation scenarios.

### DIFF
--- a/phpunit_tests/Handlers/AmbrosianLocaleEnforcementTest.php
+++ b/phpunit_tests/Handlers/AmbrosianLocaleEnforcementTest.php
@@ -226,11 +226,15 @@ final class AmbrosianLocaleEnforcementTest extends AbstractHandlerTestCase
 
     /**
      * Narrowing negotiation to the rite's languages must not narrow the *shape* of the
-     * tag it returns. The rite-level metadata declares bare languages (`it`, `la`), but
-     * the diocesan layer matches the negotiated locale against its own full identifiers
-     * (`it_IT`, `la_VA`) with a strict `in_array()` — so a negotiator that answered `la`
-     * instead of `la_VA` would fail that membership test and get silently downgraded to
-     * the diocese's first locale, turning a request for Latin into Italian.
+     * tag it returns: the rite's candidate set is all the negotiator has to answer with on
+     * the rite-level routes, so offering it only the bare `it`/`la` the rite metadata
+     * declares would report a coarser `settings.locale` than the client asked for.
+     *
+     * (This also used to be what kept the diocesan layer honest, since that layer
+     * string-compared the negotiated tag against its own `it_IT`/`la_VA`. It no longer
+     * does — #845 made it re-negotiate the original header against its own declared
+     * locales — so the diocesan rows below are now a straightforward regression guard
+     * rather than a proof of that coupling. See CalendarScopedLocaleNegotiationTest.)
      *
      * These cases pin the exact `settings.locale` each header produced before locale
      * enforcement existed; the only behavior this feature is allowed to change is what
--- a/phpunit_tests/Handlers/CalendarScopedLocaleNegotiationTest.php
+++ b/phpunit_tests/Handlers/CalendarScopedLocaleNegotiationTest.php
@@ -1,0 +1,347 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests\Handlers;
+
+use LiturgicalCalendar\Api\Enum\Rite;
+use LiturgicalCalendar\Api\Handlers\CalendarHandler;
+use LiturgicalCalendar\Api\Handlers\EventsHandler;
+use LiturgicalCalendar\Api\Http\Enum\ReturnTypeParam;
+use LiturgicalCalendar\Api\Http\Enum\StatusCode;
+use LiturgicalCalendar\Api\Http\Exception\ApiException;
+use LiturgicalCalendar\Api\Http\Negotiator;
+use Nyholm\Psr7\ServerRequest;
+use PHPUnit\Framework\Attributes\DataProvider;
+
+/**
+ * Issue #845: an `Accept-Language` **range** must be negotiated against the requested
+ * calendar's own declared `metadata->locales`, not matched against them for literal
+ * equality after having been negotiated against something else.
+ *
+ * Canada declares `["en_CA", "fr_CA"]`. Under RFC 4647 §3.3.1 basic filtering the range
+ * `fr` matches the tag `fr-CA`, so a French-speaking client sending the plain `fr` range
+ * must be served `fr_CA`. Before this fix it was served `en_CA` — a different *language*,
+ * not merely a different region of the same one — because the two halves of the decision
+ * ran in the wrong order:
+ *
+ * 1. `handle()` negotiated the header against the *rite's* candidate set, which is empty
+ *    for the Roman rite and therefore means "the API-wide locale list". `fr` exists there
+ *    as a tag in its own right, and an exact match beats a prefix match, so the header
+ *    correctly resolved to `fr`. The requested calendar was not yet known at that point.
+ * 2. Once the calendar JSON loaded, the resolved locale was tested for literal membership
+ *    in `["en_CA", "fr_CA"]` — which `fr` fails — and silently replaced by `locales[0]`.
+ *
+ * So the right matching ran against the wrong candidate set, and the right candidate set
+ * was applied with the wrong matching. The fix re-runs the *original header* through
+ * `Negotiator::pickLanguage()` once the calendar's own locales are known.
+ *
+ * The re-negotiation is deliberately narrow: it only re-opens the question when the locale
+ * already in hand is *not* one of the calendar's declared locales — i.e. exactly where the
+ * old code would have thrown it away for `locales[0]`. A locale the calendar does declare is
+ * left alone, so the #761 "unacceptable header degrades to the API-wide Latin default"
+ * behaviour survives wherever Latin is among the calendar's own locales.
+ *
+ * The header/parameter asymmetry established by #761 is preserved and asserted here: a
+ * header states a *preference* and is negotiated, while an explicit `locale` parameter is
+ * an *exact selector* naming a dataset, so `?locale=fr` is NOT widened to `fr_CA`.
+ */
+final class CalendarScopedLocaleNegotiationTest extends AbstractHandlerTestCase
+{
+    private ?string $savedServerName = null;
+
+    /**
+     * Pin `SERVER_NAME` to localhost so `Router::isLocalhost()` is true and `CalendarHandler`
+     * takes its "never touch `engineCache/`" path. The engine cache is keyed on the calendar
+     * params (locale included) but is written to a path relative to the process working
+     * directory and is blind to PHP source changes, so a cached body from an earlier revision
+     * could otherwise answer for the code under test. Mirrors
+     * {@see AmbrosianLocaleEnforcementTest::setUp()}.
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->savedServerName  = $_SERVER['SERVER_NAME'] ?? null;
+        $_SERVER['SERVER_NAME'] = 'localhost';
+    }
+
+    /**
+     * `EventsHandler::setLocale()` calls `setlocale(LC_ALL, ...)`, which persists for the
+     * lifetime of the PHPUnit process; later suites rely on gettext falling through to the
+     * untranslated (Latin) msgid rather than being left pinned to a translated catalog.
+     */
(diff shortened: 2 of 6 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
@coderabbitai review

Issue #848, opened 2026-08-21
Emit and document the canonical rite Link across /events and /data, and enumerate the missing /events Roman paths


JohnRDOrazio:
`/data` does not emit the `Link: rel="canonical"` header that `/calendar` and `/events` do, and `/events` is
missing the rite-qualified paths that would let a client discover the canonical form it is being pointed at.
Neither route family documents the header at all.

Of the two equivalent spellings of a rite-qualified route, **the explicit one is canonical**. That is already
the stated design (`src/Router.php:126-129`) and the intent is to apply it uniformly.

## 1. `Router::canonicalRiteUrl()` excludes `data`

```php
if ($riteSegmentExplicit || false === in_array($route, ['calendar', 'events'], true)) {
    return null;
}
```

Measured against the running API:

| request | `Link` header |
|---|---|
| `GET /calendar/2026` | `<…/calendar/roman/2026>; rel="canonical"` |
| `GET /events` | `<…/events/roman>; rel="canonical"` |
| `GET /events/nation/IT` | `<…/events/roman/nation/IT>; rel="canonical"` |
| `GET /data/diocese/romamo_it` | **none** |
| `GET /data/nation/IT` | **none** |

The exclusion is deliberate and its rationale is recorded at `src/Router.php:122-124`:

> `data` takes the segment but is deliberately absent from `canonicalRiteUrl()`: that header advertises the
> explicit form for cacheable read routes, and `/data` is an admin write surface where a
> `Link: rel="canonical"` on a PUT is noise.

**The concern is right; the conclusion is too broad.** `/data` is not only a write surface — `GET` and `POST`
both *read* source definitions, and those reads want the same discoverability the other families have. The
resolution is to scope the header by method rather than to withhold it from the route.

### Fix

Emit the canonical `Link` on `/data` for **read methods only — `GET` and `POST`**. Leave `PUT`, `PATCH` and
`DELETE` without it, which preserves the recorded objection exactly. Note that on `/data`, `POST` is a read verb
(the "retrieve with parameters in a request body" form of `GET`), which is why the read surface is `GET`+`POST`
rather than `GET` alone.

The method scoping is new: `canonicalRiteUrl()` currently takes no method, and `calendar`/`events` emit
regardless. Decide whether the method filter lives in the caller or in the signature, and whether
`calendar`/`events` should be scoped the same way for consistency (they accept `POST` as a read verb too, so in
practice nothing changes for them).

[...]

PR #854, opened 2026-08-21
docs(openapi): document the /data i18n sub-resource and locale parameter (#838, #839)

### LINKED ISSUE #838: openapi.json does not document the /data i18n sub-resource /data/{category}/{key}/{locale}

### LINKED ISSUE #839: The `locale` parameter is undocumented across the whole /data family

JohnRDOrazio:
Closes #838. Closes #839.

Two companion issues on the same file and the same operations, so they land together — one commit each.

## The `/data` surface, established live rather than assumed

Probed against a running server (read-only, plus authenticated write probes against the non-existent key `ZZ` so nothing could be written), then re-derived from the handler.

**`{locale}` third path segment** — works on **all seven** documented tiers, GET and POST, all 200:

```
GET/POST  /data/{nation|diocese|widerregion}/{key}/it_IT        200  translation map
GET/POST  /data/roman/{...}/{key}/it_IT                          200  translation map
GET/POST  /data/ambrosian/diocese/milano_it/it_IT                200  translation map
PUT/PATCH/DELETE  .../{key}/it_IT                                400  "Expected two path params …, received 3"
```

The write-method 400 was **established, not assumed**: unauthenticated those return 401 because auth middleware precedes the handler, so it took a login to see the real answer. `Router`'s `data` case independently wires `[GET, POST]` for a three-part path. Error surface: `404` for a locale with no stored file, `422` for an unknown key or a rite mismatch.

**`locale` parameter — #839's stated guess was wrong.** It is read from the **body on POST**, not on PUT/PATCH; PUT, PATCH and DELETE read no client-supplied locale at all:

| method | query | body |
|---|---|---|
| GET | **read** (400 on `zz_ZZ`) | ignored |
| POST | ignored | **read** (JSON, YAML, form-urlencoded) |
| PUT / PATCH / DELETE | ignored | ignored |

Two-stage validation: `400` for a locale the server does not ship, `422` for a well-formed locale the calendar does not declare (`?locale=en_US` on `IT` → "valid values for calendar IT are: it_IT").

## Keeping the two apart

The path segment is spelled **`{i18n_locale}`**, not `{locale}`, and both components state explicitly that the other is a different thing — the segment selects a stored translation map, the parameter selects the language of a calendar definition. Verified inert in combination: `/data/nation/IT/it_IT?locale=en_US` returns 200 (the calendar-locales check short-circuits for i18n requests) while `?locale=zz_ZZ` still 400s. A test fails if anyone collapses the two names.

## Schema and `$ref` choices

- **Translation map:** `LitCalTranslation.json` already described the shape exactly and is what `LitSchema::I18N` validates stored i18n files against. `$ref`'d externally via a new shared `RegionalDataI18n200` response component. A test loads a real stored file and validates it against that schema, closing the loop.
- **`locale`:** `CommonDef.json#/definitions/Locale` — the same definition `/temporale`'s `locale` parameter uses; the test compares against `/temporale`'s `$ref` rather than a literal. Deliberately **not** `AmbrosianLocale` even on the Ambrosian path, because `RegionalDataParams` accepts `fr_FR` there and the rite narrowing is a later `422`, not the `400` that `AmbrosianLocaleParam` documents.
- **`422` was missing** from the national and wider-region reads even though `?locale=en_US` produces it there — the same class of omission #843 fixed for the rite `422`.

## The diff is surgical

`openapi.json` grew 12733 → 13544 lines, **+811 insertions, zero deletions**. No reformat: this file is inconsistently escaped, so it was edited by text-level splice rather than re-serialisation. The seven new path items sit as one contiguous block before `/easter`, and the #839 edits are confined to `parameters`/`requestBody`/`responses` of existing operations, to keep #848's rebase tractable.

## Tests and mutation checks

`OpenApiDataI18nSubResourceTest` and `OpenApiDataLocaleParamTest` — 77 tests, 365 assertions — asserting the documented surface against `RegionalDataParams`/`RegionalDataHandler` rather than a second hardcoded list, in the house style of `OpenApiDataRiteSegmentTest`.

Every assertion was mutation-checked and restored: removing the ambrosian i18n path, documenting `put` on an i18n path, pointing the 200 at the wrong schema, renaming the segment to `{locale}`, dropping the param from a GET, adding it to a POST or DELETE, narrowing to `AmbrosianLocale`, dropping the 422 — each fails with a specific message.

Worth noting: the behavioural probes' **first version was vacuous** — a `null` expectation sent no locale at all, so flipping GET's expectation still passed. Restructured into an explicit where-sent × is-read grid; flipping any cell now fails.

## Verification

- `composer lint:openapi` → valid, **11 problems ignored — unchanged**, no new `.redocly.lint-ignore.yaml` entries
- `composer test:quick` → `Tests: 3050, Assertions: 178305, Skipped: 11`

The single failure in full runs is `Routes/Readonly/CalendarEtagStabilityTest`, reproduced with bare `curl` and **not from this branch** — it is an `ApiTestCase` hitting `:8000`, which serves the main checkout. Filed as #849.

## Found, deliberately not fixed — reported instead

1. **An undocumented accidental alias.** `POST /data/nation/IT` with body `{"i18nRequest":"it_IT"}` returns the translation map with no third path segment — the internal `i18nRequest` name leaks through `array_merge($params, $parsedBodyParams)`. Not documented, because enshrining an internal name as public API would be worse than the omission. Deserves its own issue, probably closed off in the handler.
2. **`CommonDef.json#/definitions/Locale` has drifted from `LitLocale`** (ICU version skew): the schema enum carries 957 entries including uppercase forms and `ks_IN`/`sd_PK`/`yi_001` which the server rejects with 400, while `LitLocale` ships ~100 the schema lacks. The `$ref` was reused anyway — it is the intended set — but `CommonDef.json` belongs to #844's branch, so it was left alone. Worth a regeneration issue.
3. `DELETE /data/nation/{key}` and `/data/widerregion/{key}` can return 422 on an unknown key but declare only `200,400,401,403,404`.
4. The 404 detail on a missing translation file leaks a server path (`The requested resource ../jsondata/sourcedata/…/i18n/bogus.json was not found`).

## Summary by CodeRabbit

* **New Features**
  * Added locale-aware retrieval for regional calendar data across GET and POST requests.
  * Added documented validation responses for unsupported and unavailable locales.
  * Expanded Ambrosian diocesan data coverage to include all standard operations.
  * Added translation-resource endpoints for national, diocesan, regional, Roman and Ambrosian calendars.
  * Translation resources are available by locale and return structured translation maps.

* **Documentation**
  * Updated the OpenAPI specification with locale parameters, request formats, response schemas and newly supported endpoints.

* **Tests**
  * Added coverage to verify locale handling, translation-resource documentation and response validation.

### DIFF
--- a/jsondata/schemas/openapi.json
+++ b/jsondata/schemas/openapi.json
@@ -7471,6 +7471,552 @@
         }
       }
     },
+    "/data/nation/{key}/{i18n_locale}": {
+      "get": {
+        "tags": [
+          "Calendar Source Data"
+        ],
+        "security": [
+          {}
+        ],
+        "summary": "Retrieve the stored translations of a national calendar",
+        "description": "Retrieve the stored translations of a national calendar's own liturgical events, as a map of `event_key` to translated name. This is a different resource from `/data/nation/{key}`, which the path hangs off: that returns the calendar definition, this returns the translation file behind it.\n\nOnly `GET` and `POST` exist at this depth. `RegionalDataHandler::validateRequestPath()` accepts two or three path segments for those two methods and exactly two for the others, and the router narrows the allowed methods to `GET` and `POST` for any three-segment `/data` path, so a `PUT`, `PATCH` or `DELETE` carrying a third segment is refused with `400 Bad Request` (`Expected two path params for PUT, PATCH and DELETE requests, received 3`). `POST` is a read alias of `GET`.\n\nThe third segment binds to `i18nRequest`. It is not the `locale` parameter that `/data/nation/{key}` accepts, and the two are documented separately for that reason.\n\nAn unknown nation is refused with `422 Unprocessable Content`; a nation that exists but has no translation file for the requested locale gives `404 Not Found`. `/data/roman/nation/{key}/{i18n_locale}` states the same rite explicitly and is the canonical form of this path.",
+        "operationId": "nationalDataI18nGET",
+        "parameters": [
+          {
+            "name": "key",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "$ref": "./CommonDef.json#/definitions/Nation"
+            }
+          },
+          {
+            "$ref": "#/components/parameters/RegionalDataI18nLocalePathParam"
+          }
+        ],
+        "responses": {
+          "200": {
+            "$ref": "#/components/responses/RegionalDataI18n200"
+          },
+          "400": {
+            "$ref": "#/components/responses/BadRequest400"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound404"
+          },
+          "422": {
+            "$ref": "#/components/responses/UnprocessableEntity422"
+          }
+        }
+      },
+      "post": {
+        "tags": [
+          "Calendar Source Data"
+        ],
+        "security": [
+          {}
+        ],
+        "summary": "Retrieve the stored translations of a national calendar",
+        "description": "Read alias of `GET /data/nation/{key}/{i18n_locale}`, returning the same translation map. See that operation for how the third path segment differs from the `locale` parameter.",
+        "operationId": "nationalDataI18nPOST",
+        "parameters": [
+          {
+            "name": "key",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "$ref": "./CommonDef.json#/definitions/Nation"
+            }
+          },
+          {
+            "$ref": "#/components/parameters/RegionalDataI18nLocalePathParam"
+          }
+        ],
+        "responses": {
+          "200": {
+            "$ref": "#/components/responses/RegionalDataI18n200"
+          },
+          "400": {
+            "$ref": "#/components/responses/BadRequest400"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound404"
+          },
+          "422": {
+            "$ref": "#/components/responses/UnprocessableEntity422"
+          }
+        }
+      }
+    },
+    "/data/roman/nation/{key}/{i18n_locale}": {
+      "get": {
+        "tags": [
+          "Calendar Source Data"
+        ],
+        "security": [
+          {}
+        ],
+        "summary": "Retrieve the stored translations of a national calendar",
+        "description": "This is the canonical form of this route: the rite is stated explicitly. The bare `/data/nation/{key}/{i18n_locale}` form is equivalent and resolves here.\n\nRetrieve the stored translations of a national calendar's own liturgical events, as a map of `event_key` to translated name. This is a different resource from `/data/roman/nation/{key}`, which returns the calendar definition itself.\n\nOnly `GET` and `POST` exist at this depth; a `PUT`, `PATCH` or `DELETE` carrying a third segment is refused with `400 Bad Request`. The third segment binds to `i18nRequest` and is not the `locale` parameter of the two-segment path.\n\nOnly the diocesan tier exists under more than one rite, so there is no `/data/ambrosian/nation/{key}/{i18n_locale}`.",
+        "operationId": "romanNationalDataI18nGET",
+        "parameters": [
+          {
+            "name": "key",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "$ref": "./CommonDef.json#/definitions/Nation"
+            }
+          },
+          {
+            "$ref": "#/components/parameters/RegionalDataI18nLocalePathParam"
+          }
+        ],
+        "responses": {
+          "200": {
+            "$ref": "#/components/responses/RegionalDataI18n200"
+          },
+          "400": {
+            "$ref": "#/components/responses/BadRequest400"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound404"
+          },
+          "422": {
+            "$ref": "#/components/responses/UnprocessableEntity422"
+          }
+        }
+      },
+      "post": {
+        "tags": [
+          "Calendar Source Data"
+        ],
+        "security": [
+          {}
+        ],
+        "summary": "Retrieve the stored translations of a national calendar",
+        "description": "Read alias of `GET /data/roman/nation/{key}/{i18n_locale}`, returning the same translation map.",
+        "operationId": "romanNationalDataI18nPOST",
+        "parameters": [
+          {
+            "name": "key",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "$ref": "./CommonDef.json#/definitions/Nation"
+            }
+          },
+          {
+            "$ref": "#/components/parameters/RegionalDataI18nLocalePathParam"
+          }
+        ],
+        "responses": {
+          "200": {
+            "$ref": "#/components/responses/RegionalDataI18n200"
+          },
+          "400": {
+            "$ref": "#/components/responses/BadRequest400"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound404"
+          },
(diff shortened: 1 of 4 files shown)


JohnRDOrazio:
All four applied — one of them with a correction to the review's own premise (`8d544b21`).

## Applied: guard the vacuous pass

Correct, and it is the same shape of bug this branch's behavioural probes were already restructured to avoid: `testOnlyTheMethodsThatAcceptAThirdSegmentAreDocumented` looped over `i18nPathKeys()` without requiring the list to be non-empty, so removing the `{i18n_locale}` paths from `openapi.json` would have made it pass having asserted nothing — the exact regression it exists to catch.

## Applied: make the schema validation the asserted outcome

`$schema->in($payload)` had its result discarded, leaving `assertIsObject()` — near-tautological after `json_decode` of an object literal — as the only assertion. A violation now reads as a schema failure rather than an unhandled exception.

## Applied: gate the default-rite branch on `combinationIsAccepted()`

Every current `PathCategory` is accepted under the default rite, so nothing changes today, but a category the default rite refused would have had its documentation *demanded* — the inverse of the intent, and inconsistent with the non-default branch, which already gated.

## Applied, with a correction: moving the write-method claim

The suggestion was to narrow `testTheWriteMethodsDeclareNoLocaleParameter` to the locale claim and "keep the 'all write methods exist' claim in the route-surface test". The narrowing is right — a future read-only `/data` tier would otherwise have failed a *locale* test with a message about a missing locale parameter.

**But that claim was not in the route-surface test.** `OpenApiDataRiteSegmentTest` only asserts that the rite-qualified path documents *the same* methods as the bare form:

```php
self::assertSame($bareMethods, $riteQualifiedMethods, '…same methods as the bare form');
```

If every `/data` path lost its write methods, both sides would still match and the suite would stay green. So narrowing the locale test alone would have dropped the coverage outright rather than relocating it.

`testEveryCalendarDefinitionPathDocumentsTheFullMethodSurface` is therefore added to the route-surface test, asserting the full `GET/POST/PUT/PATCH/DELETE` surface `Router` wires for a two-part `/data` path, and excluding the `{i18n_locale}` paths because that same `Router` case wires only GET and POST for a three-part path.

Net assertions **365 → 397**.

## Verification

- The three `/data` schema suites: `OK (93 tests, 397 assertions)`
- Mutation-checked: renaming `delete` on `/data/nation/{key}` fails the new test by name (`/data/nation/{key} must document the full method surface Router wires for a two-part /data path`); `openapi.json` restored byte-clean afterwards, `git diff` empty.

PR #855, opened 2026-08-21
fix(deploy): stop shipping dev-only files, and assert the deploy payload in CI

JohnRDOrazio:
Four tracked dev-only files were being deployed, and all four were publicly readable:

| URL | status |
|---|---|
| `/api/dev/phpstan-baseline.neon` | 200, 9337 b |
| `/api/dev/.coderabbit.yaml` | 200, 3422 b |
| `/api/dev/.redocly.lint-ignore.yaml` | 200, 1834 b |
| `/api/dev/authz/openfga-expectations.json` | 200, 883 b |

**Nothing sensitive is exposed.** `.env`, `.git/config` and `logs/` all correctly return 404, and no credentials are involved — `admin.pat`/`login-client.pat` are untracked, so they never reach the runner. The baseline is the most informative of the four, naming internal file paths and the text of every suppressed error.

## Why it happened

`rsync-exclude.txt` already listed `phpstan.neon` and `phpstan.neon.dist` — but not `phpstan-baseline.neon`, a third file in the same family. `authz/` was never listed at all. The other two are recent additions that post-date the last review of the list; `.coderabbit.yaml` shipped **within hours** of being committed in #847.

That pattern is the actual defect. The exclude list is a **denylist**, so every new dev-only file has to be remembered, and the docroot layout means forgetting one publishes it. Adding four lines fixes today's leak and leaves the next one to chance.

## The fix

**1. The four exclusions**, each with its reason recorded inline.

**2. `.github/deploy/check-deploy-payload.sh`** — enumerates what rsync would transfer, intersects it with `git ls-files` (a deploy runs from a fresh checkout, so untracked scratch files are irrelevant), and asserts the top-level entries against an **allowlist** of runtime paths:

```
src  jsondata  i18n  public  scripts  bin
composer.json  doctrine-migrations.php  .env.example
```

Each carries its justification — e.g. `composer.json` because `Router::findProjectRoot()` uses it as a root marker, `bin/` because `reconcile-outbox` is invoked server-side by systemd/cron. A new dev-only file now fails CI instead of reaching the web; adding a genuinely new runtime path is a deliberate one-line edit.

**3. Wired in twice** — a `deploy_payload` job in `main.yml` gating every PR, and a pre-flight in `deploy.yaml` before rsync, because a deploy can run from a tag or a hand-picked branch that never went through CI.

## Verification

Mutation-tested in both directions. With the exclusions reverted, the guard names all four:

```
::error::Deploy payload contains tracked entries that are not runtime files.
  unexpected: .coderabbit.yaml
  unexpected: .redocly.lint-ignore.yaml
  unexpected: authz
      authz/openfga-expectations.json
  unexpected: phpstan-baseline.neon
```

Removing only `phpstan-baseline.neon` from the list makes it name that one. With the fix applied: `Deploy payload OK: only runtime entries would ship.` Both workflow files re-parse as valid YAML.

## ⚠️ Manual cleanup still required

**Merging this does not remove the files already on the server.** Excluded paths are also protected from `--delete` — the gotcha documented at the top of `rsync-exclude.txt` — so this stops them being re-sent but fossilizes the existing copies.

They must be deleted by hand on both deploy paths:

```
rm -f  <app>/dev/phpstan-baseline.neon <app>/dev/.coderabbit.yaml <app>/dev/.redocly.lint-ignore.yaml
rm -rf <app>/dev/authz
# and the same under the production vN/ path for any that are present there
```

**Do not use `--delete-excluded`** to do this: it would wipe `.env.production`, `logs/` and `cache/`.

## The mechanism (corrected)

An earlier revision of this description said the vhost docroot is the application root. **That is wrong.** The API is served as a folder under the *frontend* docroot, via this nginx directive on the frontend domain:

```nginx
location ~ ^/api/(dev|v[3-9])(/.*)?$ {
    # If the requested file exists, serve it
    # Otherwise, route everything to the front controller
    try_files $uri /api/$1/public/index.php$is_args$args;
}
```

`try_files $uri` is the exposure: **any file physically present under `/api/{dev,vN}/` is matched by `$uri` and served straight off disk, before the front controller ever sees the request.** That is precisely why `phpstan-baseline.neon` was fetchable — it existed, so nginx served it.

PHP is not disclosed: `vendor/autoload.php` returns 200 with an empty body and `public/LitCalTestServer.php` returns 500, so `.php` under the API dir is executed rather than dumped as source. Non-PHP files are served raw.

## The durable fix is one line of nginx

Map requests into `public/` instead of the application root:

```nginx
location ~ ^/api/(dev|v[3-9])(/.*)?$ {
    try_files /api/$1/public$2 /api/$1/public/index.php$is_args$args;
}
```

With that, `/api/dev/phpstan-baseline.neon` resolves against `/api/dev/public/phpstan-baseline.neon`, which does not exist, and falls through to the front controller — the API's own 404. **The entire class of leak becomes impossible regardless of what the deploy ships**, and arbitrary non-entrypoint PHP under the app root (`vendor/autoload.php` and friends) stops being directly executable.

Only `index.php` and `LitCalTestServer.php` ship into `public/`, so nothing legitimate is lost; `LitCalTestServer.php` simply moves from `/api/dev/public/LitCalTestServer.php` to `/api/dev/LitCalTestServer.php`.

One consequence worth deciding on: `public/engineCache/` would become reachable as `/api/dev/engineCache/...`. It is already directly servable today at its longer path, so this is not new exposure — and the contents are computed calendars that the API serves publicly anyway — but if you would rather it were not addressable at all, add a `location ~ ^/api/(dev|v[3-9])/engineCache/ { return 404; }` alongside.

**This PR is still worth merging with that change**, not instead of it: the guard keeps dev-only files off the server in the first place, which matters for anything that reads the deployed directory outside nginx (backups, greps, an operator browsing the tree), and it is the only half that lives in this repo.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment validation to detect unexpected files before release.
  * Prevented development, testing, and authorisation-related files from being included in deployed packages.

* **Chores**
  * Added automated checks to confirm deployment payloads contain only approved runtime files.
  * Deployment now reports invalid payload contents and stops safely when issues are found.

### DIFF
--- a/.github/deploy/check-deploy-payload.sh
+++ b/.github/deploy/check-deploy-payload.sh
@@ -3,11 +3,19 @@
 # Fail if a deploy would ship a tracked file that is not part of the runtime
 # payload (#deploy-hygiene).
 #
-# Why this exists: the vhost docroot is the *application root*, not public/,
-# so every deployed file is readable over HTTP. rsync-exclude.txt is a
-# denylist, which means each new dev-only file at the repo root has to be
-# remembered — and twice it was not (phpstan-baseline.neon, .coderabbit.yaml),
-# both of which ended up publicly fetchable.
+# Why this exists: the API is served as a folder under the *frontend* docroot,
+# and the nginx location for it resolves `try_files $uri ...` — so any file
+# physically present under /api/{dev,vN}/ is served straight off disk before the
+# front controller sees the request. rsync-exclude.txt is a denylist, which means
+# each new dev-only file at the repo root has to be remembered — and twice it was
+# not (phpstan-baseline.neon, .coderabbit.yaml), both of which ended up publicly
+# fetchable.
+#
+# The nginx side has since been narrowed to map into public/ rather than the app
+# root, which makes the whole class unreachable over HTTP. This check is still
+# worth keeping: it stops the files reaching the server at all, which matters for
+# anything reading the deployed tree outside nginx (backups, an operator grepping
+# around), and it is the half that lives in this repo.
 #
 # This flips the guarantee: the payload is checked against an ALLOWlist of
 # top-level entries. A new dev-only file fails CI instead of reaching the web.
--- a/.github/deploy/check-deploy-payload.sh
+++ b/.github/deploy/check-deploy-payload.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+#
+# Fail if a deploy would ship a tracked file that is not part of the runtime
+# payload (#deploy-hygiene).
+#
+# Why this exists: the vhost docroot is the *application root*, not public/,
+# so every deployed file is readable over HTTP. rsync-exclude.txt is a
+# denylist, which means each new dev-only file at the repo root has to be
+# remembered — and twice it was not (phpstan-baseline.neon, .coderabbit.yaml),
+# both of which ended up publicly fetchable.
+#
+# This flips the guarantee: the payload is checked against an ALLOWlist of
+# top-level entries. A new dev-only file fails CI instead of reaching the web.
+# Adding a genuinely new runtime path is a deliberate one-line edit here.
+#
+# Only git-tracked files are considered: a deploy runs from a fresh checkout,
+# so untracked local scratch files are never on the runner.
+set -euo pipefail
+
+cd "$(dirname "$0")/../.."
+
+# Top-level entries that legitimately ship. Keep the justification with each.
+ALLOWED=(
+  "src"                     # the application
+  "jsondata"                # calendar source data + schemas
+  "i18n"                    # compiled translations
+  "public"                  # front controller + assets
+  "scripts"                 # operational scripts run ON the server (RBAC runbook)
+  "bin"                     # reconcile-outbox: systemd/cron server-side CLI
+  "composer.json"           # Router::findProjectRoot() uses it as a root marker
+  "doctrine-migrations.php" # read by MigrateHandler via PhpFile
+  ".env.example"            # deliberately shipped so admins can diff against .env.production
+)
+
+tmp="$(mktemp -d)"
+trap 'rm -rf "$tmp"' EXIT
+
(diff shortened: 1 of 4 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
Applied (`1f1fac40`).

Valid, and the repo already had the precedent I failed to follow: `jsondata_lint` in the same file is scoped to `permissions: contents: read` and checks out with `persist-credentials: false`, with a comment explaining both. `deploy_payload` does the same kind of work — it only reads the tree and enumerates what rsync would transfer, writing nothing and calling no API — so there was no reason for it to run with the default token scope.

## Also in this push: a factual correction to my own comments

Two comments I wrote (in `main.yml` and at the head of `check-deploy-payload.sh`) said *"the vhost docroot is the application root, not public/"*. **That is wrong**, and it has since been established:

The API is served as a folder under the **frontend** docroot, and the nginx location for it resolves

```nginx
location ~ ^/api/(dev|v[3-9])(/.*)?$ {
    try_files $uri /api/$1/public/index.php$is_args$args;
}
```

`try_files $uri` is the exposure — any file physically present under `/api/{dev,vN}/` is served straight off disk before the front controller sees the request. Same outcome, different cause, and the wrong cause would have sent the next reader to the wrong config file.

The script header now also records that the nginx side has since been narrowed to map into `public/`, which makes the whole class unreachable over HTTP, and why this check still earns its place: it stops the files reaching the server at all, which matters for anything reading the deployed tree outside nginx (backups, an operator grepping around), and it is the half that lives in this repo.

## Verification

- `bash -n` clean; the guard still reports `Deploy payload OK: only runtime entries would ship.`
- `main.yml` re-parses as YAML with `deploy_payload.permissions = {contents: read}` and `persist-credentials: false` on the checkout step.

PR #857, opened 2026-08-21
harden CLI entrypoints: SAPI guards, and move the WebSocket server out of public/

JohnRDOrazio:
Closes nothing on its own; hardens how the deployed tree can be entered over HTTP. Grew out of the deploy-hygiene work in #855 and a maintainer question about `LitCalTestServer.php`.

Five commits, in the order they were reasoned out.

## 1. SAPI guards on the ten `scripts/*.php` (`ef9d5c23`)

Eight of the ten ship to the server, where they are run per the RBAC runbook, and they sit under a path whose `.php` files nginx hands to php-fpm. None checked how it was invoked.

**This was not a live exposure, and an earlier claim of mine that it was is wrong.** Measured against pristine `development` with `php -S`, all ten already answer 500 with an empty body. But the reason is an accident: each carries a `#!` line, which a web SAPI treats as output, which invalidates the `declare(strict_types=1)` beneath it — so they fail to **compile** rather than run.

That accident is not a guarantee. Proven with a synthetic script in the shape a future one could easily take — shebang, no `declare`:

```
unguarded.php  ->  200   "EXECUTED - would have run"
guarded.php    ->  404   (body never runs)
```

`mint-official-key.php` is why it matters: it mints an `is_system` key exempt from the per-IP rate limit, and its own docblock rests on the claim that `is_system` is set nowhere else.

Guards are **inlined per script rather than factored into a shared `require`** — a guard that depends on resolving another path has a failure mode that one constant comparison does not — and applied to **all ten, including the two the deploy currently excludes**, because that exclusion list is a denylist of the same kind that twice let dev-only files onto the server (#855).

## 2. The last stale "docroot" comment (`67aa1e7a`)

Three comments claimed the vhost docroot is the application root. Two were corrected in #855; this one was missed and merged. The real mechanism is `try_files $uri` under the **frontend** docroot, which served any file present under `/api/{dev,vN}/` straight off disk. Same outcome, different cause — and the wrong cause sends the next reader to the wrong config file.

## 3. A guard on the WebSocket entrypoint (`2c9ce3b2`)

`LitCalTestServer.php` is the Ratchet server; it ends in `IoServer::factory(...)->run()`, binding `WS_PORT` and entering an event loop. **Unlike the `scripts/` files, nothing accidental stopped it** — no `#!`, no `declare`, so it compiled and ran. Verified on live staging: `GET /api/dev/public/LitCalTestServer.php` returned **500 with `x-powered-by`**. It executed; the only thing stopping it was the real server already holding the port.

With the WS server **down** — a restart, a crash — that request binds the port and loops forever inside an FPM worker. Repeat it and the pool is exhausted, and the worker then squats `WS_PORT` so the systemd unit cannot restart either.

Verified both directions on a spare port, since the guard must not break the real entrypoint:

| entry | result |
|---|---|
| `php bin/LitCalTestServer.php` | binds `:8899` — guard passes |
| `php -S`, GET the same file | **404**, port never bound |

## 4. Move it out of `public/` into `bin/` (`e1624c97`)

It is a CLI entrypoint, not a web one. `bin/` already ships for `reconcile-outbox`, which is likewise invoked server-side by systemd/cron. Sixteen references updated, including `deploy/systemd/litcal-websocket.service.in`. Two now-dead ignore exceptions removed (`.gitignore`'s `public/*.php` rule no longer applies; `.dockerignore` already whitelists `!bin`), and the `Dockerfile` COPY split in two.

**The move does not replace the guard.** `bin/` was previously safe from HTTP entry only because its two files carry no `.php` extension, so `location ~ \.php$` never claimed them. This file keeps its extension, so it is still handed to php-fpm at `/api/{dev,vN}/bin/LitCalTestServer.php`. The move fixes where the file lives, not whether it is reachable — and it means `location ^~ /api/{dev,vN}/bin/ { return 404; }` is now worth adding on the nginx side, where before it closed nothing.

## 5. Typo (`c487ac5b`)

`Webosocket` → `Websocket` in the start script's success message. Only occurrence in the repo.

---

## ⚠️ Deploy coordination required

`deploy/` is excluded from the rsync deploy, so updating the unit **source** does not update the installed unit under `/etc/systemd`. It still points at `public/LitCalTestServer.php` and will fail with *"No such file or directory"* on the next restart.

**`deploy/install.sh` must be re-run on the VPS** before or with the next restart. It renders the unit and restarts the service, so it also delivers the WebSocket restart that is currently outstanding. Recorded in `docs/ops/deploy-sentinel-runbook.md`.

## Verification

- `php bin/LitCalTestServer.php` binds the port from the new path (spare `WS_PORT=8898`)
- `composer lint:jsondata` still runs — the CLI-guard regression that matters most
- `check-deploy-payload.sh` still passes (`bin/` is allowlisted)
- `composer lint` clean; `composer analyse` (PHPStan level 10) `[OK] No errors`; markdownlint clean
- `vendor/bin/phpunit --filter 'Health|WebSocket'` → **OK (509 tests, 3025 assertions)**


JohnRDOrazio:
## Added: the last stale "docroot" comment (`67aa1e7a`)

Folded in at the maintainer's request rather than opened as an eighth PR — same subject (what of a deployed tree is reachable over HTTP), and it corrects wording I introduced.

Three comments claimed the vhost docroot is the application root. Two were corrected in #855; the pre-flight comment in `.github/workflows/deploy.yaml` was missed and merged with the wrong explanation. The real mechanism:

```nginx
location ~ ^/api/(dev|v[3-9])(/.*)?$ {
    try_files $uri /api/$1/public/index.php$is_args$args;
}
```

The API is served as a folder under the **frontend** docroot, and `try_files $uri` meant any file physically present under `/api/{dev,vN}/` was served straight off disk before the front controller saw the request. Same outcome, different cause — and the wrong cause would send the next reader to the wrong config file.

The comment now also records that the nginx side has since been narrowed to map into `public/`, which is what makes the class unreachable over HTTP, and what this check is still for: keeping the files off the server in the first place.

`grep` confirms no occurrence of the wrong claim remains anywhere under `.github/`. Comment-only, no behavioural change.

`development` is merged in so the branch carries the pre-flight step the comment belongs to; `check-deploy-payload.sh` still reports `Deploy payload OK`, `composer lint` clean, `composer lint:jsondata` still runs.

JohnRDOrazio:
## Added: the WebSocket entrypoint, which needed this more than `scripts/` did (`2c9ce3b2`)

`public/LitCalTestServer.php` is the Ratchet WebSocket server — systemd runs it as `php public/LitCalTestServer.php` — and it ends in `IoServer::factory(...)->run()`, binding `WS_PORT` and entering an event loop. It lives under `public/`, so nginx hands it to php-fpm and it is reachable over HTTP.

**Unlike the `scripts/` files, nothing accidental stops it.** Those carry a `#!` line above a `declare(strict_types=1)`, which a web SAPI turns into a compile error. This file has neither, so it compiles and **runs**. Verified against live staging:

```
GET /api/dev/public/LitCalTestServer.php  ->  500, x-powered-by: PHP
```

It executed. The only reason it got no further is that the real server already holds the port — an *"address in use"* failure, not a check.

**Why that is worse than it sounds.** With the WS server down — a restart, a crash — that same request binds the port and loops forever inside an FPM worker. Repeat it and the pool is exhausted. Worse, the worker then squats `WS_PORT`, so `systemctl restart litcal-websocket` fails with "address in use" and the cause is near-impossible to spot. A WS restart is currently pending, which is precisely when this is reachable.

**Verified both directions** against a spare port (`WS_PORT=8899`), since the guard has to keep the real entrypoint working:

| entry | result |
|---|---|
| CLI — `php public/LitCalTestServer.php` | binds `:8899`, guard passes |
| web — `php -S`, GET the same file | **404**, and `:8899` never bound |

Note this file is a genuine counter-example to the "docroot should be `public/`" reasoning elsewhere in this thread: pointing the docroot at `public/` would *not* have helped here, because the file is already in `public/`. Moving it to `bin/` — where it belongs, as a CLI entrypoint — is the structural fix, and is worth doing separately since it needs the systemd unit updated in the same window. This guard is what holds until then, and afterwards.

`composer lint` clean.

JohnRDOrazio:
One applied — and it caught a real error in my note, not just an ordering preference. One declined.

## Applied: run `install.sh` before the deploy (`cabc56de`)

My note said the installed unit *"will fail on the next restart"*, which implies a manual restart. **That is wrong, and understates it.** `deploy/sbin/litcal-fpm-reload.sh` restarts `litcal-websocket.service` whenever the API sentinel fires — that is, on **every** deploy:

> The WebSocket server is *restarted*, but only when the API sentinel fired. […] Without this, a deploy updates the REST API and leaves the WebSocket endpoint serving whatever it loaded at boot — indefinitely, since nothing else restarts it.

So a deploy carrying this move would land `bin/LitCalTestServer.php`, delete the `public/` copy via `--delete`, drop the sentinel, and then restart a unit still pointing at a file that no longer exists — logging `restart FAILED; sentinels kept for retry` and leaving the WebSocket server **down**, with the deploy itself reporting success.

Running `sudo deploy/install.sh` first means the unit already references `bin/` when the sentinel fires, so the deploy's own restart succeeds and nothing is needed afterwards. The runbook now says so in a heading rather than in passing.

I also recorded something an operator would otherwise misread: **`install.sh` exits non-zero when run first.** It is `set -eu` and ends in `systemctl restart`, and at that moment the new path is not on the vhost yet. That is harmless and does **not** need re-running — `render`, `daemon-reload` and `enable` all happen at lines 53–60, well before the restart at line 66, so the unit is correctly installed regardless. Without that note the natural reaction is to re-run it or assume the change did not take.

## Declined: "the deploy" → "the deployment" (line 17)

`the deploy` is used as a noun throughout this repository — `the deploy user` (×3), `the deploy drops` (×2), `the deploy token`, `the deploy path` — including in the header of the very script this runbook documents:

```sh
# Triggered by litcal-fpm-reload.path when a deploy drops <approot>/tmp/restart.txt.
```

Changing this one instance would make it the inconsistent one rather than the correct one, so the wording stands.

`composer lint:md` clean.

PR #858, opened 2026-08-22
fix(ics): stop inventing CREATED/LAST-MODIFIED, and stamp DTSTAMP in real UTC (#849)

### LINKED ISSUE #849: ICS metadata stamps: DTSTAMP is local time wearing a Z, and the three stamps should not share one value

JohnRDOrazio:
Closes #849 — and corrects the diagnosis in it.

## The issue was wrong about the cause

#849 says `DTSTAMP` is what makes the ICS ETag unstable. **It is not.** `CalendarHandler::validatorSource()` already blanks `DTSTAMP` out before the validator is computed, precisely so a regeneration stamp cannot invalidate an unchanged calendar:

```php
ReturnTypeParam::ICS => ['/DTSTAMP:\d{8}T\d{6}Z/' => 'DTSTAMP:'],
```

What it does **not** strip is `CREATED` and `LAST-MODIFIED`. Those were filled from `resolveIcalReleaseObject()`, which answered `gmdate()` — the current instant — whenever the GitHub release lookup failed. So every regeneration produced a fresh validator, a conditional GET could never answer `304`, and an ICS client re-downloaded ~370 KB each time.

Demonstrated by diffing two ICS bodies taken a second apart while the lookup was rate-limited: stripping only `DTSTAMP` (what the ETag does today) still left them different; stripping `CREATED` and `LAST-MODIFIED` as well made them identical.

**The exposure window is also wider than the issue implies.** `GHRelease.json` is cached inside the data-digest-keyed engine cache directory (`engineCache/v5_7-<digest>/`), so *any* change to `jsondata/sourcedata` or the compiled translations invalidates it. A deploy that changes calendar data, plus a rate-limited `api.github.com`, is all it takes — not a rare coincidence.

## The fix

**`resolveIcalReleaseObject()` returns `null`** instead of a fabricated date, and **`produceIcal()` omits `CREATED`/`LAST-MODIFIED`** when the release is unknown. Both are OPTIONAL in a VEVENT (RFC 5545 §3.6.1), so omitting them is legal — and honest. Stamping `CREATED` with the moment of serialization asserted something that did not happen, which is the same shape as #822/#833/#834/#835.

**`DTSTAMP` is now genuine UTC.** It was built as `date('Ymd') . 'T' . date('His') . 'Z'` — server-**local** time carrying a literal `Z`. Measured on this host: UTC `00:52`, local `02:52`, emitted `025143Z`. Two hours ahead of the instant it claimed, i.e. a malformed UTC date-time under RFC 5545 §3.3.5. It now uses `gmdate()`, and takes the release date when known so it is stable as well. The events themselves are all-day (`DTSTART;VALUE=DATE`) and carry no timezone at all — only these metadata stamps do, and theirs has to be real UTC.

Both stamps are resolved once before the loop rather than recomputed per event.

## Tests

New `IcalTimestampStabilityTest` (4 cases), in-process rather than through `ApiTestCase` — the existing `CalendarEtagStabilityTest` hits `:8000`, which serves a different checkout, so it cannot verify a branch:

1. a known release date stamps all three fields identically;
2. an unknown release date omits `CREATED`/`LAST-MODIFIED` but still emits the REQUIRED `DTSTAMP`;
3. the **ETag input is byte-identical across two generations while the release is unknown** — the actual regression;
4. `DTSTAMP` is within 120 s of the real UTC instant, which fails if it is ever built from local time again.

**Mutation-checked: all four fail against the original `CalendarHandler`** (3 errors, 1 failure), and pass against the fix.

`CalendarHandlerTest::testResolveIcalReleaseObjectFallsBackToUtcNowOnError` encoded the behaviour being removed, so it is rewritten as `...ReturnsNullOnErrorRatherThanInventingADate`, with the reasoning kept in its docblock.

## Verification

- `composer test:quick`: **no failures attributable to this change**

The 16 remaining errors are the `WebSocket/*` suite, and they reproduce **identically on pristine `development`**. They are fallout from the `bin/` move in #857: the local Docker stack starts the server with `php public/LitCalTestServer.php` (`LiturgicalCalendarFrontend/docker-compose.yml:392`) and bind-mounts `src`, `public`, `jsondata`, `logs`, `cache` — but **not `bin`** — so the file is neither at the old path nor visible at the new one. That is a frontend-repo fix, tracked separately.

## Not included

Authenticating the GitHub call or caching negative results. Both are still worth doing — they would keep `CREATED`/`LAST-MODIFIED` present more often — but neither is needed for ETag stability now that the unknown case is honest rather than invented.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved calendar metadata accuracy by ensuring timestamps use genuine UTC.
  - Calendar release metadata now reflects the actual publication date when available.
  - Ensured creation and modification timestamps remain distinct and use appropriate fallback values.
  - Prevented misleading timestamps when release information cannot be retrieved.

- **Tests**
  - Added coverage for UTC handling, timestamp validity, release-date tracking and unavailable release information.

### DIFF
--- a/phpunit_tests/Handlers/CalendarHandlerTest.php
+++ b/phpunit_tests/Handlers/CalendarHandlerTest.php
@@ -128,16 +128,15 @@ final class CalendarHandlerTest extends AbstractHandlerTestCase
 
     /**
      * When the GitHub release lookup fails (e.g. api.github.com rate-limits the server), ICS
-     * generation must not 503 — but it must not invent a date either.
+     * generation must not 503 — and this method must report "unknown" rather than pick a fallback.
      *
-     * This used to return `gmdate()`, the current instant, so that produceIcal could always emit a
-     * CREATED line. That line is hashed into the ICS ETag (unlike DTSTAMP, which validatorSource()
-     * blanks out), so an unchanged calendar got a fresh validator on every request and a
-     * conditional GET could never answer 304. It was also untrue: CREATED means when the component
-     * was created, not when this response happened to be serialized.
+     * It used to answer `gmdate()`, the current instant, so that produceIcal always had a CREATED
+     * value. That collapsed a distinction the caller needs: produceIcal dates CREATED from the
+     * release and falls back to generation time, while LAST-MODIFIED comes from the versioned cache
+     * directory's mtime and never from this value at all. Returning null keeps the decision where
+     * the difference is known.
      *
-     * Both fields are OPTIONAL in a VEVENT (RFC 5545 3.6.1), so the honest answer is null and
-     * produceIcal omits them. See #849 and IcalTimestampStabilityTest.
+     * See #849 and IcalTimestampStabilityTest for what each field ends up carrying.
      */
     public function testResolveIcalReleaseObjectReturnsNullOnErrorRatherThanInventingADate(): void
     {
--- a/phpunit_tests/Handlers/CalendarHandlerTest.php
+++ b/phpunit_tests/Handlers/CalendarHandlerTest.php
@@ -127,23 +127,26 @@ final class CalendarHandlerTest extends AbstractHandlerTestCase
     }
 
     /**
-     * When the GitHub release lookup fails (e.g. api.github.com rate-limits the
-     * server), ICS generation must not 503: it falls back to the current UTC
-     * time so produceIcal can still emit a valid CREATED line.
+     * When the GitHub release lookup fails (e.g. api.github.com rate-limits the server), ICS
+     * generation must not 503 — but it must not invent a date either.
+     *
+     * This used to return `gmdate()`, the current instant, so that produceIcal could always emit a
+     * CREATED line. That line is hashed into the ICS ETag (unlike DTSTAMP, which validatorSource()
+     * blanks out), so an unchanged calendar got a fresh validator on every request and a
+     * conditional GET could never answer 304. It was also untrue: CREATED means when the component
+     * was created, not when this response happened to be serialized.
+     *
+     * Both fields are OPTIONAL in a VEVENT (RFC 5545 3.6.1), so the honest answer is null and
+     * produceIcal omits them. See #849 and IcalTimestampStabilityTest.
      */
-    public function testResolveIcalReleaseObjectFallsBackToUtcNowOnError(): void
+    public function testResolveIcalReleaseObjectReturnsNullOnErrorRatherThanInventingADate(): void
     {
         $infoObj = (object) ['status' => 'error', 'message' => '403 rate limit exceeded'];
 
         $result = ( new \ReflectionMethod(CalendarHandler::class, 'resolveIcalReleaseObject') )
             ->invoke($this->makeHandler(), $infoObj);
 
-        self::assertIsString($result->published_at);
-        self::assertMatchesRegularExpression(
-            '/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z$/',
-            $result->published_at,
-            'Fallback published_at must be an RFC3339 UTC timestamp for the ICS CREATED field'
-        );
+        self::assertNull($result, 'a failed release lookup must report "unknown", not the current time');
     }
 
     /**
--- a/phpunit_tests/Handlers/IcalTimestampStabilityTest.php
+++ b/phpunit_tests/Handlers/IcalTimestampStabilityTest.php
@@ -1,0 +1,120 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests\Handlers;
+
+use LiturgicalCalendar\Api\DateTime;
+use LiturgicalCalendar\Api\Enum\LitGrade;
+use LiturgicalCalendar\Api\Handlers\CalendarHandler;
+use LiturgicalCalendar\Api\Http\Enum\ReturnTypeParam;
+use LiturgicalCalendar\Api\Models\Calendar\LiturgicalEvent;
+use LiturgicalCalendar\Api\Params\CalendarParams;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * The ICS representation must carry timestamps that identify the calendar, not the moment it was
+ * serialized (#849).
+ *
+ * `CREATED` and `LAST-MODIFIED` are hashed into the ETag — unlike `DTSTAMP`, which
+ * `CalendarHandler::validatorSource()` blanks out first. So filling them with `gmdate()` when the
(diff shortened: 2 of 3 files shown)


JohnRDOrazio:
Applied (`3e32f1db`). Valid, and it was **four** stale places rather than the two named — the second commit on this branch changed the behaviour and left the first commit's prose behind.

| location | said | now |
|---|---|---|
| `resolveIcalReleaseObject()` docblock summary | "without the two optional fields that date it" | the caller decides what unknown means per field |
| its `$logger->warning(...)` line | `ICS will omit CREATED/LAST-MODIFIED` | `ICS CREATED falls back to generation time` |
| its body comment | argued for saying nothing | explains why null rather than a fallback *here* |
| `CalendarHandlerTest` docblock | "the honest answer is null and produceIcal omits them" | describes the actual split |

The **log line** was the one worth catching. Documentation drift misleads a reader; a wrong warning misleads an operator reading production logs during exactly the incident it fires on — they would have gone looking for missing `CREATED` fields that are in fact present.

While correcting it I kept the `null` return and stated the reason it is still right, which the old comment had got to for the wrong reason: **the fallback is not the same for every field.** `produceIcal()` dates `CREATED` from the release and falls back to generation time, while `LAST-MODIFIED` comes from the versioned cache directory's mtime and never from this value at all. Answering `gmdate()` in the resolver collapsed that distinction and handed one invented instant to both — which is what the original code did.

`grep` confirms no occurrence of the old wording survives under `src/` or `phpunit_tests/`.

`CalendarHandlerTest` + `IcalTimestampStabilityTest` → `OK (23 tests, 99 assertions)`; `composer lint` clean; `composer analyse` (PHPStan level 10) `[OK] No errors`.

PR #860, opened 2026-08-22
fix(ics): keep LAST-MODIFIED and CREATED out of the ETag (#849)

JohnRDOrazio:
Follow-up to #858. The ICS validator neutralised `DTSTAMP` and nothing else, so the two other provenance stamps `produceIcal()` emits went into the hash raw.

## The symptom

A dev server returned a different ICS ETag on **every** request:

```
Etag: W/"9bc3d9f02a78478213282dcb8930c13a"
Etag: W/"1e2ea55ac0e163f4ac10fb70628c5d86"
Etag: W/"5e5a14dcc076b2ac42ef38b1c0209219"
```

JSON, XML and YML were all perfectly stable, which is why this read as an ICS-only flake rather than a defect.

Diffing two ICS bodies 1.2 s apart — 371,662 bytes each — they differ in exactly two properties:

| property | occurrences | in the validator? |
|---|---|---|
| `DTSTAMP` | 1096 | neutralised |
| `LAST-MODIFIED` | 1096 | **hashed raw** |

## Why `LAST-MODIFIED` moves

`LAST-MODIFIED` is the engine-cache directory's mtime, falling back to generation time when that directory cannot be resolved. That fallback is the ordinary case, not an edge one:

- localhost deliberately bypasses the cache, so `ensureCachePathExists()` never runs;
- that method is the only thing that resolves `CachePath` from the relative `engineCache/v<ver>-<digest>/` into an absolute path;
- the relative path is then resolved against the **server's** working directory.

Where that directory does not exist, `is_dir()` is false and `LAST-MODIFIED` becomes generation time, advancing every second. In the dev container:

```
WorkingDir: /var/www/html
is_dir("engineCache/v5_7-55f2f89352e2") => bool(false)
```

So a conditional request re-transferred the whole ~370 KB body for nothing — the exact defect the stable validator was introduced to remove.

`CREATED` has the same shape: normally the GitHub release date, but it falls back to generation time when that lookup fails. Dormant today, identical failure mode. Fixed here too.

## Why it looked intermittent

Same relative path. The existing live-HTTP test passes wherever `engineCache/<version>/` happens to exist relative to the server's cwd and fails where it does not — green on CI, where earlier in-process tests create that directory before `Routes/*` runs; red against a container whose docroot has none. The outcome was an accident of working directory.

That is why the new test is a **unit** test on `validatorSource()` rather than a live-HTTP one: it guards the neutralisation in every environment instead of whichever one it happens to run in. It asserts the converse as well — a genuine content difference must still move the validator, so the neutralisation cannot quietly become too broad.

## Verification

Reproduced the failing condition end-to-end (cache dir made unresolvable, so `LAST-MODIFIED` takes its generation-time fallback), then confirmed the fix against those same bodies:

```
body 1:  DTSTAMP:20260822T101400Z   LAST-MODIFIED:20260822T101400Z
body 2:  DTSTAMP:20260822T101401Z   LAST-MODIFIED:20260822T101401Z

old rule (DTSTAMP only):  e232854ce65890dae2da17d8df67bc63
                          df93dacb10a1e2e5a51068f81cc91911   <- two validators
new rule:                 one stable ETag across both
```

- `CalendarIcsValidatorSourceTest` — red before, green after
- `CalendarEtagStabilityTest` — `OK (9 tests, 18 assertions)`
- `composer lint` — clean
- `composer analyse` (PHPStan level 10) — `[OK] No errors`
- `composer test:quick` — `Tests: 3166, Failures: 1`; the single failure is `WebSocket\ExecuteValidationTest`, confirmed pre-existing by re-running it with these changes stashed (a worktree-environment failure: the containerised WebSocket server cannot reach an API on a spare port)

## Note on semantics

Neutralising all three stamps is what the weak validator already licenses. Bodies differing only in when and whence they were produced are semantically equivalent, which is precisely what `W/` asserts (RFC 9110 §8.8.1). The events themselves are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved calendar response validation so unchanged calendars retain the same validator, even when generation timestamps change.
  * Prevented unnecessary cache refreshes and downloads caused by regenerated ICS timestamp values.

* **Tests**
  * Added coverage confirming that provenance timestamp changes do not alter validation.
  * Added coverage confirming genuine calendar content changes still produce a new validator.

### DIFF
--- a/phpunit_tests/Handlers/CalendarIcsValidatorSourceTest.php
+++ b/phpunit_tests/Handlers/CalendarIcsValidatorSourceTest.php
@@ -1,0 +1,117 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests\Handlers;
+
+use LiturgicalCalendar\Api\Handlers\CalendarHandler;
+use LiturgicalCalendar\Api\Http\Enum\ReturnTypeParam;
+use LiturgicalCalendar\Api\Params\CalendarParams;
+use LiturgicalCalendar\Api\Router;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * The ICS validator source must neutralise every provenance stamp the representation embeds
+ * about itself, not only `DTSTAMP`.
+ *
+ * {@see CalendarHandler::produceIcal()} emits three of them, and all three can carry a
+ * per-request value:
+ *
+ * - `DTSTAMP` is always generation time.
+ * - `LAST-MODIFIED` is the engine-cache directory's mtime, and falls back to generation time
+ *   whenever that directory cannot be resolved — which is ordinary rather than exceptional,
+ *   since localhost deliberately bypasses the cache and never calls `ensureCachePathExists()`,
+ *   leaving `CachePath` a relative path resolved against the server's working directory.
+ * - `CREATED` is the GitHub release date, and falls back to generation time when that lookup
+ *   fails.
+ *
+ * None of the three describes the calendar; they describe when and whence this body was
+ * produced. Leaving any of them in the hash hands an unchanged calendar a fresh validator, so a
+ * conditional request re-transfers the whole body for nothing — precisely the defect the
+ * validator exists to prevent. The validator is weak (`W/`) exactly so that bodies differing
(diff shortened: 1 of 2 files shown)


JohnRDOrazio:
@coderabbitai review

Issue #745, opened 2026-07-24
Extract shared locale/LANGUAGE handling so CalendarHandler and EventsHandler don't drift


JohnRDOrazio:
## Summary

`CalendarHandler::prepareL10N()` and `EventsHandler::setLocale()` both build a locale-fallback array and manage the process-global `LANGUAGE` env var (glibc `gettext()` reads it above `LC_MESSAGES`) to set up per-request localization. This logic is duplicated and has already started to drift. It should be extracted into one shared helper so the two handlers can't diverge accidentally.

Tracking issue for the follow-up deferred from #744 (fix for #743).

## Why it wasn't done inline in #744

1. **The two build `$localeArray` differently.** `CalendarHandler` derives the region via `\Locale::getRegion()` / `\Locale::getPrimaryLanguage()` and splices region variants; `EventsHandler` uses a `baseLocale` + `strtoupper(baseLocale)` heuristic (e.g. `fr` → `fr_FR`). A clean shared helper needs these reconciled first.
2. **#742 isn't merged to `development` yet.** The `prepareL10N()` Latin-reset lives on branch `fix/prepareL10N-language-leak` (#742). Coordinating a shared extraction now would mean editing across an unmerged branch.
3. **The two now intentionally diverge on `setlocale()`-failure handling.** `prepareL10N()` throws `ServiceUnavailableException` (503) when `setlocale()` returns `false`; `EventsHandler::setLocale()` (as of #744) **degrades gracefully** — resets `setlocale(LC_ALL, 'C')`, clears `LANGUAGE`, resets ICU, and lets output fall through to the English msgid base. A naive shared helper would erase this distinction.

## Proposed work (once #742 lands)

- Extract a single helper (e.g. a trait or a small `Localization`/`LocaleResolver` service) that:
  - builds the locale-fallback array from a locale string,
  - resolves + normalizes the runtime locale via `setlocale()`,
  - sets/clears `LANGUAGE` and the ICU default deterministically,
  - handles the Latin/default reset branch.
- **Standardize on graceful degradation** for a failed `setlocale()`, rather than throwing 503.
[...]

PR #864, opened 2026-08-22
refactor(locale): share one likelySubtags reader; normalize the /events model locale

### LINKED ISSUE #749: Locale follow-ups from #745: consolidate likelySubtags readers; unify model setLocale() argument
Two minor, optional cleanups deferred from #745 (shared `LocaleConfigurator`, delivered in #746). Both were left out to keep that refactor behavior-preserving and observable-output-stable. Neither blocks anything.

## 1. Consolidate the two `likelySubtags.json` readers

Two places parse the same CLDR `supplemental.likelySubtags` data behind separate static caches:

- `CalendarParams::maximizeLocale()` — Params layer; maximizes a region-less request locale (adds script+region) for the echoed `settings.locale`.
- `Services\LocaleConfigurator::likelyRegion()` — runtime; extracts only the **region** subtag to build a glibc-acceptable `lang_REGION` candidate.

They serve different needs (full maximized locale vs. region subtag only), so a shared helper should expose both shapes without changing either caller's output. Suggested: a small `Services` helper (or a method pair) that owns loading/caching `jsondata/likelySubtags.json` once, with `maximize(string $language): string` and `regionFor(string $language): string`.

**Acceptance:** one loader/cache for `likelySubtags.json`; `CalendarParams` and `LocaleConfigurator` both consume it; no change to `settings.locale` output or to resolved runtime locales; golden master + `LocaleConfiguratorTest` stay green.

## 2. Unify the model `setLocale()` argument

The two handlers pass different values to their liturgical-event model's `setLocale()`:

- `CalendarHandler` → `LitLocale::$RUNTIME_LOCALE` (normalized, e.g. `it_IT`, or `la` for Latin) to `LiturgicalEvent::setLocale()`.
- `EventsHandler` → the raw `EventsParams->Locale` (e.g. `la_VA`) to `LiturgicalEventAbstract::setLocale()`.

This inconsistency was preserved in #746 to avoid shifting Latin model-side branching. `LiturgicalEventAbstract` branches on `self::$locale === LitLocale::LATIN_PRIMARY_LANGUAGE` (`'la'`), so passing `la_VA` vs `la` can matter. Characterize the effect on the `common_lcl` / `grade_lcl` Latin paths, then standardize on one argument (likely the normalized `runtimeLocale` now available from `ConfiguredLocale`) across both handlers.

**Acceptance:** both handlers pass the same normalized value; Latin and translated `common_lcl`/`grade_lcl` output verified unchanged (or the change is intentional and covered by a test); events + calendar suites green.

## Context
- Parent: #745 (closed, done via #746). Design/plan: `docs/superpowers/specs/2026-07-28-shared-locale-configurator-design.md`, `docs/superpowers/plans/2026-07-28-shared-locale-configurator.md`.

JohnRDOrazio:
Closes #749 — the two cleanups deferred from #745 (delivered as #746), which were left out to keep that refactor behavior-preserving and observable-output-stable.

## 1. Consolidate the two `likelySubtags.json` readers

`CalendarParams::maximizeLocale()` and `LocaleConfigurator::likelyRegion()` each parsed the same CLDR `supplemental.likelySubtags` data behind its own static cache. Both now delegate to a new `Services\LikelySubtags`, which loads and caches the map once and exposes the two shapes the callers need:

| Method | Returns | Consumer | Why that shape |
|-----------------------|-------------------------------------|----------------------|---------------------------------------------------------------|
| `maximize($language)` | full canonicalized tag `en_Latn_US` | `CalendarParams`     | maximizes a region-less request locale for `settings.locale`   |
| `regionFor($language)`| bare region subtag `US`             | `LocaleConfigurator` | glibc rejects script-bearing names, so it builds `lang_REGION` |

Both former contracts are preserved exactly, including the two *different* miss cases: `maximize()` returns its input unchanged, `regionFor()` returns `''`.

This was verified rather than assumed. A throwaway harness ran both previous implementations against the new one over **all 7788 CLDR keys**, plus misses (`qqq`, `zz`, `''`) and region-bearing inputs (`en_US`, `la`, `la_VA`, `xx-YY`): **zero divergence** on either shape.

## 2. Unify the model `setLocale()` argument

`EventsHandler::setLocale()` passed the raw `EventsParams->Locale`, while `CalendarHandler::prepareL10N()` passes the resolved runtime locale. It now passes the `runtimeLocale` from the `ConfiguredLocale` it was already building and discarding, so both handlers agree.

The characterization the issue asked for came out **narrower than feared**. Three consumers read the model's `self::$locale`, and two are already form-tolerant:

- `LitCommons::fullTranslate()` → `in_array($locale, [LitLocale::LATIN, LitLocale::LATIN_PRIMARY_LANGUAGE], true)` — accepts `la` *and* `la_VA`
- `LitGrade::i18n()` → the same check
- the three `LitMassVariousNeeds` branches in `LiturgicalEventAbstract` → strict `self::$locale === 'la'`

So `/events?locale=la` renders correct Latin `common_lcl` and `grade_lcl` today; only the Masses for Various Needs commons diverge, and those would have rendered in **English** for a Latin request (`EventsParams->Locale` defaults to `la_VA`).

**No calendar in `jsondata/sourcedata/` currently declares an MVN common**, so the defect was latent — zero observable output change — and no fixture or golden master reaches the path. Hence the dedicated model test below.

## Tests

- `LikelySubtagsTest` — pins both shapes of the shared reader, the miss contracts, and that `regionFor()` strips the script subtag.
- `MassVariousNeedsLatinLocaleTest` — pins the Latin and non-Latin MVN branches that no fixture data exercises.
- `HandlerModelLocaleParityTest` — asserts the two handlers configure their models with the same locale. **All three of its cases fail against the pre-fix handler** (`la` vs `la_VA`, `it_IT` vs `it`), so it is a real guard.

## Verification

- `composer analyse` (PHPStan level 10) — clean
- `composer lint` — clean; pre-commit and pre-push hooks ran
- `CalendarGoldenMasterTest` — 9/9 green
- `composer test:quick` — 3370 tests, 179506 assertions, 17 errors

The 17 errors are all `phpunit_tests/WebSocket/*` client read-timeouts and are **pre-existing environmental baseline, not from this change**: stashing the entire changeset and re-running that suite alone gives the identical 17/17. That suite needs a WebSocket server started from the same checkout, which this working tree does not have.

## Out of scope (flagged, not changed)

- `EventsHandler` never sets `LitLocale::$PRIMARY_LANGUAGE` / `$RUNTIME_LOCALE`, which `CalendarHandler` does. Nothing on the `/events` path reads them today, so it is inert — but it is the same family of inconsistency as item 2 if it is worth tracking.
- The three `LitMassVariousNeeds` sites still use strict `=== 'la'` while everything around them is form-tolerant. This PR fixes the caller rather than the callee, per the issue's "standardize on one argument" acceptance criterion.

## Summary by CodeRabbit

* **Improvements**
  * Improved locale handling by resolving likely language regions, including when no region is supplied.
  * Calendar and event views now consistently use the resolved runtime locale.
  * Latin-language requests display Latin text consistently, including correctly joined Masses for Various Needs.
  * Unknown or incomplete language values receive safer fallback handling.

* **Tests**
  * Added coverage for locale normalisation, translated and Latin rendering, fallback behaviour, and consistency across calendar and event views.

### DIFF
--- a/phpunit_tests/Handlers/HandlerModelLocaleParityTest.php
+++ b/phpunit_tests/Handlers/HandlerModelLocaleParityTest.php
@@ -40,6 +40,12 @@ final class HandlerModelLocaleParityTest extends AbstractHandlerTestCase
         unset($_SERVER['SERVER_NAME']);
         setlocale(LC_ALL, 'C');
         putenv('LANGUAGE');
+        // These tests drive the handlers precisely in order to mutate the models'
+        // process-global locale statics, then read them back. Restore both class
+        // defaults so a later test constructing a model without setting a locale is
+        // not silently handed this test's Italian.
+        LiturgicalEvent::setLocale(LitLocale::LATIN_PRIMARY_LANGUAGE);
+        LiturgicalEventAbstract::setLocale(LitLocale::LATIN_PRIMARY_LANGUAGE);
         parent::tearDown();
     }
 
--- a/phpunit_tests/Handlers/HandlerModelLocaleParityTest.php
+++ b/phpunit_tests/Handlers/HandlerModelLocaleParityTest.php
@@ -25,27 +25,65 @@ use LiturgicalCalendar\Api\Models\EventsPath\LiturgicalEventAbstract;
  */
 final class HandlerModelLocaleParityTest extends AbstractHandlerTestCase
 {
+    private string $savedLocale          = 'C';
+    private ?string $savedLanguageEnv    = null;
+    private string $savedIcuDefault      = 'en';
+    private string $savedPrimaryLanguage = LitLocale::LATIN_PRIMARY_LANGUAGE;
+    private string $savedRuntimeLocale   = 'en_US';
+    private bool $hadServerName          = false;
+    private ?string $savedServerName     = null;
+
     protected function setUp(): void
     {
+        // Snapshot every process-global bit of state this test (and the handlers it
+        // invokes) mutates, so tearDown() restores the world as it was found rather
+        // than imposing its own opinion of the baseline on later tests.
+        //
+        // Taken BEFORE parent::setUp() deliberately: that call can markTestSkipped()
+        // when JWT_SECRET is absent, and PHPUnit still runs tearDown() after a skip in
+        // setUp(). A snapshot taken after it would leave tearDown() restoring values it
+        // never captured.
+        $this->savedLocale          = setlocale(LC_ALL, 0) ?: 'C';
+        $languageEnv                = getenv('LANGUAGE');
+        $this->savedLanguageEnv     = false === $languageEnv ? null : $languageEnv;
+        $this->savedIcuDefault      = \Locale::getDefault();
+        $this->savedPrimaryLanguage = LitLocale::$PRIMARY_LANGUAGE;
+        $this->savedRuntimeLocale   = LitLocale::$RUNTIME_LOCALE;
+        $this->hadServerName        = array_key_exists('SERVER_NAME', $_SERVER);
+        $this->savedServerName      = $this->hadServerName ? (string) $_SERVER['SERVER_NAME'] : null;
+
         parent::setUp();
+
         // Force Router::isLocalhost() true so CalendarHandler::handle() bypasses the
         // response cache and actually runs prepareL10N().
         $_SERVER['SERVER_NAME'] = 'localhost';
+
+        // Start from a known-clean process-global locale, whatever an earlier test in
+        // this process left behind.
         setlocale(LC_ALL, 'C');
         putenv('LANGUAGE');
     }
 
     protected function tearDown(): void
     {
-        unset($_SERVER['SERVER_NAME']);
-        setlocale(LC_ALL, 'C');
-        putenv('LANGUAGE');
+        if ($this->hadServerName) {
+            $_SERVER['SERVER_NAME'] = $this->savedServerName;
+        } else {
+            unset($_SERVER['SERVER_NAME']);
+        }
+        setlocale(LC_ALL, $this->savedLocale);
+        putenv(null === $this->savedLanguageEnv ? 'LANGUAGE' : 'LANGUAGE=' . $this->savedLanguageEnv);
+        \Locale::setDefault($this->savedIcuDefault);
+        LitLocale::$PRIMARY_LANGUAGE = $this->savedPrimaryLanguage;
+        LitLocale::$RUNTIME_LOCALE   = $this->savedRuntimeLocale;
+
         // These tests drive the handlers precisely in order to mutate the models'
         // process-global locale statics, then read them back. Restore both class
         // defaults so a later test constructing a model without setting a locale is
         // not silently handed this test's Italian.
         LiturgicalEvent::setLocale(LitLocale::LATIN_PRIMARY_LANGUAGE);
         LiturgicalEventAbstract::setLocale(LitLocale::LATIN_PRIMARY_LANGUAGE);
+
         parent::tearDown();
     }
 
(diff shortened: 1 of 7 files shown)


JohnRDOrazio:
The two out-of-scope items flagged at the end of the description are now tracked as #865.

JohnRDOrazio:
Fixed in e329cffa.

Valid finding — `AbstractHandlerTestCase` does not reset those statics, and the sibling `MassVariousNeedsLatinLocaleTest` already restores the one it touches, so leaving this test dirty was inconsistent within the same PR.

Extended the fix to **both** models rather than just `LiturgicalEventAbstract`: these tests run `CalendarHandler` as well, so `LiturgicalEvent::$locale` is mutated too and would have leaked `it_IT` just the same. Verified `LiturgicalEvent::setLocale('la')` is safe to call in `tearDown()` — it builds `IntlDateFormatter`s and throws on a locale it cannot format, but `la` resolves fine.

`phpunit_tests/{Handlers,Models,Services}`: 1337 tests, 22861 assertions, green.

JohnRDOrazio:
Both fixed in be071bb2. Both were valid, and the first is worse than described.

**Skip guard (lines 103-105).** `if (200 !== \$status) markTestSkipped(...it_IT likely not installed...)` could never fire for the case it named. A missing Italian locale makes `LocaleConfigurator::configure()` **throw** `ServiceUnavailableException`, and the `ErrorHandling` middleware that renders that as a 503 is not in the in-process `AbstractHandlerTestCase` path — `handle()` is called directly. So the missing-locale case errors the test and never reaches a status code, while the guard was live only for the cases it should never have swallowed (404, 500). Now the precondition is probed directly via `setlocale()` over the same candidate list `LocaleConfigurator` tries for `it`, and 200 is asserted unconditionally.

**setUp/tearDown (lines 28-42).** Now snapshots and restores everything this test or its handlers mutate: `SERVER_NAME` (including whether it was set at all), the locale, `LANGUAGE`, the ICU default, and both `LitLocale` statics.

One deliberate deviation worth flagging. I take the snapshot **before** `parent::setUp()`, where `CalendarHandlerLocaleLeakTest` (the fullest existing example of this pattern) takes it after. I verified that PHPUnit runs `tearDown()` even when `setUp()` calls `markTestSkipped()`. So on a host without `JWT_SECRET`, a snapshot taken after `parent::setUp()` never happens, and `tearDown()` then restores values it never captured — with typed properties, from uninitialized state, which is a hard `Error`. That looks like a latent bug in `CalendarHandlerLocaleLeakTest` itself; I have left it alone as out of scope for this PR, but it is worth a look.

Verification: `phpunit_tests/{Handlers,Models,Services,Params,Enum}` — 1637 tests, 24550 assertions, green (6 pre-existing skips); PHPStan level 10 clean; `composer lint` clean. Reverting the one-line handler fix still fails all 3 cases, so it remains a real guard.

JohnRDOrazio:
Fixed in be6e6cf0. Valid — and it was an inconsistency I introduced one round earlier: `tearDown()` restored every *other* snapshotted global to its prior value while resetting these two to a hardcoded `LATIN_PRIMARY_LANGUAGE`, which is the same clobbering the previous review comment had just removed elsewhere. If an earlier test had set a locale, this silently changed state it did not own.

Both statics are now snapshotted alongside the rest — before `parent::setUp()`, for the skip-safety reason documented there — and restored with `setLocale()`.

`setLocale()` rather than a direct reflection write, deliberately: `LiturgicalEvent` derives four `IntlDateFormatter`s from the locale, so writing the static alone would leave the formatters disagreeing with it.

Verified across test classes rather than assumed: seeding both models with `fr_FR` in a preceding class and reading them back in a following one now yields `fr_FR`. Before this change the same probe yielded `la`, which is exactly the state this test did not own.

`phpunit_tests/{Handlers,Models,Services,Params,Enum}` — 1637 tests, 24556 assertions, green (6 pre-existing skips); PHPStan level 10 clean; `composer lint` clean. Reverting the one-line handler fix still fails all 3 cases.

JohnRDOrazio:
Skipping this one — the suggested change is not equivalent, and would introduce the bug the last two rounds removed.

In PHP the two values mean opposite things. Per the manual: `"0"` leaves the locale setting unaffected and only returns the current setting, while `NULL` (or `""`) **sets** the locale from the `LC_*`/`LANG` environment variables.

Measured on this host, PHP 8.4.23, under `declare(strict_types=1)`:

```
baseline current       : 'it_IT.utf8'
setlocale(LC_ALL, 0)   : returned 'it_IT.utf8' | locale AFTER = 'it_IT.utf8'
setlocale(LC_ALL, null): returned 'C'          | locale AFTER = 'C'
```

So `null` would do two harmful things at line 48, which is a *snapshot* of the locale to restore later: capture `'C'` instead of the locale actually in effect, and destroy that locale on the spot — clobbering process-global state this test does not own, which is precisely what the earlier `setUp`/`tearDown` finding asked me to stop doing. The restore in `tearDown()` would then put back a value that was never there.

`setlocale(LC_ALL, 0)` is also the established idiom here — `CalendarHandlerLocaleLeakTest:36` and `LocaleConfiguratorTest:27` both use it — and it is not a strict-types problem: `setlocale()` is variadic `mixed`, so the int argument is fine (the probe above runs under `strict_types=1`).

No change made.

PR #866, opened 2026-08-22
refactor(locale): publish the LitLocale statics from configure(); add LitLocale::isLatin()

### LINKED ISSUE #865: Locale follow-ups from #749: EventsHandler does not set the LitLocale statics; unify the Latin-locale test behind LitLocale::isLatin()
Two follow-ups surfaced while doing #749 (delivered in #864). Both are latent — neither changes any output today — and neither blocks anything. They are the same family of defect #749 item 2 belonged to: **two code paths that answer "is this request Latin?" differently.**

## 1. `EventsHandler` never sets the `LitLocale` statics

`CalendarHandler::prepareL10N()` publishes the resolved locale to two process-global statics:

```php
$configured                  = LocaleConfigurator::configure($this->CalendarParams->Locale);
LitLocale::$PRIMARY_LANGUAGE = $configured->primaryLanguage;
LitLocale::$RUNTIME_LOCALE   = $configured->runtimeLocale;
```

`EventsHandler::setLocale()` calls the same `LocaleConfigurator::configure()` but sets neither. So on the `/events` path both statics keep whatever the *previous* request in the same persistent worker left behind — or, on a cold worker, their class defaults (`'la'` and `'en_US'`, which are mutually inconsistent to begin with).

**This is inert today.** I traced every reader of both statics and none is reachable from `/events`:

| Reader | Reachable from `/events`? |
|--------------------------------------------------|--------------------------------------------------------------|
| `DecreeEventMetadata::getUrl()` | No — called only from `CalendarHandler` |
| `ReadingsCommons` | No — constructed only by `Models\Calendar\LiturgicalEventCollection` |
| `Models\Calendar\LiturgicalEventCollection` (2 sites) | No — `CalendarHandler` only |
| `Models\Calendar\LiturgicalEvent` (`month_short`/`month_long`/`day_of_the_week_*`) | No — `/events` uses the `EventsPath` models |
| `RomanTemporale` / `AmbrosianTemporale` | No — `/events` uses `EventsPath\LiturgicalEventTemporale` |
| `CalendarHandler` itself | No |

The `EventsPath` models import only `Models\Calendar\LitCommons` (which takes the locale as a parameter rather than reading the static) and `AmbrosianMissalResolver`.

So this is a **trap for future code, not a live bug**: the first `EventsPath` model or `/events` helper that reads `LitLocale::$PRIMARY_LANGUAGE` will silently pick up a stale value from an unrelated prior request — exactly the leak class of #739/#743, but through a static instead of the `LANGUAGE` env var.

**Suggested fix:** have `EventsHandler::setLocale()` set both statics from the `ConfiguredLocale` it already holds after #864, identically to `prepareL10N()`. Better still, move the assignment into `LocaleConfigurator::configure()` so no caller can forget it, and have both handlers stop doing it by hand.

**Acceptance:** the two statics hold the same values after an `/events` request as after a `/calendar` request for the same locale; no output change; events + calendar suites and golden master stay green.

## 2. Unify the Latin-locale test behind a single helper

The codebase asks "is this locale Latin?" in **eleven** places, in two mutually incompatible spellings.

**Tolerant** — accepts both `la` and `la_VA`, repeating the same literal five times:

- `Enum/LitColor.php:29`
- `Enum/LitSeason.php:158`
- `Enum/LitGrade.php:128`
- `Models/Calendar/LitCommons.php:257`
- `Models/Calendar/LitCommons.php:275`

```php
$isLatin = in_array($locale, [LitLocale::LATIN, LitLocale::LATIN_PRIMARY_LANGUAGE], true);
```

**Strict** — matches only `la`, six sites across the two event models:

- `Models/EventsPath/LiturgicalEventAbstract.php:93`, `:99`, `:104`
- `Models/Calendar/LiturgicalEvent.php:148`, `:154`, `:160`

```php
self::$locale === LitLocale::LATIN_PRIMARY_LANGUAGE
```

This split is precisely what made #749 item 2 a real (if latent) defect: the tolerant sites absorbed a `la_VA` model locale silently, so `common_lcl`/`grade_lcl` looked correct, while the strict Masses for Various Needs sites did not and would have rendered English. #864 fixed the caller — both handlers now pass the normalized `la` — but the callee asymmetry remains, and any future caller that passes `la_VA` reintroduces the same class of bug.

The strict form is not simply wrong, note: `LitMassVariousNeeds::fullTranslate(bool $latin)` takes a **bool**, whereas `LitCommons::fullTranslate(string $locale)` takes the **locale**, which is why the comparison lives at the call site in one case and inside the callee in the other.

**Suggested fix:** add `LitLocale::isLatin(string $locale): bool` and use it at all eleven sites. That removes the five duplicated literals, makes the six strict sites form-tolerant, and gives future code one obvious thing to call.

**Acceptance:** one helper; no remaining ad-hoc Latin comparison; `common_lcl`/`grade_lcl` output verified unchanged for both `la` and `la_VA` on `/calendar` and `/events`; golden master green.

## Context

- Parent: #749 (delivered via #864). See also #745 / #746 (shared `LocaleConfigurator`), and #739 / #743 for the earlier locale-leak family.
- `MassVariousNeedsLatinLocaleTest` and `HandlerModelLocaleParityTest` (added in #864) already cover the MVN branches and handler parity, so item 2 has a safety net in place before it starts.

JohnRDOrazio:
Closes #865 — the two follow-ups surfaced while doing #749 (delivered in #864). Both were latent; neither changes any output.

## 1. `configure()` publishes the `LitLocale` statics

`LocaleConfigurator::configure()` now sets `LitLocale::$PRIMARY_LANGUAGE` and `$RUNTIME_LOCALE` itself, and `CalendarHandler::prepareL10N()` drops the two lines that did it by hand.

Only `/calendar` had ever set them, so `/events` and the temporale path read whatever a prior request in the same persistent worker left behind — or, on a cold worker, the mutually inconsistent class defaults (`'la'` and `'en_US'`). No current reader of either static is reachable from those routes, so nothing is broken today; the next one added would have silently inherited a stale value, which is the leak class of #739/#743 through a static instead of the `LANGUAGE` env var.

Putting it in `configure()` rather than in each handler means no caller can forget. I checked the one thing that would have ruled this out: `configure()` has three callers — `CalendarHandler`, `EventsHandler` and `FerialEventNameGenerator` — and the third is reached only from `TemporaleHandler`, which `Router` dispatches as its own route rather than nesting inside calendar generation. So publishing cannot corrupt `$PRIMARY_LANGUAGE` mid-generation.

## 2. `LitLocale::isLatin()`

The issue said eleven sites. The real inventory is 31 comparisons in five spellings, and this PR converts the 15 where it matters:

| Spelling                                                   | Sites  | Converted |
|------------------------------------------------------------|--------|-----------|
| `in_array($locale, [LATIN, LATIN_PRIMARY_LANGUAGE], true)`   | 5      | yes       |
| `self::$locale === LATIN_PRIMARY_LANGUAGE`                   | 6      | yes       |
| `$this->CalendarParams->Locale === LATIN`                    | 3      | yes       |
| `$locale === LATIN \|\| $locale === LATIN_PRIMARY_LANGUAGE`  | 1      | yes       |
| `LitLocale::$PRIMARY_LANGUAGE === LATIN_PRIMARY_LANGUAGE`    | 16     | **no**    |

The 15 converted sites all compare a **full locale string**, where either Latin form is legitimate and the spelling therefore decides the answer — the exact asymmetry that made the `/events` Masses for Various Needs commons render in English (#749).

The 16 `$PRIMARY_LANGUAGE` comparisons are left alone deliberately: that static is a bare language subtag by construction, so comparing it to `'la'` is unambiguous and correct. Converting them would be a much larger diff across the Temporale engines and `CalendarHandler` for no reduction in risk.

One thing worth recording: the `CalendarParams` comparisons were correct but **fragile**. CLDR maximizes `la` to `la_Latn_VA`, which would fail a `=== 'la_VA'` test; they survive only because `validateLocaleParam()` falls through to reconstructing `'la_VA'` from language + region. `isLatin()` makes them independent of a normalization two layers away.

## Verification

**No observable output change, measured rather than assumed.** `/events` and `/calendar` were dumped across `la`, `la_VA`, `it` and `en` — 8 endpoint/locale combinations, ~3400 events' worth of `grade_lcl`, `common_lcl`, `color_lcl`, `month` and `day_of_the_week_long` — on this branch and on `development`. The two dumps are **byte-identical**. `engineCache` was cleared before each run, since its digest hashes only `jsondata/sourcedata` and the `.mo` files and would otherwise have served the other revision's cached calendars.

This mattered here: category 2 above genuinely *widens* behaviour (`la_VA` now takes Latin branches it previously missed), so "the tests pass" would not have been sufficient evidence on its own.

- `CalendarGoldenMasterTest` — 9/9 green
- `composer test:quick` — 3385 tests, 179534 assertions, 17 errors
- PHPStan level 10 clean; `composer lint` clean

The 17 errors are all `phpunit_tests/WebSocket/*` client read-timeouts — the same pre-existing local baseline as #864, where CI ran the same suite green. That suite needs a WebSocket server started from the same checkout, which this working tree does not have.

**New tests**, each verified to fail when the corresponding change is reverted:

- `LitLocaleIsLatinTest` — both accepted Latin forms, plus the near-misses (`la_Latn_VA`, `la-VA`, `la_va`) that must **not** count as Latin.
- `LocaleConfiguratorTest` — asserts `configure()` publishes on both the translated and the Latin reset branch, and now snapshots/restores the statics it started mutating.
- `HandlerModelLocaleParityTest` — poisons the statics to `xx`/`xx_XX`, then asserts an `/events` request sets them, so a pass proves the route publishes rather than merely agreeing with a leftover value.

Neutering `publish()` fails 3 of these.

## Summary by CodeRabbit

* **New Features**
  * Added consistent support for the recognised Latin locales `la` and `la_VA`.
  * Locale configuration now reliably applies the resolved language and runtime locale.

* **Bug Fixes**
  * Improved Latin-language translations and formatting across calendars, events, grades, seasons, colours and other content.
  * Prevented mixed Latin and non-Latin output when using supported Latin locales.
  * Expanded coverage for Latin, non-Latin and invalid locale behaviour.

### DIFF
--- a/phpunit_tests/Enum/LitLocaleIsLatinTest.php
+++ b/phpunit_tests/Enum/LitLocaleIsLatinTest.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LiturgicalCalendar\Tests\Enum;
+
+use LiturgicalCalendar\Api\Enum\LitLocale;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Covers LitLocale::isLatin(), the single Latin-locale test introduced in #865 to
+ * replace four different ad-hoc spellings across src/.
+ *
+ * Both accepted forms must answer true: `la_VA` is what the request layer settles on
+ * (CalendarParams, EventsParams), while `la` is what LocaleConfigurator resolves at
+ * runtime, since Latin has no installable system locale. Recognising only one of them
+ * is what made the /events Masses for Various Needs commons render in English (#749).
+ */
+#[CoversClass(LitLocale::class)]
+final class LitLocaleIsLatinTest extends TestCase
+{
+    /**
+     * @return array<string,array{0:string}>
+     */
+    public static function latinForms(): array
+    {
+        return [
+            'primary language subtag' => [LitLocale::LATIN_PRIMARY_LANGUAGE],
+            'full locale'             => [LitLocale::LATIN],
+            'literal la'              => ['la'],
+            'literal la_VA'           => ['la_VA'],
+        ];
+    }
+
+    #[DataProvider('latinForms')]
+    public function testBothAcceptedLatinFormsAreRecognized(string $locale): void
+    {
+        self::assertTrue(LitLocale::isLatin($locale));
+    }
+
+    /**
+     * @return array<string,array{0:string}>
+     */
+    public static function nonLatinLocales(): array
+    {
+        return [
+            'Italian'                  => ['it_IT'],
+            'Italian primary language' => ['it'],
+            'English'                  => ['en_US'],
+            'Latin American Spanish'   => ['es_419'],
+            'empty string'             => [''],
+            // Deliberately not Latin: the CLDR-maximized form is never what reaches the
+            // rendering code, and treating it as Latin would paper over a params-layer
+            // regression rather than surface it.
+            'CLDR-maximized Latin'     => ['la_Latn_VA'],
+            'hyphenated Latin'         => ['la-VA'],
+            'lowercased region'        => ['la_va'],
+        ];
+    }
+
+    #[DataProvider('nonLatinLocales')]
+    public function testOtherLocalesAreNotLatin(string $locale): void
+    {
+        self::assertFalse(LitLocale::isLatin($locale));
+    }
+}
--- a/phpunit_tests/Handlers/HandlerModelLocaleParityTest.php
+++ b/phpunit_tests/Handlers/HandlerModelLocaleParityTest.php
@@ -21,6 +21,10 @@ use LiturgicalCalendar\Api\Models\EventsPath\LiturgicalEventAbstract;
  * models branch on the strict primary-language form ('la') for the Masses for Various
  * Needs commons, so the two handlers disagreed on that path.
  *
+ * Also covers #865: LocaleConfigurator::configure() now publishes
+ * LitLocale::$PRIMARY_LANGUAGE and $RUNTIME_LOCALE, so /events sets them too rather
+ * than inheriting whatever a prior request in the same worker left behind.
+ *
  * @see \LiturgicalCalendar\Tests\Models\EventsPath\MassVariousNeedsLatinLocaleTest
  */
 final class HandlerModelLocaleParityTest extends AbstractHandlerTestCase
@@ -167,5 +171,27 @@ final class HandlerModelLocaleParityTest extends AbstractHandlerTestCase
         $eventsLocale = self::modelLocale(LiturgicalEventAbstract::class);
 
         self::assertSame($calendarLocale, $eventsLocale, 'The two handlers must configure their models with the same locale (#749).');
+    }
+
+    public function testEventsRequestPublishesTheLitLocaleStatics(): void
+    {
+        // Poison the statics first, so a pass proves /events actively sets them rather
+        // than merely agreeing with a value some earlier request happened to leave.
+        LitLocale::$PRIMARY_LANGUAGE = 'xx';
+        LitLocale::$RUNTIME_LOCALE   = 'xx_XX';
+
+        $response = ( new EventsHandler() )->handle($this->requestFor('GET', '/events', [])->withQueryParams(['locale' => 'la_VA']));
+        self::assertSame(200, $response->getStatusCode());
+
+        self::assertSame(
+            LitLocale::LATIN_PRIMARY_LANGUAGE,
+            LitLocale::$PRIMARY_LANGUAGE,
+            '/events must publish LitLocale::$PRIMARY_LANGUAGE (#865).'
+        );
+        self::assertSame(
+            LitLocale::LATIN_PRIMARY_LANGUAGE,
+            LitLocale::$RUNTIME_LOCALE,
+            '/events must publish LitLocale::$RUNTIME_LOCALE (#865).'
+        );
     }
 }
--- a/phpunit_tests/Models/Calendar/LitCommonsLatinFormsTest.php
+++ b/phpunit_tests/Models/Calendar/LitCommonsLatinFormsTest.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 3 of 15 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
Fixed in 8f66ffed. Valid, and it caught a genuine gap in my inventory.

**Why I missed them.** My category-B grep matched the literal `self::$locale === LATIN_PRIMARY_LANGUAGE`. These two sites use the plain `$locale === LATIN_PRIMARY_LANGUAGE` — a sixth spelling my enumeration never counted. They are not a scope expansion: the compared value is a full locale string, and it is in fact the *same* `$locale` parameter as the two sites in the same class I did convert at `i18n()` and `getPossessive()`.

**The defect was worse than a style inconsistency.** `fullTranslate()` assembles one string from four locale-sensitive pieces — the `De Commune` prefix, the name via `i18n()`, the possessive via `getPossessive()`, and the `vel` glue. Two accepted either Latin form, two matched only `la`. Handed `la_VA` it returned half Latin and half English:

```
From the Common Martyrum: … ; or …
```

All four now agree.

**I re-swept the rest of that spelling** rather than fixing only what was reported. The other twelve occurrences are out of scope, deliberately and for the reason already agreed for the sixteen `$PRIMARY_LANGUAGE` comparisons: eleven are `$locale = LitLocale::$PRIMARY_LANGUAGE` aliases in `CalendarHandler`, assigned on the line immediately before use, and the twelfth is `$primaryLanguage` in `LocaleConfigurator` from `\Locale::getPrimaryLanguage()`. All bare language subtags by construction, so comparing them to `la` is unambiguous.

**Validation.** New `LitCommonsLatinFormsTest` pins that both Latin forms give identical output, that neither is a mixture, and that a non-Latin locale still takes the translated path — so the widening cannot be mistaken for "always Latin". Reverting just these two sites fails 2 of its 4 cases.

The `/events` and `/calendar` dumps across `la`, `la_VA`, `it`, `en` remain **byte-identical to development**, engineCache cleared between runs. Golden master 9/9. `phpunit_tests/{Models,Enum,Handlers,Services,Params}` — 1656 tests, 24592 assertions, green. PHPStan level 10 and `composer lint` clean.

PR #869, opened 2026-08-22
test: decide handler/repository availability at class level so teardown cannot fatal

### LINKED ISSUE #868: CalendarHandlerLocaleLeakTest errors instead of skipping when JWT_SECRET is unset
`phpunit_tests/Handlers/CalendarHandlerLocaleLeakTest` snapshots its process-global state **after** `parent::setUp()`. That parent can `markTestSkipped()`, and PHPUnit still runs `tearDown()` after a skip in `setUp()` — so on a host where the skip fires, `tearDown()` reads properties the snapshot never assigned.

Every one of the seven is typed with no default initializer, so the read is a fatal `Error`, not a wrong value:

```php
final class CalendarHandlerLocaleLeakTest extends AbstractHandlerTestCase
{
    private string $savedLocale;
    private ?string $savedLanguageEnv;
    private string $savedIcuDefault;
    private string $savedPrimaryLanguage;
    private string $savedRuntimeLocale;
    private bool $hadServerName;          // <- read first in tearDown()
    private ?string $savedServerName;

    protected function setUp(): void
    {
        parent::setUp();                  // <- can markTestSkipped('… requires JWT_SECRET …')
        $this->savedLocale = setlocale(LC_ALL, 0) ?: 'C';
        // … the rest of the snapshot …
    }

    protected function tearDown(): void
    {
        if ($this->hadServerName) {       // <- Error: must not be accessed before initialization
```

`AbstractHandlerTestCase::setUp()` skips when `JWT_SECRET` is absent or shorter than 32 characters, and separately when `$requiresDatabase` is set and the DB credentials are missing.

## Impact

A contributor without `JWT_SECRET` configured — a fresh clone, or a fresh git worktree, since `.env.local` is gitignored and `git worktree add` does not copy it — sees this test **error** where every other handler test skips cleanly. The failure message (`Typed property … must not be accessed before initialization`) points at the test's own teardown and says nothing about the missing credential, so it reads as a broken test rather than an unconfigured environment.

## Evidence

The mechanism is confirmed, not assumed. A throwaway probe on PHPUnit 12.5.22 / PHP 8.4.23:

```php
final class ProbeSkipTeardownTest extends TestCase
{
    private string $typedUninitialized;
    protected function setUp(): void  { $this->markTestSkipped('probe'); }
    protected function tearDown(): void { fwrite(STDERR, $this->typedUninitialized); }
    public function testNothing(): void { self::assertTrue(true); }
}
```

```
1) ProbeSkipTeardownTest::testNothing
Error: Typed property ProbeSkipTeardownTest::$typedUninitialized must not be accessed before initialization

Tests: 2, Assertions: 0, Errors: 1, Skipped: 1.
```

So `tearDown()` does run after a skip in `setUp()`, and the uninitialized read is fatal. `CalendarHandlerLocaleLeakTest` matches that shape exactly.

I have not reproduced it end to end, because this host has `JWT_SECRET` configured and `bootstrap.php` loads `.env.local` via `Dotenv::createMutable()`, which overwrites shell environment variables — so the skip cannot be forced from the command line. To reproduce: run the file from a checkout with no `.env.local`, or with `JWT_SECRET` set to fewer than 32 characters.

## Suggested fix

Take the snapshot **before** `parent::setUp()`, so it always happens:

```php
protected function setUp(): void
{
    $this->savedLocale = setlocale(LC_ALL, 0) ?: 'C';
    // … the rest of the snapshot …

    parent::setUp();

    $_SERVER['SERVER_NAME'] = 'localhost';
}
```

`HandlerModelLocaleParityTest` (added in #864) already does this and documents why. Giving the properties safe defaults as well is cheap belt-and-braces, but it is not sufficient on its own: on a skip the defaults would make `tearDown()` restore values it never captured — e.g. unsetting a `SERVER_NAME` that was already set.

## Scope

Only this one class hard-errors. `AmbrosianLocaleEnforcementTest` and `CalendarScopedLocaleNegotiationTest` also snapshot after `parent::setUp()`, but their `$savedServerName` has a `= null` default, so on a skip they merely unset a `SERVER_NAME` they did not set — wrong, but not fatal. Worth tidying in the same pass.

## Context

Found while addressing review feedback on #864; noted there rather than fixed, as out of scope for that PR.

JohnRDOrazio:
Closes #868.

## The bug

PHPUnit runs `tearDown()` **even when `setUp()` raises a skip**. A class that snapshots process state *after* a skipping `parent::setUp()` therefore leaves its own `tearDown()` reading properties that were never assigned — fatal for typed properties without defaults.

On a host with no `JWT_SECRET` — a fresh clone, or a fresh worktree, since `.env.local` is gitignored and `git worktree add` does not copy it — two classes errored instead of skipping, with a message pointing at their own teardown rather than at the missing credential:

```
Error: Typed property CalendarHandlerLocaleLeakTest::$hadServerName must not be accessed before initialization
Error: Typed property DecreesHandlerWriteTest::$backupDir       must not be accessed before initialization
```

The issue named one class. Scanning all 302 test files for the shape found **two**; `DecreesHandlerWriteTest` was missed when filing.

## The fix

**Root cause, at the base classes.** `AbstractHandlerTestCase` and `RepositoryTestCase` now decide availability in `setUpBeforeClass()` rather than `setUp()`. A skip raised there aborts the whole class — PHPUnit runs neither `setUp()` nor `tearDown()` — so the trap is unreachable for every subclass, present and future. Both conditions are process env and therefore class-invariant, so per-test checking bought nothing.

Ordering matters and is deliberate: the checks run **before** any process-global state is pinned, because a class-level skip also means `tearDownAfterClass()` never runs to unpin it.

**A stale comment corrected.** `RepositoryTestCase` deferred its skip to `setUp()` on purpose, commenting that suite-level skip reasons are hidden without `--debug` while per-test ones are shown. That is not true on PHPUnit 12. Measured:

| Run | class-level reason | per-test reason |
|-----------------------|--------------------|-----------------|
| default               | hidden             | hidden          |
| `--display-skipped`   | shown              | shown           |

So the deferral bought nothing and cost the bug.

**The two classes are also made robust independently of the base**, because they want *different* fixes:

- `CalendarHandlerLocaleLeakTest` **snapshots ambient state** → its snapshot moves before `parent::setUp()`; that state exists whether or not the test runs.
- `DecreesHandlerWriteTest` **allocates a resource** (a copy of the decrees tree) → the allocation stays after `parent::setUp()`, since copying it for a skipped run is pure waste; instead `$backupDir` gains a `''` default and `tearDown()` returns early when it was never allocated.

`AmbrosianLocaleEnforcementTest` and `CalendarScopedLocaleNegotiationTest` move their snapshots too. They could not fatal — their properties are defaulted — but on a skip they unset a `SERVER_NAME` they never set.

**`ApiTestCase` is deliberately untouched.** It `fail()`s rather than skipping, so it has no such hazard. Its comment repeats the same stale belief about skip-message visibility, but changing `fail` to `skip` would change which runs are green — a separate decision, not a bug fix.

## Verification

**Reproduced end to end**, which the issue could not do: setting `JWT_SECRET` to a short value in the worktree gives, on `development`, 5 errors across the two classes; on this branch, **23 clean skips** carrying the `requires JWT_SECRET (32+ chars)` reason.

New `AbstractHandlerTestCaseSkipContractTest` pins both PHPUnit behaviours the fix rests on, by running fixture classes in a nested phpunit process:

1. `tearDown()` **does** run after a skip in `setUp()` — the trap itself.
2. A skip in `setUpBeforeClass()` runs **neither** hook — what makes the fix work.

If a future PHPUnit changes either, that test fails loudly rather than letting the base class quietly stop protecting anything.

- `composer test:quick` — 3391 tests, 179554 assertions, 11 skipped. The 17 errors are the pre-existing local `phpunit_tests/WebSocket/*` read-timeouts, unchanged in count and green in CI.
- `CalendarGoldenMasterTest` — 9/9
- PHPStan level 10 clean; `composer lint` clean
- Test-count accounting: 3389 on `development` + 2 new = 3391, so nothing that used to run is now being skipped.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved test setup and cleanup when required database or authentication prerequisites are unavailable.
  - Prevented skipped test classes from triggering incomplete cleanup or state restoration.
  - Improved handling of locale and server state during test initialisation.

- **Tests**
  - Added coverage for PHPUnit skip behaviour, ensuring skipped tests are reported correctly without causing lifecycle errors.
  - Improved repository and handler test reliability when external services cannot be reached.

### DIFF
--- a/phpunit_tests/Handlers/AbstractHandlerTestCase.php
+++ b/phpunit_tests/Handlers/AbstractHandlerTestCase.php
@@ -56,6 +56,36 @@ abstract class AbstractHandlerTestCase extends TestCase
 
     public static function setUpBeforeClass(): void
     {
+        // Decide whether this class can run BEFORE mutating any process-global state.
+        //
+        // A skip raised here aborts the whole class: PHPUnit then runs neither setUp()
+        // nor tearDown() nor tearDownAfterClass() for it. That is what makes the skip
+        // safe — a subclass tearDown() can never observe a snapshot its setUp() never
+        // got round to taking (#868). It is also why these checks must come first:
+        // anything pinned above a skip would never be restored.
+        //
+        // Both conditions are class-invariant (process env), so per-test checking bought
+        // nothing. Message visibility is identical either way: on PHPUnit 12 neither a
+        // per-test nor a suite-level skip reason is printed without --display-skipped,
+        // and both are printed with it.
+        $secret = self::env('JWT_SECRET');
+        if ($secret === null || strlen($secret) < 32) {
+            self::markTestSkipped(
+                'Handler test requires JWT_SECRET (32+ chars) in env. '
+                . 'See CLAUDE.md for the recommended values.'
+            );
+        }
+
+        if (static::$requiresDatabase) {
+            self::$pdo = self::connectToDatabase();
+            if (self::$pdo === null) {
+                self::markTestSkipped(
+                    'Handler test requires Postgres credentials in DB_HOST/DB_NAME/DB_USER/DB_PASSWORD. '
+                    . 'CI sets these via .env.local; locally, run scripts/init-db.sql.'
+                );
+            }
+        }
+
         // Pin Router::$apiPath + Router::$apiFilePath so handlers that read
         // them (for self-links + JsonData::*->path() filesystem lookups)
         // get stable, predictable values in tests. isset() is false for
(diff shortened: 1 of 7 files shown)


JohnRDOrazio:
Fixed in f96727e3. Valid, and it was a regression I introduced in this very PR — the most useful kind of catch.

Moving the skip out of `setUp()` and into `setUpBeforeClass()` left the pre-existing early `return` on missing `DB_*` credentials sitting **above** the new class-level check, so that path bypassed the skip entirely. And `setUp()` no longer had one. Repository tests then ran against a null connection.

**Measured, not reasoned about.** With `DB_HOST` unset:

```
before this fix:  Tests: 17, Assertions: 0, Errors: 17
after  this fix:  Tests: 17, Assertions: 0, Skipped: 17
                  "Repository tests require Postgres credentials in DB_HOST/…"
```

**Fix.** Restructured the way `AbstractHandlerTestCase` already was in the previous commit: `connect()` returns `?PDO` and every unavailable path *returns* null rather than returning early from `setUpBeforeClass()`. The single `self::$pdo === null` check can no longer be bypassed, so the class either has a working connection or never runs — no third state. `setUp()` can therefore dereference `self::$pdo` directly again, without the null-safe operator I had reached for.

Both unavailable branches verified to skip with their own message, since they take different routes through `connect()`:

- missing credentials → `Repository tests require Postgres credentials in DB_HOST/…`
- unreachable server → `Postgres unreachable: SQLSTATE[08006] … No route to host`

I also re-checked whether `AbstractHandlerTestCase` had the same hole rather than assuming the pattern was safe there: it does not — its helper already returned null instead of returning early, which is exactly why extracting one here fixes it.

Validation: Repositories + Handlers with credentials present — 704 tests, 7021 assertions, green. `composer test:quick` unchanged at 3391 tests, 11 skipped, the same 17 pre-existing WebSocket read-timeouts. PHPStan level 10 and `composer lint` clean.

Issue #793, opened 2026-08-16
Authorization: one unknown FGA object type zeroes every user's scope map


JohnRDOrazio:
## Summary

`ResourceAdminService` wraps its whole object-type loop in a single `try`, so a `RuntimeException` from **any one**
type discards the scopes collected for **all** of them and returns a fully-empty scope map. A single unknown or
misconfigured object type therefore degrades into a silent, global authorization-gating outage rather than a
localized one.

## This already happened

API #785 added the `rite_calendar_test` object type to `VIEWER_OBJECT_TYPES` / `TEST_OBJECT_TYPES`, but the model
change went into a file the runbook had declared dead instead of a cdcf-infra PR, so the type never reached the
deployed model. `listObjects` on it returned `400 type_not_found`, which surfaced as an `OpenFgaApiException`
(a `RuntimeException`) — and the loop-wide catch then emptied every type's list.

The visible effect was not "test scopes are missing". It was **dashboard card gating collapsing for every user**,
reddening `rbac/07-dashboard-card-scoping`, `rbac/12`, `rbac/13`, `rbac/15` in the frontend e2e suite. The cause
took a dedicated investigation to find, because the symptom pointed nowhere near the missing type.

cdcf-infra#31 added the type and has been deployed, so today's trigger is gone. **The amplifier is not.** The next
object type added to those constants ahead of its model change reproduces the same global outage.

## Suggested fix

Move the `try` inside the loop so a failure degrades only the type that failed, and log which type failed and why.

Two follow-on decisions worth making deliberately rather than by default:

- **Fail-closed per type or fail-closed globally?** Returning partial scopes on a partial failure is the
  less-surprising behaviour, but it does mean a user briefly sees fewer cards rather than none. Either is
  defensible; silently returning *nothing* with no log line is not.
- **Should an unknown object type be loud at startup?** A cheap assertion that every type in
  `ADMIN_OBJECT_TYPES` / `VIEWER_OBJECT_TYPES` / `TEST_OBJECT_TYPES` exists in the deployed model would have
  caught this at deploy time instead of via a red e2e suite days later. cdcf-infra has a consumer-expectations
  validator (its PR #30 registers this API in it) that may be the right home for that check.

## Found

During the whole-branch review of #792 (rite-aware `/tests`), while diagnosing e2e failures that turned out to be
unrelated to that branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #711, opened 2026-07-14
Evaluate parallelizing OpenFGA listObjects fan-out in the scope endpoints


JohnRDOrazio:
## Context

The three scope endpoints resolve a user's OpenFGA relations with one **sequential** `listObjects` HTTP round-trip per object type:

- `GET /auth/admin-scopes` — 4 calls (`ResourceAdminService::resolveScopes()`)
- `GET /auth/test-scopes` — 6 calls (`resolveTestScopes()`)
- `GET /auth/dashboard-scopes` — 8 calls (`resolveScopes()` + `resolveViewerScopes()`), added in #710

The dashboard-scopes calls sit on the frontend admin dashboard's server-side render path (blocking before HTML; the frontend memoizes per request and global admins skip the call entirely, so this fires once per non-admin dashboard render). CodeRabbit flagged the sequential fan-out on #710 (see [the discussion](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/pull/710#issuecomment-4968228138)); it was deliberately deferred there because it is a cross-cutting transport change.

## The question

Is the sequential fan-out a measurable problem worth changing the transport abstraction for?

Constraints that make this non-trivial:

- **OpenFGA has no batch ListObjects.** `BatchCheck` exists only for the Check API, so server-side batching is not an option; the only lever is client-side concurrency.
- **`OpenFgaClient` is built on PSR-18** (`ClientInterface::sendRequest()`), which is synchronous by contract. Concurrency requires either widening the abstraction to Guzzle-specific promises (`sendAsync()` + `Promise\Utils::all`), a `curl_multi` layer behind a custom interface, or accepting a Guzzle hard dependency.
- Each new scope surface has grown the fan-out (4 → 6 → 8); future resource cards would add more types.

## Plan

1. **Measure first**: instrument or benchmark `resolveScopes()` + `resolveViewerScopes()` against a production-like OpenFGA (same-host/in-cluster). If the 8-call sequence is low-single-digit milliseconds total, close this as not-warranted and record the numbers here.
2. If warranted, prototype the least invasive option: an optional `listObjectsMany(array $queries): array` on `OpenFgaClient` that uses Guzzle's concurrent pool when the injected client is a Guzzle client, falling back to the sequential loop for any other PSR-18 client (keeps the constructor contract and the test seams intact — the existing MockHandler-based tests keep working).
[...]

PR #880, opened 2026-08-23
fix(auth): bound the OpenFGA scope fan-out, and collapse its duplicated handler plumbing

### LINKED ISSUE #878: Bound the OpenFGA scope fan-out with a wall-clock budget so a hung OpenFGA cannot pin php-fpm workers for 45s
## The problem

`OpenFgaClient::fromEnv()` builds its Guzzle client with `['timeout' => 5, 'connect_timeout' => 2]`
— **per call**, with no budget across a sequence of calls. The scope endpoints issue a fan-out of
sequential `listObjects` calls through that one client:

| endpoint                  | resolver                                     | calls |
|---------------------------|----------------------------------------------|-------|
| `GET /auth/admin-scopes`  | `resolveScopes()`                            | 4     |
| `GET /auth/test-scopes`   | `resolveTestScopes()`                        | 8     |
| `GET /auth/dashboard-scopes` | `resolveScopes()` + `resolveViewerScopes()` | 9     |

If OpenFGA **hangs** rather than refusing connections — TCP accepts, then nothing comes back — each
call burns its full 5 s read timeout, `ResourceAdminService::listObjectsIsolated()` fails closed for
that one type and moves on to the next, and `/auth/dashboard-scopes` occupies a php-fpm worker for
up to **45 s**. A refused connection is harmless by contrast: it fails fast on `connect_timeout`.

Three things make this worse than it first looks:

1. **The caller is already gone.** `AuthHelper::fetchDashboardScopes()` in the frontend uses its own
   `['timeout' => 5, 'connect_timeout' => 2]` and returns the fail-closed result after 5 s. The API
   keeps working for another 40 s on a response nobody will read. PHP only notices the disconnect
   when it next writes output, and this path writes nothing until the very end.
2. **php-fpm will not stop it.** The deployment checklist deliberately requires
   `request_terminate_timeout >= 600` (see `docs/superpowers/specs/2026-05-10-issue-545-deployment-design.md`
   §11) for the long calendar routes, so the process manager's backstop is ten minutes away. And
   `max_execution_time` does not count time blocked in socket reads, so it does not apply either.
3. **The fan-out only grows.** It went 4 → 6 → 8 → 9 as object types were added
   (`rite_calendar_test` most recently); every new resource card multiplies the worst case again.

So a hung authorization server converts every admin-dashboard render into a 45-second worker, and a
handful of concurrent renders can exhaust the pool. This is an availability problem, not a latency
one — see #711, which measured the happy path at ~1.2 ms per call and was closed for that reason.

## Options

1. **Fan-out wall-clock budget** (preferred). Give `ResourceAdminService` a deadline for the whole
   resolve — say 2–3 s. `listObjectsIsolated()` checks the remaining budget before each call and,
   once it is spent, returns `[]` for the remaining types without dialing, logging the short-circuit.
   This reuses the existing per-type fail-closed semantics exactly: a skipped type is
   indistinguishable, to every caller, from a type that errored. Response shapes do not change.
2. **Lower the per-call timeouts on this path.** 5 s is generous for what production measures in
   single-digit milliseconds. 1 s read / 1 s connect would cap the fan-out at ~9 s on its own, and is
   a one-line change — but it is a blunt instrument shared with every other `OpenFgaClient` caller
   (`RoleCascadeService`, `PermissionAdminHandler`), so it probably wants to be env-configurable
   rather than hardcoded lower.
3. Both: configurable timeouts *and* a budget. The budget is what actually bounds the total; the
   timeout is what bounds a single stuck call.

Concurrency (the approach #711 evaluated) would also bound the total at ~5 s, but it costs a
Guzzle-specific transport path plus a PSR-18 fallback, both of which have to reproduce the
per-`(relation, type)` fail-closed classification exactly. A deadline gets the same bound for a
fraction of the surface area.

## Also worth a look while in here

- `ResourceAdminService::filterByAdminAccess()` fans out `check()` calls per access request
  (de-duplicated by resource, but unbounded in the number of distinct resources). Same hang exposure,
  same fix shape.
- `RoleCascadeService` loops `listObjects()` over relation × type on the role-revocation path.

## Acceptance

- A hung OpenFGA bounds `/auth/dashboard-scopes` (and the other two scope routes) to a small fixed
  wall-clock budget rather than `calls × timeout`.
- Response shapes and per-type fail-closed semantics are unchanged; existing handler tests stay green
  without modification.
- A test proves the bound — e.g. a Guzzle `MockHandler` with delayed responses, asserting the
  resolver returns within the budget and that the un-probed types come back empty rather than missing.

Split out of #711, which measured the happy-path latency and was closed as not warranted.

### LINKED ISSUE #879: Deduplicate the OpenFGA client plumbing and the scope-handler skeleton shared by five handlers
## The duplication

Five handlers carry a byte-identical copy of the same OpenFGA client plumbing:

- `src/Handlers/Auth/AdminScopesHandler.php`
- `src/Handlers/Auth/TestScopesHandler.php`
- `src/Handlers/Auth/DashboardScopesHandler.php`
- `src/Handlers/Admin/NotificationsHandler.php`
- `src/Handlers/Admin/AccessRequestAdminHandler.php`

```php
private ?OpenFgaClient $fgaClient = null;

private function isFgaClientAvailable(): bool
{
    return $this->fgaClient !== null || OpenFgaClient::isConfigured();
}

private function getFgaClient(): OpenFgaClient
{
    if ($this->fgaClient === null) {
        $this->fgaClient = OpenFgaClient::fromEnv();
    }
    return $this->fgaClient;
}
```

The three **scope** handlers go much further than that: `diff -u AdminScopesHandler.php
DashboardScopesHandler.php` is 30 lines, and every one of them is either the class name, the
docblock, or the last six lines of `handle()`. Lines 32–90 — constructor, the two FGA accessors,
preflight, `setAccessControlAllowOriginHeader()`, `validateRequestMethod()`,
`validateAcceptHeader(LAX)`, the `Content-Type` + `Cache-Control: no-store` pair, the `oidc_user`
attribute extraction, both `UnauthorizedException` branches, `OidcAuthMiddleware::isAdmin()`, and the
`isFgaClientAvailable()` guard around the resolve — are the same in all three files. They differ only
in which `ResourceAdminService` method they call and what key names they encode:

| handler                    | resolver               | payload keys                                          |
|----------------------------|------------------------|-------------------------------------------------------|
| `AdminScopesHandler`       | `resolveScopes()`      | `is_global_admin`, `is_resource_admin`, `admin_scopes` |
| `TestScopesHandler`        | `resolveTestScopes()`  | `is_global_admin`, `editor`, `admin`                   |
| `DashboardScopesHandler`   | `resolveScopes()` + `resolveViewerScopes()` | the above three plus `viewer_scopes` |

That's ~60 lines triplicated, and it is load-bearing: the fail-closed contract (empty scopes but
`is_global_admin` still honored from the token when OpenFGA is unavailable) is re-implemented
independently in each file, so a fix to one is a fix to one.

## Proposal

Two independent pieces — either is useful alone:

1. **A `Concerns` trait for the client plumbing**, alongside the existing
   `src/Handlers/Concerns/ResolvesOutboxTooling.php`: `ResolvesFgaClient`, holding the nullable
   property, the constructor-injection seam that the tests rely on, and the two accessors. All five
   handlers use it. Purely mechanical.
2. **A shared base for the three scope handlers.** `AbstractScopesHandler` owns the constructor
   (`GET`, `AcceptHeader::JSON`, `allowCredentials`) and a `final handle()` carrying the whole
   preflight → CORS → method → accept → `sub` extraction → `isGlobalAdmin` sequence, then delegates
   to one abstract method:

   ```php
   /** @return array<string, mixed> */
   abstract protected function buildScopePayload(string $sub, bool $isGlobalAdmin, ?ResourceAdminService $service): array;
   ```

   A `null` `$service` is the "OpenFGA not configured" case, which is exactly where each subclass
   already supplies its own fail-closed literal (`[]`, `['editor' => [], 'admin' => []]`,
   `array_fill_keys(VIEWER_OBJECT_TYPES, [])`). Each subclass shrinks to a docblock and about eight
   lines.

## Constraints

- **Response shapes must not change.** The three payloads are consumed by the frontend
  (`AuthHelper::fetchDashboardScopes()` and friends) and are covered by existing handler tests under
  `phpunit_tests/Handlers/` — those tests are the contract and should pass unmodified.
- Keep the `?OpenFgaClient $fgaClient = null` constructor seam: the handler tests inject a
  `MockHandler`-backed client through it.
- Fail-closed semantics per #793 stay exactly as they are; this is a code-shape change only.

## Acceptance

- No behavioral diff: existing `phpunit_tests/Handlers/*ScopesHandler*` tests pass untouched.
- `composer analyse` (PHPStan level 10) and `composer lint` clean.
- The `getFgaClient()` / `isFgaClientAvailable()` pair exists in one place.

Noted during #710's review and again while evaluating #711 (closed); it stands on its own and no
longer needs to be coordinated with either.

JohnRDOrazio:
Closes #878. Closes #879.

Two commits, in that order: the mechanical dedup first, then the budget on top of the deduped shape. Both fell out of the #711 evaluation — #711 measured the happy-path latency (~1.2 ms per lookup) and was closed as not-warranted, but the measuring turned up one real defect and one pile of duplication.

## 1. `refactor(handlers)`: one copy of the OpenFGA plumbing, one scope-handler skeleton

**Seven** classes carried a `getFgaClient()`, in two variants — one more than #879 estimated. Six were a byte-identical private pair (`isFgaClientAvailable()` + a memoizing `getFgaClient()`); the seventh, in `ResolvesOutboxTooling`, rebuilt the client — and therefore its Guzzle connection pool — on **every call**, and offered no injection seam. All of them now compose `Concerns\ResolvesFgaClient`, so the outbox path (`TestsHandler`, `RegionalDataHandler`) gains memoization as a side effect.

On top of that, the three scope endpoints shared ~60 lines of triplicated `handle()`: method and Accept validation, CORS, the `no-store` directive, extracting `sub` with both `UnauthorizedException` branches, reading the global-admin role, and gating on OpenFGA availability. `AbstractScopesHandler` now owns that in a `final handle()` and delegates to one hook:

```php
abstract protected function buildScopePayload(string $sub, bool $isGlobalAdmin, ?ResourceAdminService $service): array;
```

A `null` service is precisely the "OpenFGA unavailable" case, which is exactly where each subclass already kept its fail-closed literal. The three handlers keep their docblocks and shrink to that one method — 258 lines deleted, 43 added, plus the two new files.

No behavior change. Response shapes, key order, and #793's per-type fail-closed semantics are untouched, and **the existing handler tests pass unmodified** — they were the contract.

## 2. `fix(auth)`: a wall-clock budget on the fan-out

The actual defect, and it isn't latency.

Each lookup carries its own 5s read timeout, which bounds a single stuck call but not a sequence of them. The resolvers issue those in a loop — 4 / 8 / 9 calls for admin-, test- and dashboard-scopes — so an OpenFGA that **hangs** rather than refusing (a refused connection fails fast and is harmless) held a php-fpm worker for up to **45 seconds**. The caller was long gone: `AuthHelper::fetchDashboardScopes()` gives up at its own 5s timeout and renders the fail-closed dashboard. Nothing else would have stopped it either — the deployment checklist requires `request_terminate_timeout >= 600` for the long calendar routes, and `max_execution_time` does not count time blocked in socket reads. A handful of concurrent dashboard renders against a hung authorization server could exhaust the pool.

`ResourceAdminService` now stamps a deadline when a fan-out opens and refuses to dial past it. **A refused lookup returns exactly what an errored one returns**, so the budget reuses #793's fail-closed contract instead of inventing a second one — every `viewer_scopes` key still present, scopes gathered before the trip preserved, response shapes identical. `filterByAdminAccess()` is covered too (its per-request `check()` sweep has the same exposure), and a nested resolver reuses the outer window rather than refreshing it, so the budget bounds the whole sequence rather than each leg.

**Be precise about what this bounds:** the budget gates whether the *next* lookup is dialed; it cannot interrupt one already in flight. The true worst case is the budget plus one read timeout (~8s), not 3s. The point is that it no longer *grows with the object-type list*, which went 4 → 6 → 8 → 9 as resource cards were added. Squeezing the remainder means lowering `OpenFgaClient`'s per-call timeouts — shared with every other caller, and deliberately left alone here.

3s is ~250x the measured cost of the largest fan-out (~12 ms for nine calls), so it cannot be reached by a merely busy server. It is a constant with a constructor override rather than an env var: no new deployment surface to document or validate, and the tests drive it alongside an injected monotonic clock, so nothing sleeps.

## Testing

Six new tests in `ResourceAdminServiceTest`, written first and watched fail. Because the initial red was a constructor error rather than a behavioral failure, each guard was then **mutated out** individually to confirm the assertions actually catch it: disabling the `listObjects` guard fails 4 of them, disabling the `check` guard fails the 5th, and the 6th is a guard against a budget so tight it would skip healthy calls.

- 2117 tests green across `Handlers/`, `Services/`, `Http/`, `Enum/`, `Models/`, `Params/`, `Methods/` and `RouterTest`
- The 6 skips are pre-existing and environmental (an OpenFGA store this host does not run) — verified identical on the branch point, 530 tests / 6 skipped before, 536 / 6 after
- `composer analyse` (PHPStan level 10) and `composer lint` clean
- `Routes/*` deliberately not run: they hit `localhost:8000`, which serves a different checkout, so a green there would say nothing about this branch

## Deliberately out of scope

- `RoleCascadeService`'s relation × type loops — a background revocation path, not a worker-render one
- `AccessRequestAdminHandler`'s own `check()` loop — bounded by one request's permission count (typically 1–3) rather than a growing type list, and it throws `ForbiddenException` rather than failing closed silently
- Lowering `OpenFgaClient::fromEnv()`'s 5s/2s timeouts — shared with every caller; worth its own change if the residual ~8s ever matters

## Summary by CodeRabbit

* **New Features**
  * Added consistent handling for authentication scope requests, including validation, authentication, CORS, and cache controls.
  * Scope responses now fail safely with empty permissions when authorisation services are unavailable.

* **Bug Fixes**
  * Added safeguards to stop lengthy permission checks once the processing time limit is reached.
  * The time limit now applies across related permission operations.
  * Previously resolved permissions and response fields are preserved when processing stops early.
  * Access checks fail closed and record a single summary when requests are skipped.

### DIFF
--- a/phpunit_tests/Services/ResourceAdminServiceTest.php
+++ b/phpunit_tests/Services/ResourceAdminServiceTest.php
@@ -463,4 +463,180 @@ final class ResourceAdminServiceTest extends TestCase
 
         self::assertCount(count(ResourceAdminService::ADMIN_OBJECT_TYPES), $service->resolveScopes('cei-admin'));
     }
+
+    /**
+     * The MockHandler behind the most recent {@see serviceWithClock()} service.
+     * Its remaining count is how a test proves a call was never dialed.
+     */
+    private ?MockHandler $mockQueue = null;
+
+    /**
+     * A service whose fan-out clock the test drives by hand.
+     *
+     * The clock reads a by-reference float rather than the wall clock, and the
+     * queued responses are callables that advance it — so "this OpenFGA call took
+     * two seconds" is expressed deterministically, with no sleeping.
+     *
+     * @param array<int, GuzzleResponse|callable> $responses Queued, replayed in order.
+     * @param float $now Advanced by the queued callables; passed by reference so the clock closure sees it move.
+     */
+    private function serviceWithClock(array $responses, float &$now, float $budget, ?LoggerInterface $logger = null): ResourceAdminService
+    {
+        $mock   = new MockHandler($responses);
+        $guzzle = new GuzzleClient(['handler' => HandlerStack::create($mock)]);
+        $psr17  = new Psr17Factory();
+        $client = new OpenFgaClient(
+            apiUrl: 'http://openfga.test',
+            storeId: 'test-store',
+            modelId: 'test-model',
+            httpClient: $guzzle,
+            requestFactory: $psr17,
+            streamFactory: $psr17,
+            apiToken: 'test-token'
+        );
+
+        $this->mockQueue = $mock;
+
+        return new ResourceAdminService(
+            $client,
+            $logger ?? new CollectingLogger(),
+            budgetSeconds: $budget,
+            clock: static function () use (&$now): float {
+                return $now;
+            }
+        );
+    }
+
+    /**
+     * A queued response that costs $seconds of fan-out budget.
+     */
+    private static function costing(float &$now, float $seconds, GuzzleResponse $response): callable
+    {
+        return static function () use (&$now, $seconds, $response): GuzzleResponse {
+            $now += $seconds;
+            return $response;
+        };
+    }
+
+    public function testResolveViewerScopesStopsDialingOnceTheFanOutBudgetIsSpent(): void
+    {
+        $now = 0.0;
+        // Two calls at 2s each exhaust a 3s budget: the first is dialed at t=0, the
+        // second at t=2 (still inside), and by t=4 the budget is gone.
+        $service = $this->serviceWithClock([
+            self::costing($now, 2.0, new GuzzleResponse(200, [], '{"objects":["general_roman_calendar:temporale"]}')),
+            self::costing($now, 2.0, new GuzzleResponse(200, [], '{"objects":["national_calendar_test:IT"]}')),
+            new GuzzleResponse(200, [], '{"objects":["diocesan_calendar_test:romamo_it"]}'),
+            new GuzzleResponse(200, [], '{"objects":["general_roman_calendar_test:temporale"]}'),
+            new GuzzleResponse(200, [], '{"objects":["rite_calendar_test:ambrosian"]}'),
+        ], $now, 3.0);
+
+        $scopes = $service->resolveViewerScopes('cei-admin');
+
+        // Every key is still present — the dashboard distinguishes "empty" from "missing".
+        self::assertSame(ResourceAdminService::VIEWER_OBJECT_TYPES, array_keys($scopes));
+        self::assertSame(['temporale'], $scopes['general_roman_calendar']);
+        self::assertSame(['IT'], $scopes['national_calendar_test']);
+        self::assertSame([], $scopes['diocesan_calendar_test']);
+        self::assertSame([], $scopes['general_roman_calendar_test']);
+        self::assertSame([], $scopes['rite_calendar_test']);
+
+        // The point of the budget: the remaining three were never dialed at all.
+        self::assertCount(3, $this->mockQueue);
+    }
+
+    public function testResolveScopesKeepsWhatItResolvedBeforeTheBudgetWasSpent(): void
+    {
+        $now     = 0.0;
+        $service = $this->serviceWithClock([
+            self::costing($now, 5.0, new GuzzleResponse(200, [], '{"objects":["national_calendar:IT"]}')),
+            new GuzzleResponse(200, [], '{"objects":["diocesan_calendar:romamo_it"]}'),
+            new GuzzleResponse(200, [], '{"objects":["wider_region:Europe"]}'),
+            new GuzzleResponse(200, [], '{"objects":["general_roman_calendar:temporale"]}'),
+        ], $now, 3.0);
+
+        self::assertSame(
+            [['object_type' => 'national_calendar', 'object_id' => 'IT']],
+            $service->resolveScopes('cei-admin')
+        );
+        self::assertCount(3, $this->mockQueue);
+    }
+
+    public function testResolveTestScopesBudgetSpansBothRelationLoops(): void
+    {
+        $now = 0.0;
+        // One 5s editor probe spends the whole budget, so neither the remaining
+        // editor types nor any admin type is dialed.
+        $service = $this->serviceWithClock(
+            array_merge(
+                [self::costing($now, 5.0, new GuzzleResponse(200, [], '{"objects":["national_calendar_test:IT"]}'))],
+                array_map(
+                    static fn(): GuzzleResponse => new GuzzleResponse(200, [], '{"objects":["national_calendar_test:XX"]}'),
+                    range(1, 7)
+                )
+            ),
+            $now,
+            3.0
+        );
+
+        self::assertSame(
+            [
+                'editor' => [['object_type' => 'national_calendar_test', 'object_id' => 'IT']],
+                'admin'  => [],
+            ],
+            $service->resolveTestScopes('cei-admin')
+        );
+        self::assertCount(7, $this->mockQueue);
+    }
+
+    public function testFilterByAdminAccessFailsClosedOnceTheBudgetIsSpent(): void
+    {
+        $now      = 0.0;
+        $service  = $this->serviceWithClock([
+            self::costing($now, 5.0, new GuzzleResponse(200, [], '{"allowed":true}')),
+            new GuzzleResponse(200, [], '{"allowed":true}'),
+        ], $now, 3.0);
+        $requests = [
+            ['id' => 'A', 'permissions' => [['object_type' => 'national_calendar', 'object_id' => 'IT', 'relation' => 'admin']]],
+            ['id' => 'B', 'permissions' => [['object_type' => 'national_calendar', 'object_id' => 'US', 'relation' => 'admin']]],
+        ];
+
+        // A was checked inside the budget and survives; B is excluded without a call.
+        $kept = $service->filterByAdminAccess($requests, 'cei-admin');
+
+        self::assertSame(['A'], array_column($kept, 'id'));
+        self::assertCount(1, $this->mockQueue);
+    }
+
+    public function testFanOutBudgetExhaustionIsLoggedOnceWithTheSkippedCount(): void
+    {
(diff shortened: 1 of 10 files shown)


JohnRDOrazio:
CodeRabbit's Moderate merge-risk note was correct, and it found a hole worth spelling out — fixed in cf95fb05.

**The bug:** I armed the budget per *fan-out*, but two endpoints run more than one fan-out per request through a single `ResourceAdminService`:

- `GET /auth/dashboard-scopes` → `resolveScopes()` then `resolveViewerScopes()`
- `GET /admin/notifications` → `resolveScopes()` then `filterByAdminAccess()`

Each got its own fresh 3s window. So against a hung OpenFGA the dashboard's worst case was two budgets plus two read timeouts (~16s), not the ~8s the PR body claimed — on precisely the endpoint #878 was filed about. The nesting support I had written only covered a resolver reached *through* another, which is not what these are: they're sequential siblings.

**The fix:** the deadline is stamped on the first lookup an instance makes and never refreshed, so it spans every fan-out that instance performs. Every call site already constructs a service per request, which makes it a per-request bound — and that constraint is now documented on the constant ("do not hold an instance across requests"). The owner-flag plumbing is gone: `withinBudget()` stamps lazily, and `reportSkipped()` no longer has to decide who closes the window, so the change is a net simplification.

Two new tests, one per endpoint shape, both watched failing first — before the fix they showed the second sweep dialing every one of its lookups on a budget that was already spent.

**Corrected numbers for the PR description above:** worst case is the budget plus one read timeout (~8s) *per request*, down from ~45s for dashboard-scopes — and it no longer grows with either the object-type list or the number of resolver calls an endpoint makes.

538 tests green across `Services/`, `Handlers/Auth/` and `Handlers/Admin/`; PHPStan level 10 and phpcs clean.

PR #889, opened 2026-08-25
fix(setup-zitadel): read the ports from .env instead of guessing them

JohnRDOrazio:
Follow-up to #888, closing the last place where the port could silently disagree with itself.

## The bug

`scripts/setup-zitadel.sh` writes `ZITADEL_ISSUER` into every `.env` it touches and registers the OIDC redirect URIs for both apps — but it resolved `ZITADEL_PORT`, `FRONTEND_PORT` and `TESTS_PORT` from its own shell environment only, never from `.env`, which is where `docker compose` reads them.

So on a stack running `ZITADEL_PORT=8090`:

```bash
./scripts/setup-zitadel.sh --update-env      # writes ZITADEL_ISSUER=http://localhost:8080
```

It wrote the wrong issuer straight back over the working value, silently undoing the configuration it was asked to update. #888 fixed the derivation but left the input hardcoded to the shell.

`FRONTEND_PORT` and `TESTS_PORT` are the same bug with a sharper edge — they build the redirect URIs registered with the **LiturgicalCalendar Frontend** and **LiturgicalCalendar Tests** OIDC apps (`create_oidc_app` call sites). A redirect URI on the wrong port fails login outright rather than degrading.

## The fix

An `env_file_value` helper reads a single `KEY=value` from `${PROJECT_DIR}/.env`. All three ports now fall back through **exported variable → `.env` → built-in default**, which is compose's own precedence.

`.env` is parsed with `grep`/`cut`, never `source`d: it holds secrets, and arbitrary shell in a config file must not be executed. The helper always exits 0 so `set -e` can't trip on a missing file or absent key.

The config block moved below the `SCRIPT_DIR`/`PROJECT_DIR` definitions, since it now depends on them. `ZITADEL_URL` isn't *used* until line ~111, so nothing observes the reorder.

## On matching compose exactly

I checked the accepted `.env` spellings against compose rather than assuming them. Compose reads all four of these as the same value:

```
KEY=v      KEY="v"      '  KEY=v'      'KEY = "v"'
```

My first cut of the helper matched the first three and **missed `KEY = "v"`** — which would have reintroduced precisely the disagreement this change exists to prevent. The pattern now tolerates whitespace around `=` and wrapping quotes of either kind.

## Verification

Seven scenarios against the real code path (including `PROJECT_DIR` resolution, exercised the way the Frontend's wrapper extracts this script):

| Case | Result |
|---------------------------------|-----------------------------|
| no `.env` at all | `:8080` / 3000 / 3003 |
| `.env` sets `ZITADEL_PORT=8090` | `http://localhost:8090` |
| exported var vs `.env` | exported wins (`:9999`) |
| `KEY = "8095"` (compose-style) | `http://localhost:8095` |
| commented-out key | ignored; siblings still read |
| empty `.env` | defaults |
| single-quoted value | `TESTS_PORT=3009` |

And against the live stack: with the real `ZITADEL_PORT=8090` in the Frontend's `.env`, it resolves `ZITADEL_URL=http://localhost:8090`. `bash -n` clean.

## Incidental finding

This script already creates a **second OIDC app, "LiturgicalCalendar Tests"**, with redirect URI `http://localhost:${TESTS_PORT}/auth/callback.php` — and it is present in the running instance (`projections.apps7` lists it alongside "LiturgicalCalendar Frontend"). That client registration was assumed to be a manual prerequisite for the planned UnitTestInterface OIDC migration; it is not.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Zitadel configuration parsing for common `KEY=value` and `KEY:value` formats.
  - Correctly handles whitespace, quoted values, inline comments, and comment-only values.
  - Ensured environment-file contents are read safely without executing shell code.
  - Improved reliability when settings are missing or incomplete.
  - Preserves explicit Zitadel URLs and warns when their port differs from the configured port.

### DIFF
--- a/scripts/setup-zitadel.sh
+++ b/scripts/setup-zitadel.sh
@@ -44,56 +44,76 @@ ROLES=("admin" "developer" "calendar_editor" "test_editor")
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_DIR="${SCRIPT_DIR}/.."
 
-# Read a single KEY=value out of the project's .env WITHOUT sourcing it.
+# Resolve a single configuration key the way `docker compose` would.
 #
-# `docker compose` reads .env for its own interpolation, so the ports declared there
-# are the stack's source of truth. This script has to agree with them: it writes
-# ZITADEL_ISSUER into .env files and registers OIDC redirect URIs in Zitadel, and a
-# port it guessed differently from compose produces exactly the silent disagreement
-# those values exist to prevent. Before this, `./scripts/setup-zitadel.sh --update-env`
-# run from a bare shell would happily overwrite a ZITADEL_PORT=8090 stack's issuer
-# with http://localhost:8080.
+# compose reads .env for its own interpolation, so the ports declared there are the
+# stack's source of truth. This script has to agree with them: it writes ZITADEL_ISSUER
+# into .env files and registers OIDC redirect URIs, and a port it resolved differently
+# from compose produces exactly the silent disagreement those values exist to prevent.
 #
-# Precedence matches compose: an exported shell variable beats .env, which beats the
-# built-in default. The accepted spellings were derived by testing compose rather than
-# assumed, because a parser that is merely close reintroduces the disagreement this
-# helper exists to remove. Compose reads all of these as the same value:
-#     KEY=v      KEY="v"      KEY = "v"      '  KEY=v'
-#     KEY: v                    (colon delimiter)
-#     KEY=v # comment           (comment needs WHITESPACE before the '#')
-# but `KEY=v# x` keeps the literal `v# x`, so the '#' is only a comment when a space
-# precedes it. Quoted values take the quoted span and ignore any trailing text.
+# PRIMARY PATH: ask compose itself, via `docker compose config --environment`. That is
+# authoritative by construction — it cannot disagree with compose — and costs ~40ms.
+# It matters because compose's .env handling is richer than it looks: it accepts `:` as
+# well as `=`, strips a comment only when whitespace precedes the '#', and interpolates
+# values, so `ZITADEL_PORT=${LOCAL_PORT:-8080}` resolves rather than arriving literally.
+# Three successive hand-written approximations of that parser each missed a case.
 #
-# Parsed with grep/cut rather than `source`, because .env holds secrets and arbitrary
-# shell in a config file must never be executed. Always exits 0 so `set -e` cannot
-# trip on a missing file or an absent key.
+# FALLBACK: a conservative literal reader, for when docker is absent or PROJECT_DIR has
+# no valid compose file. It handles the common spellings but deliberately does NOT try to
+# interpolate; a value it cannot resolve is reported rather than guessed at.
+#
+# Precedence in both paths matches compose: exported shell variable, then .env, then the
+# built-in default. .env is never `source`d — it holds secrets, and arbitrary shell in a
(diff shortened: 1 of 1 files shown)


JohnRDOrazio:
Both findings were correct, and both are fixed in `90ae8468`. I verified them against compose rather than reasoning about them — and the verification is what showed the parser was genuinely wrong, not merely stricter.

| `.env` line | compose reads | old parser | now |
|---|---|---|---|
| `ZITADEL_PORT: 8090` | `8090` | *missed → fell back to 8080* | `8090` |
| `ZITADEL_PORT=8090 # comment` | `8090` | `8090 # comment` → invalid URL | `8090` |
| `ZITADEL_PORT=8090# nospace` | `8090# nospace` | `8090# nospace` | `8090# nospace` |

That third row is why the fix isn't "strip everything after `#`" — a `#` is only a comment when whitespace precedes it, and compose keeps the literal otherwise. Quoted values now take the quoted span and ignore trailing text, matching compose on `KEY="v" # c`. There's a differential harness that pushes all ten spellings through both compose and the parser and asserts they agree; it passes.

On the second finding, I kept the precedence as designed — an explicit `ZITADEL_URL` still wins, because a deployment on a real domain needs it and the port is meaningless there — but took the "validate intentional overrides" branch of your suggestion. A `ZITADEL_URL` naming a *different* port than `ZITADEL_PORT` now emits a warning on stderr instead of silently choosing, since that combination would health-check, call, and write an issuer for a port compose isn't serving. A portless URL such as `https://auth.example.org` warns about nothing.

Verified against the live stack too: with the real `ZITADEL_PORT=8090` in the Frontend's `.env`, the script resolves `http://localhost:8090`.

JohnRDOrazio:
Fixed in `2b4ea45f` — and this finding is the one that showed the *approach* was wrong, not just incomplete. You're right: compose interpolates values inside `.env`, so `ZITADEL_PORT=${LOCAL_PORT:-8080}` resolved for compose while my parser returned the literal.

That was the third case in a row where a hand-written approximation of compose's `.env` handling disagreed with compose — colon delimiter, then whitespace-delimited trailing comments, now interpolation. Each fix made the parser *closer* without making it *equivalent*, and "closer" still produces exactly the silent disagreement this change exists to remove.

So I stopped approximating and asked compose:

```bash
docker compose config --environment
```

Authoritative by construction — it cannot disagree with compose — and it costs ~40ms, cached so one call serves all three ports. The real stack resolves end to end in 185ms.

The literal parser survives only as a fallback for when docker is absent or `PROJECT_DIR` has no valid compose file. It deliberately does **not** attempt interpolation: a value it can't resolve now warns, naming the key and the file, and falls back to the built-in default rather than guessing. Reporting beats a plausible wrong answer.

Differential harness, ten `.env` spellings through both compose and the script, asserting agreement:

| `.env` line | compose | script |
|---|---|---|
| `ZITADEL_PORT=8090` | `8090` | `8090` |
| `ZITADEL_PORT: 8090` | `8090` | `8090` |
| `ZITADEL_PORT = "8090"` | `8090` | `8090` |
| `ZITADEL_PORT=8090 # comment` | `8090` | `8090` |
| `ZITADEL_PORT=8090# nospace` | `8090# nospace` | `8090# nospace` |
| `ZITADEL_PORT=${BASE}` | `8090` | `8090` |
| `ZITADEL_PORT=${UNSET_X:-8077}` | `8077` | `8077` |

Plus four fallback-path cases with the compose file removed. All green; CI green.

Thanks for pushing on this one — the first two fixes were treating symptoms.

PR #891, opened 2026-08-25
fix(ws): raise the handshake header limit so a logged-in user can connect

JohnRDOrazio:
Fixes a live-server symptom on `litcal-tests.johnromanodorazio.com`: **every WebSocket connection fails while the user is logged in at `litcal-staging`, and starts working the instant they log out.** The browser reports only `WebSocket connection failed`.

## Root cause

Ratchet caps the handshake request at `HttpRequestParser::$maxSize` — **4096 bytes** — and answers `413 Request Entity Too Large` above it. That is ample for a WebSocket handshake, until the browser attaches cookies.

The WS host shares a registrable domain with the sites that authenticate against Zitadel, so a `COOKIE_DOMAIN`-scoped session is sent here on **every handshake**, despite this server never reading a cookie. Zitadel's tokens are JWTs — measured on a live session, `litcal_access_token` (1082 B) and `litcal_id_token` (1586 B) alone are ~2.7 KB of `Cookie:` header, before a refresh token or anything else the domain carries. Add a browser's ordinary handshake headers and 4096 is reachable.

## Measured, not inferred

A raw handshake against this server, varying only the cookie size. **Before:**

```
~3820 B  ->  HTTP/1.1 101 Switching Protocols
~4020 B  ->  HTTP/1.1 101 Switching Protocols
~4220 B  ->  HTTP/1.1 413 Request Entity Too Large
```

**After:**

```
~4420 B  ->  HTTP/1.1 101 Switching Protocols
~16020 B ->  HTTP/1.1 101 Switching Protocols
~16420 B ->  HTTP/1.1 413 Request Entity Too Large
```

## The fix

`LargeHeaderHttpServer`, a small `HttpServer` subclass, sets the parser's limit to 16 KB. Ratchet exposes no setter and no constructor argument, and `$_reqParser` is `protected`, so a subclass is the supported way to reach it.

The cap is a denial-of-service guard, so it is **raised, not removed** — 16 KB clears a full Zitadel session plus ordinary headers, and still bounds what an unauthenticated client can make this server allocate.

## Why not stop sending the cookies instead

A host cannot opt out of a parent-domain cookie, and narrowing `COOKIE_DOMAIN` would break the cross-subdomain sharing it exists to provide. The waste is real — several kilobytes per handshake that this server discards — but it is not addressable from this side.

## Note on deployment

This needs the WebSocket server restarted to take effect (`systemctl restart` on the unit that runs `bin/LitCalTestServer.php`), since it is a long-running process rather than a per-request one.

## Summary by CodeRabbit

* **New Features**
  * WebSocket connections now support larger handshake headers, accommodating substantial session cookies.
  * Added configurable handshake size limits, with a higher default limit to reduce connection failures.

* **Bug Fixes**
  * Prevented oversized handshakes from being rejected prematurely while retaining explicit maximum-size enforcement.
  * Server entrypoint now rejects non-command-line execution requests safely.

* **Tests**
  * Added coverage for accepted, rejected, and custom handshake-size limits.

### DIFF
--- a/bin/LitCalTestServer.php
+++ b/bin/LitCalTestServer.php
@@ -1,6 +1,5 @@
 <?php
 
-
 // Refuse any entry that is not the CLI. This file is the Ratchet WebSocket server — systemd runs it
 // as `php bin/LitCalTestServer.php` — and it ends in `IoServer::factory(...)->run()`, i.e. it
 // binds WS_PORT and enters an event loop.
@@ -58,7 +57,7 @@ if (null === $autoloaderPath) {
 require_once $autoloaderPath;
 
 use Ratchet\Server\IoServer;
-use Ratchet\Http\HttpServer;
+use LiturgicalCalendar\Api\Http\Server\LargeHeaderHttpServer;
 use Ratchet\WebSocket\WsServer;
 use LiturgicalCalendar\Api\Health;
 use Dotenv\Dotenv;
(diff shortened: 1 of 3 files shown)


JohnRDOrazio:
@coderabbitai review

Issue #904, opened 2026-08-28
Define officially supported locales as a curated resource, and stop 500ing on unofficial ones


JohnRDOrazio:
## The problem

There is no usable definition of "officially supported locale".

`CalendarMetadataProvider::FULLY_TRANSLATED_LOCALES` (`src/Services/CalendarMetadataProvider.php:56`)
holds `['en', 'fr', 'it', 'nl', 'la']`, but it is **`private` and referenced exactly once** — in
`buildLocales()`, to filter the `/calendars` discovery response. Nothing else in `src/` or
`phpunit_tests/` consults it. It describes the contract without enforcing it.

Meanwhile `LitLocale::getSupportedLocales()` returns `self::$values + self::$AllAvailableLocales`,
where the latter is `\ResourceBundle::getLocales('')` — roughly 700 locales, filtered only for POSIX.
So at the validation layer, "supported" effectively means "a locale that exists somewhere".

Three different sets are conflated, and only the loosest is expressed in code:

| Set                              | Defined by            | Members                                                          |
| -------------------------------- | --------------------- | ---------------------------------------------------------------- |
| Locales the API *accepts*        | `LitLocale` → ICU     | ~700, uncurated                                                    |
| Locales with event **names**     | `i18n/*`              | de, es, fr, hr, hu, id, it, la, nl, pl, pt, pt_BR, sk, vi          |
| Locales with **lectionary** data | files on disk         | en, fr, hr, it, la, nl                                             |
| Locales declared **official**    | the private const     | en, fr, it, la, nl                                                 |

## Observed failure

`https://litcal.johnromanodorazio.com/api/dev/calendar?locale=hr` currently returns **500**.

Croatian is not an officially supported locale, so incomplete data for it should degrade, not error.
Instead `ReadingsMap::getReadings()` throws when an offset is missing — even though **its own
docblock says it returns null**:

> *If the offset does not exist, it returns null.*

The return type does not include `null` either. The documented contract is graceful degradation; the
implementation never matched it. This is what took down the Croatian calendar when a decree was
created without full lectionary coverage (see #903).

## Decided design

**1. `jsondata/supportedLocales.json`** — a curated resource, sibling to the existing top-level
reference data (`likelySubtags.json`, `world_dioceses.json`), replacing the private const as the
single source of truth for hosted and self-hosted deployments alike.

**2. `LocaleReadinessChecker`** — answers "does this locale have every resource an official locale
[...]

Issue #902, opened 2026-08-28
Edits made through the API never reach the repository, and are destroyed by the next deploy


JohnRDOrazio:
## What happened

After the recent decree-interface PRs merged, I used the new editor to create the decree inscribing
**Saint John Henry Newman, Priest and Doctor of the Church**, into the General Roman Calendar
(`StJohnNewman_Create`, Prot. N. 760/25, 9 November 2025, optional memorial on 9 October).

The API wrote it correctly to disk. But the deployment has no git working tree, so there is no way to
commit it and push it back to GitHub — and the next deploy will delete it.

This is not specific to decrees. **Every write path in the API has the same problem.**

## Current state

| Deployment           | Decrees | `StJohnNewman_Create` |
| -------------------- | ------: | --------------------- |
| Repo (`development`) |      14 | absent                |
| `api/dev` (staging)  |      15 | **present**           |
| `api/v5` (production)|      14 | absent                |

The decree exists only on the staging deployment's disk. `api/dev` is redeployed on **every push to
`development`**, so it is one merge away from being lost.

## Why this happens

Two independent causes:

1. **Deploys overwrite source data.** `.github/workflows/deploy.yaml` rsyncs `--archive --delete`
   from a fresh runner checkout. `jsondata/` is on the allowlist in
   `.github/deploy/check-deploy-payload.sh` and appears nowhere in
   `.github/deploy/rsync-exclude.txt`, so every deploy reverts server-side edits — and `--delete`
   removes files the server created.
2. **There is no working tree to commit from.** `.git/` is excluded from the deploy payload and the
   deploy user is chrooted, by design — see the deploy-hygiene rationale in
   `check-deploy-payload.sh`. The old "edit on the server, commit from the server" path no longer
   exists.

Affected write paths:

- `RegionalDataHandler` — 7 `file_put_contents()`, 3 `unlink()` (national, diocesan and wider-region
  calendars, plus their i18n files)
- `DecreesHandler` — 4 `file_put_contents()` (`decrees.json`, i18n, lectionary readings)
- `TestsHandler` — 1 `file_put_contents()`, 1 `unlink()`

## Immediate mitigation

The Newman decree must be recovered from the staging host before the next merge to `development`.
The affected source files are:

```text
jsondata/sourcedata/rite/roman/decrees/decrees.json
jsondata/sourcedata/rite/roman/decrees/i18n/{locale}.json
[...]

PR #906, opened 2026-08-28
feat(locales): locale readiness checker and admin curation surface

JohnRDOrazio:
Refs #904. Follows #905, which made the supported-locale list a curated resource.

## What this adds

**`LocaleReadinessChecker`** — answers "does this locale have everything an official locale must
have?" with five probes:

| Probe | Requires |
| ------------------- | ----------------------------------------------------------- |
| `gettext_catalogue` | `i18n/{locale}/` (skipped for `en`, the source language) |
| `lectionary_corpora`| all ten General Roman corpora |
| `decree_readings`   | a readings entry for every **newly created** decreed event |
| `decree_names`      | a non-empty name for every named decreed event |
| `universal_missals` | the four universal missals translated |

Promotion is not a label change — it flips `ReadingsMap` from degrading quietly to throwing. So this
is the pre-flight that has to pass first, and the guard that keeps already-official locales complete.

**`GET /admin/locales`** and **`GET /admin/locales/{locale}`** — the curation surface, restricted to
global Zitadel admins.

**CI guard** — `testEveryOfficialLocaleIsReady` fails the build if an official locale's data
regresses, with the missing identifiers in the failure message.

## Two bugs the checker found in itself

Running it against real data before shipping caught both:

1. **Only `createNew` decrees need decree readings.** `setProperty` (raising a grade) and
   `makeDoctor` (adding a title) leave the event in the sanctorale, where its readings already live.
   The first draft required entries for all of them and so reported **all five official locales as
   incomplete** for data that is perfectly correct.
2. **Which decrees supply a *name* is not derivable from the action.** `makeDoctor` needs one;
   `setProperty` raising a grade does not. The required set is now the union of keys across every
   locale's i18n file, which asserts the invariant that actually matters: whatever any locale names,
   every official locale must name too.

## Result

The checker independently reproduces the conclusion reached by hand on #904:

```text
en  official:yes ready:YES  all checks passed
fr  official:yes ready:YES  all checks passed
it  official:yes ready:YES  all checks passed
la  official:yes ready:YES  all checks passed
nl  official:yes ready:YES  all checks passed
hr  official:no  ready:no   decree_names -> StJohnNewman
es  official:no  ready:no   missing 10 of 10 lectionary corpora
de  official:no  ready:no   corpora, decrees lectionary, 12 names
```

**Croatian is one Croatian name away from being promotable.** Supply a name for `StJohnNewman` and
`hr` passes every probe.

## Read-only, and the response says why

`curation.writable` is `false` and carries its reason. Promotion writes
`jsondata/supportedLocales.json` on the deployed server, which the next deploy would overwrite
(#902), so an operator seeing a green "ready" is told why there is no promote button rather than left
to wonder. Promotion stays a reviewed pull request until change requests can carry that write — the
right audit trail for a governance decision regardless.

## Verification

- 22 new tests, 48 assertions; 53 green across the locale suites, the handler and the router.
- `composer analyse` — PHPStan level 10, no errors.
- `/admin/locales` returns 401 unauthenticated, 403 for a non-admin, 404 for a locale with no
  resources on disk.

A companion frontend PR adds the admin view over these endpoints.

## Summary by CodeRabbit

* **New Features**
  * Added read-only admin endpoints to list known locales and view detailed readiness reports.
  * Reports show official status, readiness, failed checks, advisory findings and missing requirements.
  * Added checks for catalogues, lectionary content, decree readings, decree names and missal translations.
  * Added guidance explaining that locale promotion is managed through the supported-locales pull request process.

* **Tests**
  * Added comprehensive coverage for access control, locale reporting, readiness checks, advisory results, JSON output and missing locales.

### DIFF
--- a/jsondata/schemas/openapi.json
+++ b/jsondata/schemas/openapi.json
@@ -9910,7 +9910,8 @@
                         "exists",
                         "parses",
                         "validates"
-                      ]
+                      ],
+                      "expected_locales": null
                     },
                     {
                       "id": "sanctorale:roman:US_2011:i18n",
@@ -9923,7 +9924,8 @@
                         "exists",
                         "parses",
                         "validates"
-                      ]
+                      ],
+                      "expected_locales": null
                     }
                   ]
                 }
--- a/jsondata/schemas/openapi.json
+++ b/jsondata/schemas/openapi.json
@@ -101,6 +101,10 @@
       "description": "Curation surface for the locales the API declares officially supported. An official locale promises complete data: missing data for one is an error, while a locale outside the list degrades quietly."
     },
     {
+      "name": "Admin - Outbox",
+      "description": "Inspection and retry management for the OpenFGA reconciliation outbox. Global admins only."
+    },
+    {
       "name": "Applications",
       "description": "Developer application management (create, update, delete applications and API keys)"
     },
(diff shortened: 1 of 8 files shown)


JohnRDOrazio:
## Review follow-ups

**1. `knownLocales()` scanned only `sanctorum`** — valid, fixed. It now scans every one of the ten
corpora plus the decrees lectionary.

I did **not** extend it to the missals, though, because doing so changes the answer for the worse:

```text
current  : de en es fr hr hu id it la nl pl pt pt_BR sk vi   (15)
+ missals: … plus en_US, it_IT                               (17)
```

`en_US` and `it_IT` exist only as translations of the **national** editions
(`propriumdesanctis_US_2011`, `propriumdesanctis_IT_1983`). Surfacing them would offer them as
candidates for *General Roman* promotion, which they are not. Universal-missal coverage is still
probed per locale by `checkUniversalMissals()`; it just does not define the candidate set. Comment in
the code records this.

**2. `AbstractHandlerTestCase`** — adopted. Verified first that it does not skip: it gates on
`JWT_SECRET`, which is present, and `$requiresDatabase` defaults false. The test now uses
`requestFor()` and `decodeJsonBody()` like the other 58 handler tests.

## A finding worth review attention

Following a suggestion to require non-empty values rather than mere key presence, I checked what that
would actually do. It surfaced a real content gap:

| Locale | Official | Decreed events with a **blank** readings entry |
| ------ | -------- | --------------------------------------------- |
| en     | yes      | 1 of 11                                        |
| la     | yes      | **9 of 11**                                    |
| fr     | yes      | **10 of 11**                                   |
| it     | yes      | **10 of 11**                                   |
| nl     | yes      | **10 of 11**                                   |

These events appear in **no** `sanctorum` file in any locale, so the decrees lectionary is their only
source of readings — a blank entry means the calendar serves none. This is not a structural nuance;
those calendars genuinely have no readings for most decreed celebrations.

So gating on non-empty readings would mark **four of the five locales the API already advertises as
officially supported** as not ready, and turn the CI guard red immediately.

I have therefore added `decree_readings_populated` as an **advisory** check: it measures and reports
the gap, is surfaced in the admin interface with a neutral badge, and does not influence `ready()`.
Promoting it to gating is dropping one argument, once the content exists.

Names are a different matter and *are* gated: all five official locales already have complete
decreed-event names, so requiring non-empty names costs nothing and catches the next regression —
which is precisely what blocks Croatian today.

Whether to close the readings gap or soften the standard is a content decision, not one I should make
inside a review fix.

JohnRDOrazio:
## Added: OpenAPI coverage

Two follow-ups since the last review, both in `jsondata/schemas/openapi.json`.

**1. The new endpoints are documented** — `/admin/locales` and `/admin/locales/{locale}`, with a tag
and three schema components. The `advisory` flag is called out explicitly, since a consumer needs to
know an unpassed advisory probe does not make a locale unready.

**2. `/admin/outbox` is documented too.** While adding the above I noticed it was the one admin route
absent from the description, so a generated client had no way to reach it. Both operations are now
described, read from `OutboxAdminHandler` rather than inferred:

- `GET /admin/outbox` — row listing, or per-status counts plus oldest-pending age when `summary` is
  set. Modelled as a `oneOf`, because the two response shapes genuinely differ. Filters: `status`,
  `access_request_id`, `limit` (1–200, default 50), `offset`.
- `POST /admin/outbox/{id}/retry` — resets a `failed_terminal` row to `pending`; **404** when no such
  row, **409** when the row is in any other state.

Nullable fields (`last_error`, `last_error_code`, `completed_at`, `oldest_pending_age_seconds`) are
typed as nullable rather than plain strings, because the handler does return null for them.

**Every route the admin dispatcher registers is now documented.**

## A note on how these edits were made

`openapi.json` mixes escaped and literal Unicode — the em dash appears 58 times escaped and 6 times
literal. No JSON encoder reproduces the file, so round-tripping it rewrites unrelated descriptions
whichever `ensure_ascii` setting is used: my first attempt produced a **13,919-line diff** for what
should have been an addition.

Both changes here were therefore spliced in as text: **219 + 345 lines added, zero deletions**.
Redocly reports the description valid with **74 warnings, unchanged from `development`**, so the
additions are semantically inert.

That formatting trap is filed separately as #907, along with a proposal for a CI guard that would
have caught it. Worth fixing before the next person edits this file programmatically.

JohnRDOrazio:
## Review follow-up — both findings were real, and one was worse than reported

I reproduced each before changing anything.

### 1. An unreadable decree corpus reported READY

Flagged as "a missing or invalid decrees.json produces an empty decree list". Confirmed, and the
consequence is worse than an empty list: with `decrees.json` **missing or corrupt** but every other
resource present, the whole report came back ready.

```text
decrees.json missing:
  decree_readings            passed:yes   all 0 newly created events have a readings entry
  decree_names               passed:yes   all 14 decreed events are named
  decree_readings_populated  passed:yes   all 0 newly created events have readings text
  >>> ready(): TRUE
```

`decrees()` collapsed *"there are no decrees"* and *"the decrees cannot be read"* into the same empty
array, so every downstream probe passed vacuously. A silent false pass **inside the tool built to
prevent silent false passes** — it would have told an operator a locale was safe to promote while the
corpus was unreadable.

`decrees()` is now nullable, and a gating `decree_source` check turns the null into a failure:

```text
missing  ready:false  en: NOT ready — failing decree_source
corrupt  ready:false  en: NOT ready — failing decree_source
```

It is deliberately **locale-independent**: if the corpus cannot be read, nothing about any locale has
actually been verified, so reporting per-locale readiness would be meaningless.

### 2. `describe()` counted failed advisories as passed

Confirmed:

```text
before : fr: ready (6 checks passed)        # 5 passed, 1 advisory failed
after  : fr: ready (6 of 7 checks passed, 1 advisory not met)
```

It misrepresented precisely the gap the advisory tier exists to surface. The clause is omitted when
there is nothing to report, so a fully-clean locale still reads simply `ready (2 of 2 checks passed)`.

## Verification

- 30 tests in the locale suites, 70 assertions, green. The corpus guard is pinned by a data provider
  covering **absent**, **malformed** and **non-array** corpora against a full copy of the real data
  tree, so `decrees.json` is the only variable and the other probes genuinely still pass.
- `composer analyse` — PHPStan level 10, no errors.
- Full suite: only the three pre-existing `LitCalTestServerTest` failures, which fail identically on
  a clean `development` checkout.

No frontend change needed — the review page renders whatever checks the report returns, so
`decree_source` appears automatically.

PR #903, opened 2026-08-28
feat(decrees): inscribe St John Henry Newman, Doctor of the Church

JohnRDOrazio:
Closes #493.

## The decree

Prot. N. 760/25 of 9 November 2025 — Pope Leo XIV decreed that **Saint John Henry Newman, Priest and
Doctor of the Church**, be inscribed in the General Roman Calendar, his Optional Memorial to be
celebrated by all on **9 October**.

| | |
| --------------- | ---------------------------------------------- |
| `decree_id`     | `StJohnNewman_Create`                          |
| `event_key`     | `StJohnNewman`                                 |
| Date            | 9 October                                      |
| Grade           | 2 (Optional Memorial)                          |
| Colour          | white                                          |
| Common          | Pastors: For One Pastor; Doctors               |
| `since_year`    | 2026                                           |
| Action          | `createNew`                                    |

The Latin text of the decree lives under the Vatican's `/it/` path rather than `/la/`, so it is
carried as a `urls_langs` override rather than through `url_lang_map` — exactly the per-language
override mechanism added in #491.

## How this was produced

The decree was authored through the decree editor against the staging deployment on 2026-08-28 and
written correctly to disk. It could not be committed from there: **the deployment has no git working
tree**, and deploys rsync `--archive --delete` from a fresh checkout with `jsondata/` in the payload,
so the next push to `development` would have destroyed it.

These files were therefore recovered from the staging host and committed here by hand. The contents
are byte-for-byte what the API wrote — nothing was retyped or reconstructed.

That failure mode is not specific to this decree and is tracked separately in #902.

## Contents

**20 files, 77 insertions, 0 deletions** — every change is purely additive:

- `decrees/decrees.json` — the decree entry
- `decrees/i18n/{14 locales}.json` — the event name (`en`, `it`, `la`, `fr`, `nl` carry real
  translations; the remainder are placeholders pending Weblate)
- `decrees/lectionary/{en,fr,it,la,nl}.json` — the readings

Readings: Sir 39:8-14 (gr. 6-11) · Ps 39:2, 4ab, 7-8a, 8b-9, 10 · Mt 23:9b, 10b · Mt 13:47-52

## Verification

- `DecreesSourceSchemaTest`, `SourceDataUrlSchemeTest`, `SchemaValidationTest` — 116 tests, 299
  assertions, green.
- Diffed every file against `development` before committing; zero deletions across all 21 candidate
  files (`lectionary/es.json` and `lectionary/hr.json` were rewritten byte-identically by the handler
[...]

PR #914, opened 2026-08-29
feat(data): source-data publisher — phase 2

JohnRDOrazio:
Refs #902. Phase 2 of the source-data change requests design: **the publisher**. 26 commits, 37 files, +7,031/−55.

Phase 1 (#912) made source-data edits become change requests in Postgres instead of files on disk, because the
deployed server has no git checkout and an `rsync --archive --delete` deploy destroyed anything written there —
which is how the St John Henry Newman decree was nearly lost (#903). Approved requests then sat in Postgres with
nothing to publish them. This is that publisher.

## What it does

```text
approved batch  →  claim (Postgres)
                →  GitHub App auth (RS256 JWT → installation token, cached 50 min)
                →  blob → tree → commit → ref → pull request
                →  publication_status: none → queued → open
```

A cron runner claims one batch at a time and publishes it to a **stable per-resource branch**
(`litcal-data/<resource_type>/<resource_id>`), so the pull request rolls. The commit's **author is the editor**
and the **committer is the App** — that split is the whole point, and it is what preserves per-editor authorship
in the repository.

Phase 3 (merge detection) supplies `merged`/`closed`; nothing here writes them.

## Design decisions worth knowing before review

- **Ancestor exclusion is by age, not by status.** Publishing a batch also publishes the older batch it accumulated
  onto. Marking that ancestor `merged` would assert it was published — and if the containment assumption ever broke,
  the publisher (which selects approved-and-not-merged rows) would skip a batch that was never published and lose it.
  Excluding by age asserts nothing, so the worst case degrades from lost data to a suboptimal rebuild base.
- **`releaseClaim()` reports an observed status, not a row count.** Zero rows conflated "another runner published it"
  with "it was reclaimed to `none`" — and the latter made a real GitHub outage report exit 0 and hammer the API.
- **`force: false`, always.** Serialisation per resource is optimistic concurrency, not mutual exclusion: the loser of
  a same-resource race gets a 422 and retries against the moved ref. That makes `force: false` load-bearing rather
  than merely defensive.
- **A 422 is not an outage.** GitHub answering in detail is positive evidence the API is healthy — the opposite of what
  stop-don't-hammer exists for. It still consumes an attempt, so a branch that 422s forever parks rather than retrying
  forever.
- **A crash strands a batch**, which no exception handler can catch, so batches idle past a grace period are reclaimed.
  The grace is sized against a *whole publish* (`maxRequestsPerBatch × requestTimeout` — the decrees corpus is 28
  requests), not against one request.

## Two defects worth calling out, because both had green test suites

- **The runner's own logger threw on every record it wrote.** The cron script took `LoggerFactory::create()`'s default
  `includeProcessors: true`, which attaches a processor that throws on non-request context. The runner logs batch ids —
  so in production every log call would have thrown, *including those inside its catch blocks, before `releaseClaim()`
  ran*, stranding the batch. Every test injected a hand-built logger with no processors; the class was right, the
  wiring was wrong, and no test crossed that seam.
- **A malformed `GITHUB_REPOSITORY` was an uncaught fatal that every alarm channel called healthy.** The script caught
  only `\RuntimeException` while `fromEnv()` throws `InvalidArgumentException`. Result: exit 255 (undocumented), nothing
  logged past "run starting", and `/health` reporting `configured` because the check tested non-emptiness. A pasted repo
  URL would have accumulated approved work forever with no signal anywhere.

Eleven defects were found in total — four concurrency, two production-only. Every one was found by *running* something;
none by reading. Three were in areas a previous review had already declared correct.

## Not in this PR

- **OpenFGA tuples are not purged for deleted calendars or tests.** Nothing between "PR merged" and "next redeploy"
  performs that purge, so a deleted diocese's former editors retain edit access on an object whose files are gone. This
  is deliberate — only merge detection knows the deletion actually happened — and the runbook states it explicitly.
- **Claim ownership** (a claim token compared in `releaseClaim()`'s `WHERE`) is deferred to phase 3, which must touch the
  same columns and contract anyway. Today a late release can revoke another runner's live claim; bounded, visible, and
  non-destructive.
- **A GitHub App is not registered yet.** Everything is built and tested against mocks. Registration is a documented
  operator step: Contents RW / Pull requests RW / Metadata R, with the private key by path **outside the deployed tree**.
  Until it is set, `/health` reports the publisher unconfigured and approved work accumulates — visibly.

## Verification

4,136 tests / 492,454 assertions / 0 failures. PHPStan level 10 clean on `src` **and** standalone on
`scripts/publish-sourcedata.php` (CI's `phpstan.neon.dist` scans `paths: [src]` only). phpcs, `lint:md`, `lint:jsondata`
all clean. Concurrency is tested with two real OS subprocesses via `proc_open`, because PHP is single-threaded and no
in-process test can make two claims genuinely concurrent.

## Summary by CodeRabbit

- **New Features**
  - Added automated publishing of approved source-data changes to GitHub through commits and rolling pull requests.
  - Added retry handling, stale-claim recovery and parking for repeatedly failed batches.
  - Added publisher configuration options and cron-based execution with clear success and failure summaries.
  - Extended health reporting to show publisher readiness and parked batches.

- **Documentation**
  - Updated operational guidance, release notes and design documentation for publishing workflows, monitoring and known limitations.

- **Tests**
  - Added comprehensive coverage for publishing, retries, concurrency, authentication, GitHub operations and health reporting.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -237,6 +237,10 @@ SOURCEDATA_CHANGE_REQUESTS=false
 # scripts/publish-sourcedata.php) — it is a bearer credential with write access, so see
 # docs/ops/change-request-runbook.md, "Credentials on disk", before widening either.
 # GITHUB_APP_PRIVATE_KEY_PATH=/etc/litcal/github-app.pem
+# NOTE: GITHUB_REPOSITORY is also a built-in GitHub Actions variable, injected into every job as
+# "owner/repo". Anything running inside Actions therefore inherits a well-formed value for it even
+# when nothing here sets one. Harmless in practice — the App id, installation id and key are still
+# required — but do not read a populated GITHUB_REPOSITORY as evidence that the publisher is set up.
 # GITHUB_REPOSITORY=Liturgical-Calendar/LiturgicalCalendarAPI
 # Branch each resource's first publish branches from (default: development)
 # GITHUB_BASE_BRANCH=development
--- a/.env.example
+++ b/.env.example
@@ -222,9 +222,11 @@ SOURCEDATA_CHANGE_REQUESTS=false
 ##
 # Source Data Publisher (phase 2)
 # Publishes approved change-request batches (SOURCEDATA_CHANGE_REQUESTS above) as commits
-# and rolling pull requests via a GitHub App. Without GITHUB_APP_ID, GITHUB_APP_INSTALLATION_ID,
-# GITHUB_APP_PRIVATE_KEY_PATH, and GITHUB_REPOSITORY all set, approved change requests
-# accumulate unpublished — GET /health reports this.
+# and rolling pull requests via a GitHub App. Unless GITHUB_APP_ID, GITHUB_APP_INSTALLATION_ID,
+# GITHUB_APP_PRIVATE_KEY_PATH, and GITHUB_REPOSITORY are all set AND well-formed, approved
+# change requests accumulate unpublished — GET /health reports this. A malformed value counts:
+# GITHUB_REPOSITORY must be exactly "owner/repo", so a pasted URL or a trailing slash is
+# reported the same way an absent value is.
 # Leave unset for a self-hosted instance that does not run the change-request queue.
 ##
 # GITHUB_APP_ID=
--- a/.env.example
+++ b/.env.example
@@ -218,3 +218,22 @@ ALLOW_ENV_ADMIN_FALLBACK=false
 # Leave false for a self-hosted instance that has neither.
 ##
 SOURCEDATA_CHANGE_REQUESTS=false
+
+##
+# Source Data Publisher (phase 2)
+# Publishes approved change-request batches (SOURCEDATA_CHANGE_REQUESTS above) as commits
+# and rolling pull requests via a GitHub App. Without GITHUB_APP_ID, GITHUB_APP_INSTALLATION_ID,
+# GITHUB_APP_PRIVATE_KEY_PATH, and GITHUB_REPOSITORY all set, approved change requests
+# accumulate unpublished — GET /health reports this.
+# Leave unset for a self-hosted instance that does not run the change-request queue.
+##
+# GITHUB_APP_ID=
+# GITHUB_APP_INSTALLATION_ID=
+# Private key path must live outside the deployed tree, not under the web root
+# GITHUB_APP_PRIVATE_KEY_PATH=/etc/litcal/github-app.pem
+# GITHUB_REPOSITORY=Liturgical-Calendar/LiturgicalCalendarAPI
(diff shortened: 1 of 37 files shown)


JohnRDOrazio:
Thanks — four of the six were valid and are fixed in `07fb15c3`. Two are deliberate deferrals; reasoning below so they aren't re-raised.

**Fixed**

- **Plan paths (runbook:14).** Both cited `.superpowers/sdd/...` directories are git-ignored scratch workspaces, deleted when each plan finished — so the runbook pointed at two paths that exist in no checkout. Now points at the committed plans under `docs/superpowers/plans/`.
- **Approval-time schema re-validation (runbook:611).** Correct that it does not happen; the spec was actively wrong to label it Phase 2 when phase 2 did not build it. Relabelled, with the exposure stated: a batch approved against one schema and published after that schema changed. It is bounded rather than silent — the resulting PR runs `lint:jsondata` and schema validation, so an incompatible payload fails visibly instead of reaching `development` — but that is a backstop on the wrong side of the gate, and phase 3 should validate before publishing. The same commit relabels `base_sha` rebase detection, which cannot be built at all today: `recordPublication()` overwrites every row's `base_sha` with the batch-level branch head, destroying the per-file bookkeeping a rebase check would need.
- **Docblock placement.** Correct — I inserted the enumeration test between a docblock and the test it described. Each is now attached to its own test.
- **`putenv()` in the unconfigured checks.** Genuinely fragile, not cosmetic: `getEnvString()` reads `$_ENV` and falls back to `getenv()`, so an inherited process value could have decided those assertions. Both call sites now clear the process environment too.

**Deferred, deliberately**

- **Purging OpenFGA tuples on a merged deletion (runbook:584).** This is phase 3 by design, not an oversight — only merge detection knows the deletion actually happened, and purging at publish time would revoke real authorization on the strength of a proposal that can still be closed unmerged. The gap is stated explicitly in the runbook rather than glossed. Your nuance about **preserving the `admin` tuple so ownership survives recreation** is a good catch that the spec did not record, and I've noted it for the phase 3 design.
- **Requeue after a successful GitHub write (`PublishRunner.php:207`).** Assessed rather than assumed: the retry's `getRef()` returns the head the first attempt pushed, the tree is content-identical, `updateRef()` fast-forwards cleanly, and `findOpenPullRequest()` reuses the open PR. Net effect is one extra empty-diff commit — no data loss, no duplicate PR, no non-fast-forward. Persisting intent before `updateRef` would trade that for a new failure mode (intent recorded, publish fails) on a path that runs on every publish, so it is not obviously a win. Documented in the runbook so a reviewer seeing two identical commits knows why.

Full suite green after the fixes: 4,136 tests / 492,454 assertions / 0 failures.

Issue #924, opened 2026-08-30
ChangeRequestBatch omits pr_number, rejected_reason and publication_settled_at, so a rejection reason is write-only


JohnRDOrazio:
## Summary

`ChangeRequestBatch` omits four columns the table already stores and a client demonstrably needs. All four
are populated by phases 2 and 3; none is readable through any API.

## Missing, and what each blocks

| Column | Populated by | What its absence prevents |
| --- | --- | --- |
| `pr_number` | `recordPublication()` (phase 2) | The History view's "link to the pull request once one exists", which the design promises. Only `ChangeRequestNotification` exposes it today, so the link is reachable from a notification but not from the batch itself. |
| `rejected_reason` | `decideBatch()` (phase 1) | A submitter learning **why** their proposal was refused. `POST /admin/change-requests/{batchId}/reject` accepts a `reason` and stores it; no read path returns it, so the reason is write-only. |
| `publication_settled_at` | phase 3 (`MergePollRunner`) | A history row showing when a batch merged or closed. It is the notification cursor and is deliberately written once, which also makes it the right timestamp to display. |
| `branch`, `commit_sha`, `merge_commit_sha` | phases 2 and 3 | Tracing a batch to what actually landed. Lower priority than the three above, but they are stamped for forensics and currently only reachable by querying Postgres directly. |

`rejected_reason` is the one worth treating as a defect rather than a gap: the API deliberately collects a
human explanation at the moment of refusal and then discards it from every read path, so the feature reads
as implemented while doing nothing for the person it exists for.

## Where

`ChangeRequestBatch` in `jsondata/schemas/openapi.json` (`additionalProperties: false`, so adding fields is
a schema change clients will see), and the `SELECT` in `SourceDataChangeRequestRepository::listBatches()`
plus its `hydrateBatch()`.

Note the batch-level columns are stamped on **every row** of a batch, so the aggregate needs the same
treatment the existing per-batch columns get rather than a bare `MAX()` that would hide a mixed state.

## Scope note

Adding fields to a response is backward compatible for well-behaved clients, but this schema is
`additionalProperties: false` and is used to generate client code, so it warrants a CHANGELOG entry.

Refs #902.

PR #936, opened 2026-08-30
feat(notifications): notify the submitter on the review decision, and document /data's `data` key

### LINKED ISSUE #925: A rejected change request notifies nobody: notify the submitter on the review decision, not only on publication
## Summary

A submitter is never told that their change request was approved or rejected. Phase 3 notifies only when a
batch **settles on GitHub** (`change_request_published`, on merge or close). The review decision itself —
the moment a human actually judges the proposal — produces no notification at all.

This was called out as a known gap in
`docs/superpowers/specs/2026-08-30-sourcedata-merge-detection-design.md`, which says it is "a phase-1 gap
and belongs to its own issue". This is that issue.

## The consequence today

The two outcomes are asymmetric, and the bad one is the silent one:

- **Approved** -> the batch publishes, merges, and the submitter eventually gets a
  `change_request_published` notification. Late, but it arrives.
- **Rejected** -> `review_status` becomes `rejected`, `publication_status` stays `none`, nothing is ever
  published, and therefore **no notification is ever generated**. The submitter's proposal simply stops,
  with no event to observe.

So the only way an editor learns their work was refused is by revisiting `GET /auth/change-requests` and
noticing the status changed. Combine that with the rejection reason being unreadable (see the sibling
issue on `ChangeRequestBatch` fields) and a rejected editor learns neither that it happened nor why.

## Proposed shape

Emit a notification on both transitions in `decideBatch()`, alongside the existing
`change_request_published` type. The substrate exists: `UserNotificationsResponse.items` is already a
discriminated `oneOf` and clients are already required to switch on `type`, so a third shape is additive
in the way the second one was.

Design points worth deciding explicitly:

- **Cursor.** `publication_settled_at` is the publication cursor and is deliberately written once. A
  review decision is a different event at a different time; `approved_at` exists, but there is no
  equivalent stamp for a rejection. Decide whether to reuse `updated_at` (which moves on every claim,
  release and reclaim, and so cannot serve as a cursor) or add a `review_settled_at`.
- **Rejection reason in the payload.** The notification is the natural place to carry it, but the reason
  is currently unreadable everywhere; the two should land together or the notification will say only that
  something was refused.
- **Auto-approval.** A resource admin's own writes are auto-approved in the same request. Notifying
  someone about their own immediate action is noise; suppress it, as the write response already carries
  `disposition: "approved"`.

## Also worth fixing at the same time

The frontend's notification bell **cannot receive any of this for admins**: in admin mode it polls
`/admin/notifications` exclusively, which carries pending access-request counts and no change-request
items at all. So resource admins and global admins — exactly the people whose batches auto-approve —
receive no change-request notifications of any kind today. That half is a frontend fix, but it is worth
knowing here, because a new notification type that no admin can see solves less than it appears to.

Refs #902.

### LINKED ISSUE #933: OpenAPI declares the /data write responses additionalProperties:false but the handler emits an undocumented `data` key
## Summary

Every `/data/*` `PUT` and `PATCH` response schema in `openapi.json` is `additionalProperties: false` and
declares only `success`, `disposition` and `change_request`. `RegionalDataHandler` also emits a `data` key
on all of them. The spec and the implementation disagree on all fourteen operations, and a client written
from the spec would not know `data` exists — while `LiturgicalCalendarFrontend`'s `extending.js` already
depends on it.

## The discrepancy

Handler (`src/Handlers/RegionalDataHandler.php`), the diocesan create path:

```php
$responseObj          = new \stdClass();
$responseObj->success = "Calendar data created for Diocese \"{$diocese_name}\" (Nation: \"{$nation}\")";
$responseObj->data    = $rawPayload;
foreach ($changeRequest as $key => $value) {
    $responseObj->{$key} = $value;
}
```

`$responseObj->data` is set at `:399`, with siblings at `:516`, `:584`, `:702`, `:787` and `:877`.

Spec, for the same operation (`/data/diocese/{key}` `put` `201`):

```json
{
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "success": { "type": "string" },
    "disposition": { "$ref": "#/components/schemas/ChangeRequestDisposition" },
    "change_request": { "$ref": "#/components/schemas/ChangeRequestSubmission" }
  },
  "required": ["success", "disposition"]
}
```

All fourteen affected operations:

| Path                              | Methods       |
| --------------------------------- | ------------- |
| `/data/nation/{key}`              | `put` `patch` |
| `/data/roman/nation/{key}`        | `put` `patch` |
| `/data/diocese/{key}`             | `put` `patch` |
| `/data/roman/diocese/{key}`       | `put` `patch` |
| `/data/widerregion/{key}`         | `put` `patch` |
| `/data/roman/widerregion/{key}`   | `put` `patch` |
| `/data/ambrosian/diocese/{key}`   | `put` `patch` |

## Why this matters

Because `additionalProperties: false` makes it a **wrong-red**, not merely an omission: a strict response
validator run against the live API rejects a response the API is designed to return. Contract tests and
generated clients both break on a correct server.

It also has a queue-mode edge that the frontend is about to trip over. `data` is set from `$rawPayload`
**unconditionally**, in both disk mode and queue mode. In queue mode nothing was written, so `data` is the
*proposed* payload rather than a stored resource — but the spec, by omitting the key entirely, gives no
place to say so. `extending.js` reads `responseData.data.metadata` after a diocesan `PUT` and registers the
returned `diocese_id` as a real calendar, which is precisely the queued-write misreport being tracked on
the frontend side.

## The decision this needs

Whether to document `data` or to stop emitting it is a real choice, not a formality:

- **Document it** — add `data` to the fourteen response schemas, and describe explicitly that in a
  non-`applied` disposition it is the proposed payload and not a stored resource. Preserves the current
  frontend behaviour; requires the frontend to branch on `disposition` before trusting it (which it must
  do regardless).
- **Stop emitting it on non-`applied` dispositions** — leave `data` documented as present only when
  `disposition` is `applied`. Makes the response self-describing and removes the trap at the source, at
  the cost of a contract change the frontend must land first.

The second is the better shape; the first is the safer sequencing. Either way `openapi.json` and the
handler must end up agreeing, and `composer lint:openapi` should not be able to pass while they do not.

## Scope

1. Decide between the two options above.
2. Update the fourteen response schemas in `jsondata/schemas/openapi.json` (edit programmatically — the
   file round-trips through `json.dumps(indent=2, ensure_ascii=True)`).
3. Add a test that asserts the real handler response validates against the documented schema, so the two
   cannot drift again.

Found during the #902 frontend sweep; refs #902.

JohnRDOrazio:
Closes #925
Closes #933

Two contract fixes to the source-data change-request surface. They are unrelated
in mechanism but both are cases where the API's observable behaviour and its
documented contract disagreed.

## #925 — a rejected change request notified nobody

Phase 3 announced a batch only once it **settled on GitHub**
(`change_request_published`, on merge or close). The review decision itself
produced nothing, and the two outcomes were asymmetric in the worst way:

- **Approved** → publishes, merges, submitter eventually gets
  `change_request_published`. Late, but it arrives.
- **Rejected** → `review_status` becomes `rejected`, `publication_status` stays
  `none`, nothing is ever published, so **no notification was ever generated**.

`GET /auth/notifications` now carries a third item shape,
`change_request_reviewed`.

### Design decisions

**One type, not two.** `change_request_reviewed` carries the outcome in a
`review_status` field (`approved` | `rejected`), exactly as the union's existing
`access_request_reviewed` carries `approved` | `rejected` | `revoked` in its
`status`. Two types would give clients two shapes identical apart from a
constant, and would make "did anything happen to my batch" a two-branch question
in every consumer. The frontend distinguishes the outcomes from the one field,
with no second request.

**The outcome comes from a new column, not from `review_status`.** This is the
load-bearing decision. `review_status` is where the batch is *now*, not what was
decided: `markBatchClosedUnmerged()` writes `review_status = 'rejected'` when a
published batch's pull request closes unmerged — on a batch a human **approved**.
An inbox keyed on it would tell that submitter their proposal was refused, and
date the refusal to the moment they were approved. So
`sourcedata_change_requests.review_decision` records the outcome **as decided**,
is written only by `decideBatch()`, and is never moved. That batch is not left
silent either: its close is reported by the publication item, with
`publication_status: "closed"`, at the time it actually closed. Both halves are
pinned by `testAClosedUnmergedPullRequestDoesNotRewriteTheReviewDecision`.

`rejected_reason` is likewise scoped to the decision in SQL
(`CASE WHEN review_decision = 'rejected' …`), because `markBatchClosedUnmerged()`
also stamps its close text into that column — an approval must not surface it.

**The cursor is `approved_at`, and no `review_settled_at` was added.** The issue
suggested either reusing `updated_at` (impossible — it moves on every claim,
release, reclaim and record) or adding a stamp. Neither was needed: despite its
name, `decideBatch()` already stamps `approved_at` for **both** outcomes, and its
`WHERE review_status = 'submitted'` guard makes that a single write per batch —
precisely the property `publication_settled_at` has on the publication axis. A
`review_settled_at` would always equal it and could only drift. The misleading
name is corrected in a `COMMENT ON COLUMN` so a reader of the schema is not
misled, and in the `decideBatch()` docblock.

**Self-decisions are suppressed** via
`approved_by_sub IS DISTINCT FROM submitted_by_sub`. The auto-approval a resource
admin's own write receives inside the same request is not news, and the write
response already answered `disposition: "approved"` synchronously. This equally
covers an admin approving their own batch through the admin endpoint.

**Composition with #924 (concurrent, separate branch).** That issue exposes
`rejected_reason` on `ChangeRequestBatch`. Nothing here touches
`ChangeRequestBatch` or its schema. The new notification declares its **own**
`rejected_reason` property inline, reading the same DB column, so the two compose
(the notification shows the reason inline; the batch endpoint remains the place
to fetch it for a batch not in the inbox page) and the OpenAPI edits do not
overlap.

**Backfill.** Historical decisions are reconstructed in two exact cases rather
than by copying `review_status` wholesale: a batch still at
`publication_status = 'none'` was never published so nothing rewrote its status;
anything past `none` must have been approved to be claimable. They become visible
the first time a user opens their inbox — intended, since these decisions were
never announced at all.

**Known, out of scope.** As the issue notes, the frontend bell polls
`/admin/notifications` in admin mode and will not show any of this to admins
until that is fixed frontend-side.

## #933 — the OpenAPI `/data` write responses contradicted the handler

All fourteen `/data/*` `PUT`/`PATCH` response schemas were
`additionalProperties: false` and declared only `success`, `disposition` and
`change_request`, while `RegionalDataHandler` also sets `$responseObj->data`
(`:399`, `:516`, `:584`, `:702`, `:787`, `:877`). A strict response validator
therefore rejected a response the API is designed to return, and a client written
from the spec did not know the key existed — while the frontend's `extending.js`
already depends on it.

**Option 1 implemented: document `data`, do not remove it.** Removing it is the
better long-term shape but a breaking contract change the frontend must land
first, and the frontend fix for exactly this is being written concurrently.

The description does not merely add the key; it states the trap: `data` is set
from the raw payload **unconditionally**, in both disk mode and queue mode, so
when `disposition` is not `applied` nothing was written and `data` is the
*proposed* payload, not a stored resource — a client must branch on `disposition`
before trusting it. It also records that the `i18n` member is stripped before the
echo, since those translations are written as separate per-locale files.

`data` is deliberately **not** in `required`, so a later revision may stop
emitting it on non-`applied` dispositions without a second breaking change.

`openapi.json` was edited programmatically; the file round-trips exactly through
`json.dumps(indent=2, ensure_ascii=False)`, verified before and after.

## The regression guard

`phpunit_tests/Handlers/RegionalDataWriteResponseSchemaTest.php`:

- drives the **six real response-assembly sites** (nation/diocese/widerregion ×
  PUT/PATCH) through `RegionalDataHandler` in queue mode — the disposition under
  which `data` is at its most misleading — and validates each real response body
  against the response schema `openapi.json` declares for that operation, via a
  JSON Pointer into the document;
- asserts all **fourteen** documented operations declare `data` identically (the
  rite-qualified spellings route to the same code), that the description states
  the `disposition`/`applied` condition, and that `data` is not `required`.

`NotificationsHandlerTest::testTheInboxValidatesAgainstTheDocumentedUnion` does
the same on the other side: a real inbox containing both an
`access_request_reviewed` and a `change_request_reviewed` item, validated against
`UserNotificationsResponse`. Because every member of the `oneOf` is
`additionalProperties: false`, an item shape the union does not list matches zero
branches and fails.

## Verification

```
$ composer lint          → exit 0
$ composer analyse       → [OK] No errors (PHPStan level 10, 390 files)
$ composer lint:openapi  → Woohoo! Your API description is valid. 🎉
$ openapi round-trip     → canonical round-trip identical: True
```

`composer test:quick` reports `Tests: 4179, Errors: 21, Failures: 100`. **Every
one of those is in `phpunit_tests/Routes/` (104) or `phpunit_tests/WebSocket/`
(53)** — the suites that talk to `localhost:8000` and the WebSocket server, and
in this environment `localhost:8000` answers `404` to everything (including
`/easter`, which this branch does not touch). Verified pre-existing: with the
branch stashed, i.e. a tree identical to `origin/development`,
`Routes/Readonly/EasterTest` + `HealthTest` fail identically with
`Failed asserting that 404 is identical to 200`.

Everything this branch actually touches is green:

```
$ vendor/bin/phpunit --exclude-group slow --exclude-group golden-master-generate \
    phpunit_tests/{Handlers,Repositories,Schemas,Services,Models,Http,Params,Methods,Enum,Database}
OK, but there were issues!
Tests: 3306, Assertions: 34572, Warnings: 1, Skipped: 6.
```

(The one warning is the pre-existing `mkdir(): File exists` in
`DiskSourceDataWriterTest`.)

## Migration

`Version20260901120000` adds `review_decision` (nullable, `CHECK` on
`approved`/`rejected`), a `COMMENT ON COLUMN` correcting `approved_at`'s
misleading name, a partial index `idx_scr_decided_for_submitter`, and the
backfill. Applied and verified locally; pinned by two new assertions in
`SourceDataChangeRequestSchemaTest`.

## Summary by CodeRabbit

* **New Features**
  * Added inbox notifications when submitted change requests are approved or rejected, including the decision and any rejection reason.
  * Notifications distinguish review outcomes from later publication updates and exclude decisions made by the submitter.
  * Documented the `data` field returned by `/data/*` write operations, allowing clients to inspect submitted payloads alongside the disposition.

* **Documentation**
  * Updated API and notification documentation to describe the new response and notification formats.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,31 @@
   reviewer or a stopped poller, since an undetected merge is indistinguishable from an unreviewed one from
   this side. See `docs/ops/change-request-runbook.md`'s "Merge detection (phase 3)" section for the full
   operator playbook.
+* **notify a change request's submitter on the review decision, not only on publication**, see issue
+  [#925](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/925). Phase 3 announced a batch
+  only once it settled on GitHub, so an approval arrived late and a **rejection arrived never**: a rejected
+  batch is never published, so no notification was generated at all, and the only way an editor learned their
+  work was refused was to notice a status change on `GET /auth/change-requests`. `GET /auth/notifications`
+  now carries a third item shape, `change_request_reviewed`, whose `review_status` (`approved`/`rejected`)
+  and `rejected_reason` let a client tell the two outcomes apart without a second request. It is additive:
+  the `items` list was already a discriminated union and clients were already required to switch on `type`.
+  One item per batch, keyed on a new `sourcedata_change_requests.review_decision` column that records the
+  outcome AS DECIDED — deliberately not `review_status`, which merge detection rewrites to `rejected` when a
+  published batch's pull request closes unmerged, on a batch a human approved. A decision an editor made on
+  their own batch (the auto-approval a resource admin's own write receives inline) produces no notification:
+  the write response already answered `disposition: "approved"`. Historical decisions are backfilled, so they
+  become visible the first time a user opens their inbox.
+* **document the `data` key the `/data/*` write responses have always emitted**, see issue
+  [#933](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/933). All fourteen `PUT`/`PATCH`
+  response schemas were `additionalProperties: false` and declared only `success`, `disposition` and
+  `change_request`, while `RegionalDataHandler` also sets `data` — so a strict response validator rejected a
+  response the API is designed to return, and a client generated from the spec did not know the key existed.
+  No behaviour changes: `data` is documented rather than removed, because removing it is a contract change
+  the frontend must land first. The documentation states the trap explicitly — `data` is set from the raw
+  payload **unconditionally**, in both write modes, so under a `submitted` or `approved` disposition nothing
+  was written and it is the *proposed* payload rather than a stored resource. Clients must branch on
+  `disposition` before trusting it. A new test drives the real handler and validates its response against the
+  documented schema, so the two cannot drift again.
 * **`PATCH` on the seven `/data/*` calendar routes now answers `200 OK` instead of `201 Created`**, see
   issue [#913](https://github.com/Liturgical-Calendar/LiturgicalCalendarAPI/issues/913). The affected
   routes are `/data/nation/{key}`, `/data/roman/nation/{key}`, `/data/diocese/{key}`,
--- a/docs/superpowers/specs/2026-08-30-sourcedata-merge-detection-design.md
+++ b/docs/superpowers/specs/2026-08-30-sourcedata-merge-detection-design.md
@@ -243,6 +243,15 @@ produces. Review decisions (approve, reject at the gate) are equally notificatio
 unnotified today, but that is a phase-1 gap and belongs to its own issue rather than being smuggled
 in here.
 
+> **Amended, #925.** That issue was filed and closed the gap: a third item type,
+> `change_request_reviewed`, reports the decision itself. A batch can now produce one item of each
+> kind at different times, and a REJECTED batch — which never publishes, and so produced nothing at
+> all under phase 3 alone — produces the review item. Its cursor is `approved_at`, which
+> `decideBatch()` stamps for both outcomes under its single-shot `review_status = 'submitted'`
+> guard; the outcome itself is a new `review_decision` column, because `review_status` is the
+> batch's current position and `markBatchClosedUnmerged()` rewrites it to `rejected` on a batch a
+> human approved. Decisions an editor made on their own batch are suppressed.
+
 ### Merged in PHP, not in SQL
 
 `fetchInbox()` queries each source separately, merges the two lists in PHP, sorts by timestamp
--- a/jsondata/schemas/openapi.json
+++ b/jsondata/schemas/openapi.json
@@ -6467,6 +6467,11 @@
                         "National calendar created for nation {NATION}"
                       ]
                     },
+                    "data": {
+                      "type": "object",
+                      "additionalProperties": true,
+                      "description": "The payload the request proposed, echoed back with its `i18n` member removed — those translations are written as separate per-locale files, and the handler strips the member before echoing.\n\n**Not necessarily a stored resource. Branch on `disposition` before trusting it.** The handler sets this key UNCONDITIONALLY, in both write modes. Only `disposition: \"applied\"` means the bytes were written to the source-data tree and this describes what is now stored. Under `submitted` or `approved` the write was recorded as a change request and NOTHING was written: this is the PROPOSED content, and reading it as a stored resource — registering the `metadata.diocese_id` it names as an existing calendar, say — reports a queued write as a completed one.\n\nDocumented rather than required: it is emitted on every write response today, and is declared here so a strict response validator stops rejecting a response the API is designed to return (#933). It is deliberately absent from `required` because the better long-term shape is to stop emitting it when the disposition is not `applied`, and clients must not be built to depend on its presence."
+                    },
                     "disposition": {
                       "$ref": "#/components/schemas/ChangeRequestDisposition"
                     },
@@ -6625,6 +6630,11 @@
                     "success": {
                       "type": "string",
                       "example": "National calendar updated for nation {NATION}"
+                    },
+                    "data": {
+                      "type": "object",
+                      "additionalProperties": true,
+                      "description": "The payload the request proposed, echoed back with its `i18n` member removed — those translations are written as separate per-locale files, and the handler strips the member before echoing.\n\n**Not necessarily a stored resource. Branch on `disposition` before trusting it.** The handler sets this key UNCONDITIONALLY, in both write modes. Only `disposition: \"applied\"` means the bytes were written to the source-data tree and this describes what is now stored. Under `submitted` or `approved` the write was recorded as a change request and NOTHING was written: this is the PROPOSED content, and reading it as a stored resource — registering the `metadata.diocese_id` it names as an existing calendar, say — reports a queued write as a completed one.\n\nDocumented rather than required: it is emitted on every write response today, and is declared here so a strict response validator stops rejecting a response the API is designed to return (#933). It is deliberately absent from `required` because the better long-term shape is to stop emitting it when the disposition is not `applied`, and clients must not be built to depend on its presence."
                     },
                     "disposition": {
                       "$ref": "#/components/schemas/ChangeRequestDisposition"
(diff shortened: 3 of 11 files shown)


JohnRDOrazio:
Valid, and there were **two** silent-shrink holes rather than one. Fixed in a0c703cb.

## The second hole

Beyond deriving the cases from the document under test, the provider also did:

```php
if ($schema === null || !isset($schema['properties']['disposition'])) {
    continue;
}
```

So a response that *lost* `disposition` didn't fail — it left the run. Both holes shrink the contract while
the suite stays green, which is the failure mode the test exists to prevent.

## What changed

The fourteen `(path, method, status)` triples are now enumerated in `expectedWriteOperations()`, and a new
test compares them against the document **as a set, in both directions**. A set rather than a count: a count
catches a deletion but not a swap, and its failure message names nothing. `disposition` is now asserted per
operation instead of filtering the provider, so losing it fails that operation.

## Verified by mutation — both previously passed

Deleting `/data/ambrosian/diocese/{key}`:

```
The set of /data/* write responses in openapi.json has changed. If a rite or route was
added, extend expectedWriteOperations() — the new operation inherits the `data` contract.
Failed asserting that two arrays are identical.
```

Stripping `disposition` from `PATCH /data/nation/{key}` 200:

```
PATCH /data/nation/{key} (200) must document `disposition`; it is what tells a client
whether `data` was stored
Tests: 14, Assertions: 42, Failures: 1
```

`openapi.json` restored after each; `git diff` clean.

## One note on the tradeoff

The old docblock justified discovery as "a new rite spelling picks itself up", which is a real property this
change gives up — adding a rite now reds the list until someone extends it. I think that is the right trade
rather than a regression: a new rite's write routes inherit the `data` trap (the key is set from the raw
payload unconditionally, so under a non-`applied` disposition it is the *proposed* payload and not a stored
resource), and that deserves a deliberate acknowledgement rather than silent enrolment. The failure message
says exactly what to do.

`composer lint`, `composer analyse` ([OK] No errors), `composer lint:jsondata` and the suite under coverage
(`OK (35 tests, 213 assertions)`, zero warnings) all pass.

PR #937, opened 2026-08-30
feat(change-requests): expose what a batch proposes, and the columns it stamps

### LINKED ISSUE #923: No way to see what a change request proposes: add a batch detail endpoint with per-file content
## Summary

There is no way to see **what a change request actually proposes**. An administrator deciding a batch is
shown a resource id and a list of file paths, and nothing else. This blocks the change-request review UI in
`LiturgicalCalendarFrontend`, which cannot be built as designed until it exists.

## Current state

`GET /admin/change-requests` returns `AdminChangeRequestListResponse`, whose `ChangeRequestBatch` is
`additionalProperties: false` and carries no file content. `SourceDataChangeRequestRepository::listBatches()`
aggregates only:

```text
batch_id, resource_type, resource_id, review_status, publication_status,
submitted_by_sub, submitted_by_name, submitted_by_email,
approved_by_sub, created_at, updated_at,
COUNT(*) AS file_count, ARRAY_AGG(path) AS paths
```

There is no per-batch detail route: `src/Router.php` registers list, approve and reject under
`/admin/change-requests` and nothing else.

The content is in the table — `content` is populated per row, and `findUnpublishedContent()` reads it for
the publisher — it is simply never exposed to a reviewer.

## Why this matters more than a missing convenience

Approval is the **only** human gate in the whole phase 1-3 design. Everything downstream is automatic: an
approved batch is claimed, committed, and opened as a pull request without further review. Asking an
administrator to approve a change they cannot see makes that gate ceremonial. It also pushes the real
review onto the resulting pull request, which is the opposite of what #902 set out to build — the pull
request is meant to be the durable record of a decision already taken, not the place the decision happens.

## Proposed shape

`GET /admin/change-requests/{batchId}` returning the batch plus its rows, each with at least:

- `path`, `operation`
- `content` — the proposed bytes
- ideally the **current** bytes for the same path, so a client can render a diff rather than a blob; the
  publisher already resolves this when it builds a commit, and a reviewer needs exactly the same comparison

Authorization must be the same specific-batch check `approve()`/`reject()` already perform, not the
post-SQL list filter — see `ChangeRequestAdminHandler` on why those are deliberately different.

Consider whether a submitter-facing equivalent under `/auth/change-requests/{batchId}` should land at the
same time, so an editor can review their own proposal before an administrator sees it. Same content, a
different authorization rule (own `sub` only).

## Notes

- Content can be large; the decrees corpus batches 22 files. Whether the detail route streams, paginates
  by row, or simply accepts the size is worth deciding rather than discovering.
- A `delete` row has `content IS NULL` by table constraint, so the shape must allow null rather than
  treating it as missing data. Note `operation = 'delete'` does not mean the resource is being deleted —
  see the runbook's warning about locale removal on a live calendar.

Blocks the review UI. Refs #902.

### LINKED ISSUE #924: ChangeRequestBatch omits pr_number, rejected_reason and publication_settled_at, so a rejection reason is write-only

JohnRDOrazio:
Closes #923
Closes #924

## #923 — a reviewer can now see what a change request proposes

Two new routes:

- `GET /admin/change-requests/{batchId}` — the reviewer's view
- `GET /auth/change-requests/{batchId}` — the submitter's view of their own batch

Both return `{batch, files, content_included}`. `batch` is exactly the object the listing routes already
return; each entry in `files` carries `path`, `operation`, `base_sha`, the **proposed** bytes (`content`)
and the bytes **currently** at that path in the deployed source tree (`current_content`), so a client
renders a diff rather than a blob.

Approval is the only human gate in the phase 1–3 design — everything downstream of it is automatic — so a
reviewer shown a resource id and a list of file paths was being asked to approve a change nobody had read.

**Authorization is re-checked on the specific batch id**, through the same OpenFGA path `approve`/`reject`
use, and never inherited from the post-SQL list filter: that filter answers "which of these may you
review", not "may you review this one". Both routes answer `404` rather than `403`, so a refusal cannot
confirm that a batch the caller cannot touch exists. The submitter route is scoped to the caller's own
`sub` **in SQL** (on the rows and again on the collapsed batch), so a handler bug cannot widen it.

`include_content=false` suppresses the file bodies while keeping their byte counts, for a client sizing a
large batch (the decrees corpus batches 22 files). `content_included` on the response distinguishes a
suppressed null from the null a `delete` row carries by table constraint — `chk_scr_delete_has_no_content`
makes that null data, not missing data.

## #924 — the columns that were stored but unreadable

`ChangeRequestBatch` gains `rejected_reason`, `pr_number`, `branch`, `commit_sha`, `merge_commit_sha` and
`publication_settled_at`, on **both** `GET /admin/change-requests` and `GET /auth/change-requests` (they
share the schema).

`rejected_reason` is the defect rather than the gap: `POST /admin/change-requests/{batchId}/reject`
deliberately collects a human explanation, `decideBatch()` stores it, and no read path returned it — so it
was write-only and a submitter could never learn why they were refused.

### Submitter vs. admin visibility — the deliberate call

**All six go to submitters as well as reviewers, and nothing was removed.** None of the six identifies a
reviewer. The one column that does, `approved_by_sub`, already existed on the submitter-facing response
before this PR; removing it would be a breaking change to a `required` field of an
`additionalProperties: false` schema used for codegen, and is out of scope here. The rule applied for new
fields is therefore: *a submitter sees the fate of their own change, never a new fact about who judged it.*
Every one of the six describes the change's fate — a rejection reason written for them, a pull request
link their History view is meant to offer, and the git facts that become public the moment the PR opens.

Aggregation: `MIN()`, matching the existing per-batch columns and deliberately **not** `MAX()`. SQL
aggregates skip NULLs, so on a hypothetically mixed batch `MIN()` surfaces the stamped value instead of
hiding it behind the NULLs of rows a partial update missed. `pr_number` is narrowed with `requireInt()` so
a client generated from the schema gets the `integer` openapi.json declares, not the string PDO's pgsql
driver may hand back.

## Schema

`jsondata/schemas/openapi.json` gains `ChangeRequestFile` and `ChangeRequestBatchDetail` plus the two new
paths. The batch is **nested** under `batch` rather than spread at the top level, because
`ChangeRequestBatch` is `additionalProperties: false` and an `allOf` bolting `files` alongside its
properties would be invalid against its own first branch.

Edited programmatically (a Python script), never by hand, and verified with a round-trip diff: the file
still re-serializes exactly through `json.dumps(indent=2, ensure_ascii=True)` apart from the pre-existing
literal-non-ASCII lines, whose count is unchanged.

**These schemas generate client code**, so a regenerated client sees six new required fields on
`ChangeRequestBatch`. Every one is nullable and is null on a batch that has not reached the phase that
writes it. Hence the CHANGELOG entry, and a key-for-key contract test asserting the handler's body and the
schema agree about which keys exist — the one thing `additionalProperties: false` drift makes invisible to
every other assertion.

## Verification

```
composer lint         # phpcs — clean (no output)
composer analyse      # PHPStan level 10 — [OK] No errors
composer lint:openapi # Redocly — "Woohoo! Your API description is valid."
composer test:quick   # 4168 tests; 21 errors / 100 failures
```

Every one of those `test:quick` failures is in `Routes/*` (which hit `localhost:8000`, the *shared* main
checkout, not this branch) or `WebSocket/*` (which need the WS server). **The identical set fails on
untouched `development`**: the run was repeated with these changes stashed, and the failing
class/count table is byte-identical (`Routes/Readonly/Temporale` 24, `CanonicalRiteLinkHeader` 14,
`Routes/ReadWrite/RegionalData` 12, …). Outside those two families the run is completely clean — zero
failures in `Handlers/*`, `Repositories/*`, `Services/*`, `Models/*`.

New tests (none in the `slow` group):

- `phpunit_tests/Handlers/ChangeRequestAdminHandlerTest.php` — proposed **and** current content against a
  path that really exists on disk; delete-row nulls; `include_content=false` keeping sizes; an
  unrecognised value rejected; authorization re-checked on the specific batch id; unknown/malformed batch
  ids; the #924 columns on a rejected batch and on a published-then-merged one; two key-for-key schema
  contract tests.
- `phpunit_tests/Handlers/AuthChangeRequestHandlerTest.php` — the submitter's own detail read, a submitter
  reading *why* they were rejected, someone else's batch answering 404, malformed id, trailing segment.
- `phpunit_tests/Repositories/SourceDataChangeRequestRepositoryTest.php` — `findBatchSummary()` returning
  the same object the listing does, its submitter scoping, `getBatchBySubmitter()` scoped in SQL, and the
  aggregate carrying the decision and publication columns.

Docs: a runbook section on reading a batch and a table of the batch fields, plus two CHANGELOG entries.

## Summary by CodeRabbit

- **New Features**
  - Added detailed change-request batch views for submitters and administrators.
  - View proposed files, current file contents, file sizes, and operations.
  - Added an option to hide file contents while retaining metadata.
  - Exposed review decisions and publication details in batch listings and details.
- **Documentation**
  - Updated the changelog, API specification, and operational runbook with the new endpoints, fields, and content-handling behaviour.
- **Bug Fixes**
  - Improved access control so batch details are only visible to authorised viewers.


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
Valid, and it caught a second staleness in the same table. Fixed in 1d840f19.

`review_decision` was added to `ChangeRequestBatch` after `development` was merged in (it arrives with #936,
which introduced the column), and the runbook's field reference was not updated with it.

While adding it: the table also described `rejected_reason` as **"Why a batch was refused"**, which is wrong
on the one path where the distinction matters. A batch whose pull request closed unmerged carries text the
merge poller generated, and nobody refused it — the same conflation #936 fixed in the notification path.

## What changed

- `review_decision` added to the table, written by `decideBatch()`, once.
- `rejected_reason`'s description no longer claims refusal.
- A new section on the single path where `review_status` and `review_decision` disagree, with the concrete
  shape a client sees for an approved-then-closed batch:

  | Field                | Value                       |
  |----------------------|-----------------------------|
  | `review_status`      | `rejected`                  |
  | `review_decision`    | `approved`                  |
  | `publication_status` | `closed`                    |
  | `rejected_reason`    | the poller's generated text |

  and the point of it: a client showing "your change was refused" off `review_status` there would be telling
  the submitter something untrue.

Every value in that example was read off `markBatchClosedUnmerged()`'s `UPDATE` rather than inferred — it
sets `publication_status = closed` and `review_status = rejected`, stamps `publication_settled_at`, and
leaves `review_decision` untouched (confirmed: zero occurrences in the method).

`composer lint:md` → `Summary: 0 issues in 0 files` across 97 files.

Issue #918, opened 2026-08-30
Re-validate a change request against the current schema before publishing


JohnRDOrazio:
Deferred from phase 3 (#902), which was scoped without it deliberately rather than by oversight.

`SourceDataChangeRequestRepository::approveBatch()` is a single status `UPDATE`. Nothing re-checks the
proposed content against the current JSON schema at the moment of approval, so a batch approved against
one schema and published after that schema changed produces a pull request whose CI fails
`composer lint:jsondata`.

The exposure is bounded rather than silent — the failure surfaces on the pull request instead of reaching
`development` — but it is a backstop on the wrong side of the gate. An administrator approves something
the system has already accepted, and only later does CI reveal it was no longer valid. Validating before
publication would keep the approval decision and the validity check in the same moment.

See `docs/superpowers/specs/2026-08-30-sourcedata-merge-detection-design.md`, section "Scope", and the
"Two landmines in the accumulation base" discussion in
`docs/superpowers/specs/2026-08-28-sourcedata-change-requests-design.md`.

PR #938, opened 2026-08-30
feat(locales): add /admin/locales write route for locale curation

### LINKED ISSUE #926: Locale curation is still hardcoded read-only: add an /admin/locales write route now that #902 has shipped
## Summary

`GET /admin/locales` reports `curation.writable = false` unconditionally, with a reason that no longer
holds. #902 has shipped all three phases, so the durable write path the locale curation UI was waiting for
now exists — but nothing was wired up to it, and there is still no write route.

## Current state

`src/Handlers/Admin/LocalesAdminHandler.php:120-122` hardcodes:

```php
'writable' => false,
'reason'   => 'Promotion writes jsondata/supportedLocales.json, which the next deploy would overwrite (see issue #902). Promote by opening a pull request against that file.',
```

`openapi.json` documents `GET` only on `/admin/locales` and `/admin/locales/{locale}`.

On the frontend side `admin-locales.php` already exists, is global-admin gated, renders
`curation.writable` / `curation.reason` verbatim, and shows per-locale readiness. It is a read-only page
waiting for exactly this.

Meanwhile `LocaleReadinessChecker` now has its CI guard and a `/health` block (#904), so the pre-flight
test that gates a promotion is in place.

## The decision this needs first

`supportedLocales.json` lives at `jsondata/supportedLocales.json` — a **top-level** reference resource,
sibling to `likelySubtags.json` and `world_dioceses.json`. It is **outside** `jsondata/sourcedata`, which
is where every existing change-request path is rooted.

The design carved out exactly one comparable exception, the `tests` partition, which needed
`TestScopeResolver` as its scoping authority. This needs the equivalent decision made deliberately:

- a **new resource type** for curated top-level reference data, with its own `resource_id` derivation and
  its own OpenFGA object type; or
- **reuse an existing type** the way `decrees` reuses `general_roman_calendar`, accepting that the
  authorization story becomes "whoever administers the general Roman calendar curates locales".

Whichever is chosen, `SourceDataSchemaResolver` (added for approval-time re-validation) needs a mapping
for the path, or the batch will be approved without being validated.

## Proposed work

1. Decide the resource-type question above.
2. Add promote/demote through the `SourceDataWriter` seam, so a promotion becomes a reviewable change
   request in queue mode and a direct disk write in disk mode, exactly as every other source-data write
   does.
3. Gate a promotion on `LocaleReadinessChecker` passing for that locale — the check exists precisely to be
   this pre-flight, and a locale promoted while unready tightens enforcement and starts throwing on
   missing readings.
4. Flip `writable` to reflect the real state (queue mode configured and the caller authorized) rather than
   a constant, and replace the reason string.
5. Document the route in `openapi.json`.

Then the frontend's read-only notice can become a Promote action; that side is small once this lands.

Worth noting the governance argument for routing this through change requests rather than a direct write:
promoting a locale is a curation decision that deserves an audit trail and a second pair of eyes, which is
what a reviewed pull request against `supportedLocales.json` gives it — the same reasoning #904 used when
it deferred this in the first place.

Refs #904, #902.

JohnRDOrazio:
Closes #926. Unblocks LiturgicalCalendarFrontend#506 — the read-only notice in `admin-locales.php` can become a Promote action.

## What this does

`GET /admin/locales` reported `curation.writable = false` unconditionally, with a reason citing #902 — which had already shipped. The durable write path the curation UI was waiting for existed; nothing was wired to it and there was no write route.

- `POST /admin/locales/{locale}/promote`
- `POST /admin/locales/{locale}/demote`

Both go through the `SourceDataWriter` seam, so a promotion is a reviewable change request in queue mode and a direct disk write in disk mode, exactly like every other source-data write. POST matches the house idiom for admin actions (`/admin/access-requests/{id}/approve`, `/admin/change-requests/{batchId}/approve`); neither takes a request body.

## Resource identity — the decision from #926, applied

`ChangeResource::supportedLocales()` returns `general_roman_calendar:supported_locales`, a fixed id on the existing type, exactly as `decrees()` does, and `supported_locales` joins `AccessRequestRepository::GRC_OBJECT_IDS`.

**No OpenFGA model migration.** Object ids are not part of the authorization model — only types and relations are — so this needs none, which matters while the dev store is deliberately held on the additive model until both APIs run merged code.

## The validation gap this had to close first

`SourceDataSchemaResolver` had no mapping for `jsondata/supportedLocales.json`, and there was **no schema for the file at all**. `forPath()` answered `null`, which `ChangeRequestSchemaValidator` reads as *not validated* rather than *invalid* — a malformed promotion would have been approved unchecked (#918's gate, blind here).

So this adds `jsondata/schemas/SupportedLocales.json` (draft-07, no `$id`, `$ref`s `CommonDef.json#/definitions/Locale`), a `LitSchema::SUPPORTED_LOCALES` case with role **`SOURCE`**, and the resolver mapping built from the same `JsonData::SUPPORTED_LOCALES_FILE` constant the handler writes through.

`SOURCE` because the line that role draws is stored-versus-emitted, not which directory a file sits in: these are bytes the repository stores and a change request proposes, never a response the API emits. `SchemaRole::SOURCE`'s docblock is widened to say so. The file is deliberately **not** a `CheckableItem` — `/validations` enumerates the corpus a calendar is assembled from and asks exists/parses/validates/covers, while this is one top-level file with no locale-folder shape to cover, whose real invariant (do the locales it names actually have their data?) is already asserted far more strongly by `composer lint:locales` and `/health`'s `locale_readiness` block.

The load-bearing constraint the schema encodes is `official`'s `minItems: 1`: an empty list is **not** "no official locales" — `SupportedLocales::official()` substitutes its built-in `FALLBACK` for an empty or unreadable resource, so a write that emptied it would silently restore the historical five.

## Promotion is gated; demotion is gated differently, on purpose

**Promotion** requires `LocaleReadinessChecker::check($locale)->ready()`, and the gate is **not bypassable** — no `force` parameter, no role that skips it. An official locale is served strictly (`ReadingsMap::getReadings()` throws instead of degrading), so promoting an unready locale converts silent gaps into 500s on a calendar the API advertises. If the data is complete the checker says so; if the checker is wrong, the fix is to the checker. A promotion also drops the locale's `candidates` note, whose prose is false once it is official.

**Demotion is deliberately not readiness-gated.** Removing a locale *loosens* enforcement — the same missing readings that would have thrown now degrade quietly — so it can never turn a working calendar into a 500. What it *can* do is change published output silently: `/calendars` stops advertising the locale and affected events serve nothing instead of erroring. That is a governance risk, answered by review and the audit trail, not by a readiness probe. Two integrity guards still apply, both `409`: the locale must currently be official, and the **last** official locale may not be removed, for the `FALLBACK` reason above.

## `writable` now reflects reality

`curation` reports `writable`, a new `mode` (`change_request` | `disk` | `misconfigured`) and prose the frontend still renders verbatim. A **misconfigured** queue mode — flag set, Postgres and/or OpenFGA unreachable — refuses the write with `503` instead of falling back to disk. The calendar write handlers tolerate that fallback because refusing would stop calendar editing outright; promotion is a rare governance action with an understood manual alternative, so an edit the next deploy silently reverts is the worse outcome.

## Aggregate-file read

`supportedLocales.json` holds the whole official set in one file, so the handler reads current content through `unpublishedSourceContent()` rather than off disk. In queue mode a submitter's previous promotion never reaches disk, and rebuilding from disk would drop it — the defect that once lost a decree behind a `201`.

## Tests

- `Handlers/LocalesAdminCurationTest` — disk mode against a throwaway tree (real writable resource copy, symlinked data folders), asserting every refusal branch is *forced* rather than assumed: unready locale, already-official, unknown locale, non-official demote, last-locale demote, misconfigured mode, unknown action, unauthenticated, non-admin, wrong method. `tearDown` digests the committed resource so a stray write to it fails this class rather than the next one.
- `Handlers/LocalesAdminQueueModeTest` — queue mode (DB-gated, skips cleanly): the proposal targets `general_roman_calendar:supported_locales`, no file is touched, a second promotion accumulates onto the first unpublished one, and the queued row is claimed by `SupportedLocales.json` and passes `ChangeRequestSchemaValidator`.
- `Schemas/SupportedLocalesSchemaTest`, `Schemas/OpenApiAdminLocalesTest`, plus updates to `ChangeResourceTest`, `SchemaRoleTest`, `SourceDataSchemaResolverTest`, `AccessRequestRepositoryConstantsTest` and `LocalesAdminHandlerTest` (whose `writable === false` / `#902` assertion had become a lie).

Nothing added to the `slow` group.

## Verification

```
composer lint       → exit 0
composer analyse    → [OK] No errors (PHPStan level 10, 389 files)
composer lint:openapi → Woohoo! Your API description is valid. 🎉
phpunit (all in-process suites: Database Enum Handlers Http Methods Models
         Params Repositories Schemas Services Test)
                    → OK, Tests: 3371, Assertions: 34634, Warnings: 1, Skipped: 6
```

`composer test:quick` also reports 21 errors / 100 failures, entirely in `Routes/*` and `WebSocket/*`. Those are HTTP tests against the shared `localhost:8000` / `:8082` servers, which do not serve this branch. Confirmed pre-existing: `origin/development` at `3aa2dd8b` produces the byte-identical tally — `Tests: 157, Errors: 21, Failures: 100, Warnings: 1, Skipped: 11` — from the same worktree.

`jsondata/supportedLocales.json` itself is unmodified; `openapi.json` edits were made programmatically and are confined to the `/admin/locales` paths plus one new component, to stay out of the way of the branches concurrently editing that file.

## Summary by CodeRabbit

- **New Features**
  - Administrators can promote eligible locales to the official supported list.
  - Administrators can demote officially supported locales, with safeguards preventing removal of the final official locale.
  - Locale changes can be applied directly or submitted as change requests, depending on deployment configuration.
  - The locale administration view now reports the active curation mode and a clear explanation.

- **Bug Fixes**
  - Promotion requests now validate locale readiness and prevent duplicate or invalid changes.
  - Improved handling of unavailable or misconfigured locale-management resources.

### DIFF
--- a/jsondata/schemas/SupportedLocales.json
+++ b/jsondata/schemas/SupportedLocales.json
@@ -1,0 +1,42 @@
+{
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "title": "SupportedLocales",
+    "type": "object",
+    "description": "The curated promise of which locales this deployment considers officially supported. Not a description of what happens to be on disk: an official locale is served strictly, so a missing readings entry throws rather than degrading quietly. Keyed by calendar, though the General Roman Calendar is currently the only calendar that curates a set.",
+    "additionalProperties": false,
+    "required": ["general_roman_calendar"],
+    "properties": {
+        "general_roman_calendar": {
+            "$ref": "#/definitions/CuratedLocaleSet"
+        }
+    },
+    "definitions": {
+        "CuratedLocaleSet": {
+            "type": "object",
+            "additionalProperties": false,
+            "required": ["official"],
+            "properties": {
+                "official": {
+                    "type": "array",
+                    "description": "The locales officially supported for this calendar. Never empty: an empty list is indistinguishable from an unreadable file, and SupportedLocales::official() would silently substitute its built-in fallback for it.",
+                    "minItems": 1,
+                    "uniqueItems": true,
+                    "items": {
+                        "$ref": "./CommonDef.json#/definitions/Locale"
+                    }
+                },
+                "candidates": {
+                    "type": "object",
+                    "description": "Locales that have resources but are not officially supported, each mapped to the prose reason it has not been promoted. A locale must not appear both here and in `official`; a promotion drops its candidate note.",
+                    "propertyNames": {
+                        "$ref": "./CommonDef.json#/definitions/Locale"
+                    },
+                    "additionalProperties": {
+                        "type": "string",
+                        "minLength": 1
+                    }
+                }
+            }
+        }
+    }
+}
--- a/jsondata/schemas/openapi.json
+++ b/jsondata/schemas/openapi.json
@@ -4756,6 +4756,123 @@
         }
       }
     },
+    "/admin/locales/{locale}/promote": {
+      "post": {
+        "tags": [
+          "Admin - Supported Locales"
+        ],
+        "security": [
+          {
+            "BearerAuth": []
+          },
+          {
+            "CookieAuth": []
+          }
+        ],
+        "summary": "Promote a locale to officially supported",
+        "operationId": "adminPromoteSupportedLocale",
+        "description": "Adds the locale to `general_roman_calendar.official` in jsondata/supportedLocales.json, through the same write seam every other source-data edit uses: a reviewable change request where SOURCEDATA_CHANGE_REQUESTS is configured, a direct disk write where it is not. See the `curation` object of GET /admin/locales for which of the two this deployment will do.\n\nTakes no request body.\n\nGated on the readiness probes of GET /admin/locales/{locale}, and the gate is NOT bypassable — there is no force parameter and no role that skips it. Promotion is not a label: an official locale is served strictly, so `ReadingsMap::getReadings()` throws on a missing readings entry instead of degrading quietly. Promoting an unready locale therefore converts silent gaps into 500s on a calendar the API advertises as supported.\n\nA promotion also drops the locale's note from `candidates`, whose prose says why it was not yet official and is false once it is.\n\nRestricted to global admins. Holding `admin` on `general_roman_calendar:supported_locales` does not open the endpoint; it decides whether the resulting change request is auto-approved or waits for a reviewer.",
+        "parameters": [
+          {
+            "name": "locale",
+            "in": "path",
+            "required": true,
+            "description": "The locale to promote. Must be one of the candidates returned by GET /admin/locales.",
+            "schema": {
+              "type": "string"
+            },
+            "example": "hr"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/SupportedLocaleCurationResponse"
+                }
+              }
+            }
+          },
+          "401": {
+            "$ref": "#/components/responses/Unauthorized401"
+          },
+          "403": {
+            "$ref": "#/components/responses/Forbidden403"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound404"
+          },
+          "409": {
+            "$ref": "#/components/responses/Conflict409"
+          },
+          "422": {
+            "$ref": "#/components/responses/UnprocessableEntity422"
+          },
+          "503": {
+            "$ref": "#/components/responses/ServiceUnavailable503"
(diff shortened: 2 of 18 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
@coderabbitai review

Issue #943, opened 2026-08-30
Add sanctorale write routes (PUT/PATCH /missals/{missal_id}) through the SourceDataWriter seam


JohnRDOrazio:
## Summary

`/missals` and `/missals/{missal_id}` are `GET`-only, so the sanctorale — the largest body of curated source data in the project — is the only major source-data resource with no write path. Meanwhile `assets/js/missals-editor.js` in the frontend has been issuing `PUT` requests to these routes for some time, which is why its 405 branch is the only branch that ever runs (LiturgicalCalendarFrontend#503).

Verified against `development` @ `0f57fdd7`.

## Current state

```text
/missals                 [GET]
/missals/{missal_id}     [GET]
/decrees                 [GET]
/decrees/{decree_id}     [GET, PUT, PATCH, DELETE]   ← the pattern to follow
```

Every other source-data write already routes through `src/Services/SourceData/` — `SourceDataWriter`, `DiskSourceDataWriter`, `ChangeRequestSourceDataWriter`, `SourceDataPublisherFactory` — via the `Handlers/Concerns/WritesSourceData.php` concern. Sanctorale writes must use the same seam so they inherit change-request queueing, `disposition` reporting and publication for free, rather than growing a second write path.

## What a sanctorale write has to maintain

This is the part that makes it more than "add PUT to the handler". An event is spread across three co-located files that must stay consistent, all keyed on `event_key`:

```text
missals/{missal}/{missal}.json          structure: month, day, event_key, grade, common, calendar, color
missals/{missal}/i18n/{locale}.json     one name per event_key, per locale (14 for 1970)
lectionary/sanctorum/{locale}.json      rite-level readings per event_key (6 locales)
missals/{missal}/lectionary/{locale}.json   per-missal readings, national missals only
```

`DecreesHandler` already enforces the equivalent invariants for decrees, and its solutions should be lifted rather than re-derived:

- **`DecreesHandler.php:762`** — *"FINDING 3: reject event_key changes — orphans i18n/lectionary entries permanently."* The same guard is needed here; renaming a key through an edit endpoint silently orphans every sidecar entry.
- **`DecreesHandler.php:837`** — `removeKeyFromLocaleFiles($eventKey, …)` on delete, with GC across both the i18n and lectionary folders (`:838`).

[...]

Issue #940, opened 2026-08-30
Ambrosian propriumdesanctis.json breaks the {dir}/{dir}.json missal naming convention


JohnRDOrazio:
## Summary

Every Roman missal directory contains a JSON file named after the directory — `{dir}/{dir}.json`. The Ambrosian sanctorale breaks that: the directory is `propriumdesanctis_2024` but the file inside is `propriumdesanctis.json`. Any consumer that enumerates missals by convention resolves the wrong path for exactly one missal in the tree.

Verified against `development` @ `0f57fdd7`.

## The inconsistency

```text
rite  dir                         file(s)
────────────────────────────────────────────────────────────
roman      propriumdesanctis_1970      propriumdesanctis_1970.json      ✓
roman      propriumdesanctis_2002      propriumdesanctis_2002.json      ✓
roman      propriumdesanctis_2008      propriumdesanctis_2008.json      ✓
roman      propriumdesanctis_IT_1983   propriumdesanctis_IT_1983.json   ✓
roman      propriumdesanctis_US_2011   propriumdesanctis_US_2011.json   ✓
roman      propriumdetempore           propriumdetempore.json           ✓
ambrosian  propriumdetempore           propriumdetempore.json           ✓
ambrosian  propriumdesanctis_2024      propriumdesanctis.json           ✗  ← unsuffixed
```

Seven of eight follow `{dir}/{dir}.json`. The eighth does not, and it is the only one where the directory carries an edition-year suffix that the filename drops.

## Why it matters

The convention is load-bearing for anything that discovers missals by walking the tree rather than reading a manifest. A resolver written against the observable rule — join the directory name to itself — returns a path that does not exist for the Ambrosian sanctorale, and the failure is a silent "missal has no data" rather than a loud error.

This is not hypothetical: the sanctorale editor being designed in LiturgicalCalendarFrontend#503 enumerates missals across both rites, and had to special-case this one directory.

## Options

1. **Rename the file to `propriumdesanctis_2024.json`** — restores the convention, and is the option that needs no code anywhere. Requires updating whatever currently reads the unsuffixed name.
2. **Rename the directory to `propriumdesanctis`** — also self-consistent, but discards the edition year from the path, which the Roman tree uses to distinguish editions. Not recommended.
3. **Add a manifest** and stop relying on the naming convention at all — the most robust, but much larger than the problem.

Option 1 is the smallest change that makes the tree uniform.

## Also worth deciding while here

[...]

Issue #939, opened 2026-08-30
StIsidore is one event_key for two different saints, so they contend for one lectionary entry


JohnRDOrazio:
## Summary

`StIsidore` is used as the `event_key` for **two different saints**. Because the sanctorale lectionary is keyed on `event_key` alone, the two saints necessarily contend for one set of readings, and the US calendar gets a wrong answer whichever tier wins resolution.

Verified against `development` @ `0f57fdd7`.

## The collision

`rite/roman/missals/propriumdesanctis_1970/propriumdesanctis_1970.json` — **Isidore of Seville**:

```json
{"month": 4, "day": 4, "event_key": "StIsidore", "grade": 2,
 "common": ["Pastors:For a Bishop", "Doctors"], "calendar": "GENERAL ROMAN", "color": ["white"]}
```

`i18n/en.json` confirms it: `"Saint Isidore, Bishop and Doctor of the Church"`.

`rite/roman/missals/propriumdesanctis_US_2011/propriumdesanctis_US_2011.json` — **Isidore the Farmer**:

```json
{"month": 5, "day": 15, "event_key": "StIsidore", "grade": 2, "grade_display": null,
 "common": ["Holy Men and Women:For One Saint"], "calendar": "US", "color": ["white"]}
```

`i18n/en_US.json` gives only `"Saint Isidore"`. Different date (May 15 vs April 4), different common, different person.

This is the **only** shared key between the two files — the other 13 US_2011 entries are new. Every other missal delta is fully disjoint from 1970.

## Why the lectionary makes it a correctness bug, not just a wart

There are two lectionary tiers:

1. `rite/roman/lectionary/sanctorum/{locale}.json` — rite-level, 6 locales (`en`, `fr`, `hr`, `it`, `la`, `nl`)
2. `rite/roman/missals/{missal}/lectionary/{locale}.json` — per-missal, present **only** for the two national missals `IT_1983` and `US_2011`

Both tiers currently define `StIsidore`:

```jsonc
// rite/roman/lectionary/sanctorum/en.json  — these are Seville's
{"first_reading": "2 Corinthians 4:1-2,5-7", "responsorial_psalm": "Psalm 37",
 "gospel_acclamation": "John 15:5", "gospel": "Luke 6:43-45"}

// missals/propriumdesanctis_US_2011/lectionary/en_US.json  — placeholder, all empty
[...]

PR #948, opened 2026-08-31
feat(missals,lectionary): aggregated missal i18n and sanctorale lectionary read routes

### LINKED ISSUE #941: GET /missals/{missal_id} exposes only one locale's names, so translations cannot be viewed or edited side by side
## Summary

`GET /missals/{missal_id}` merges the sanctorale name for exactly **one** negotiated locale into each row. A client that needs to show or edit the translations side by side has no way to ask for them, and must issue one request per locale — 14 for `propriumdesanctis_1970` — with no guarantee it has found them all.

Verified against `development` @ `0f57fdd7`.

## Current behaviour

`MissalsHandler.php:214-227` picks a single locale, then folds it into the rows:

```php
$locale     = RomanMissal::isLatinMissal($missalId)
            ? ( in_array($this->params->baseLocale, $missalMetadata->locales) ? … : LitLocale::LATIN_PRIMARY_LANGUAGE )
            : ( in_array($this->params->Locale, $missalMetadata->locales) ? $this->params->Locale : $missalMetadata->locales[0] );
$i18nFile   = RomanMissal::getSanctoraleI18nFilePath($missalId) . $locale . '.json';
$i18nObj    = Utilities::jsonFileToObject($i18nFile);
$missalRows = Utilities::jsonFileToObjectArray($missalJsonFile);

foreach ($missalRows as $idx => $row) {
    $key = $row->event_key;
    if (property_exists($i18nObj, $key)) {
        $missalRows[$idx]->name = $i18nObj->{$key};
    }
}
```

The response is therefore lossy by construction: one name per row, no indication of which locale it came from, and no way to see that a key is translated in 12 locales but missing in 2.

Note also that `openapi.json` documents the `200` for this route as the `Missal` schema — which is the **metadata** shape (`missal_id, name, region, year_published, year_limits, locales, api_path`). The route actually returns an array of sanctorale rows. That is separate schema drift worth fixing alongside.

## Why it matters

The sanctorale editor in LiturgicalCalendarFrontend#503 needs to present every translation of an event at once — that is the point of the screen. Fanning out N requests per missal to reassemble something the API already has on disk is both slow and racy, and it cannot distinguish "this locale has no translation for this key" from "this locale file does not exist".

`DecreesHandler.php:324` already does the right thing for the sibling resource:

> Aggregate the readings for an event across every lectionary locale file

There is a working precedent in this codebase for aggregating a per-locale sidecar across all locales. The missals path should use it.

## Proposal

Add an opt-in aggregated shape rather than changing the default (the current single-locale response is the right default for calendar consumers):

- `GET /missals/{missal_id}?locales=all` returning, per row, a `names` map of `{locale: string}` instead of a single flattened `name`; **or**
- a sidecar route `GET /missals/{missal_id}/i18n` returning `{locale: {event_key: name}}` for every locale in `$missalMetadata->locales`

The sidecar is probably cleaner: it leaves the existing row shape untouched, and it is the natural write target for the corresponding PUT/PATCH.

Either way the response must distinguish a **missing key** from an **empty string**, since the empty-string placeholder is already a real convention in this data (see `propriumdesanctis_US_2011/lectionary/en_US.json`, where `StIsidore` has four empty reading fields).

## Context

One of three routes the sanctorale editor needs. Companions: the sanctorale lectionary read route, and sanctorale writes. Refs LiturgicalCalendarFrontend#503.

### LINKED ISSUE #942: The sanctorale lectionary has no read route, so readings cannot be viewed as an editable resource
## Summary

The sanctorale lectionary is real, curated, on-disk data that **no route exposes**. `CalendarHandler` and `DecreesHandler` read it internally to attach readings to computed events, but there is no way for a client to read the readings for an `event_key` as an editable resource.

Verified against `development` @ `0f57fdd7`.

## The data exists; the route does not

Two tiers:

1. **Rite-level** — `JsonData::LECTIONARY_SAINTS_FOLDER` → `jsondata/sourcedata/rite/roman/lectionary/sanctorum/{locale}.json`, currently 6 locales (`en`, `fr`, `hr`, `it`, `la`, `nl`), 210 keys in `en`.
2. **Per-missal** — `JsonData::MISSAL_LECTIONARY_FOLDER` / `MISSAL_LECTIONARY_FILE`, resolved by `RomanMissal::getLectionaryFilePath()`. `RomanMissal::$lectionaryPath` maps only the two **national** missals:

    ```php
    self::USA_EDITION_2011   => '/propriumdesanctis_US_2011/lectionary/',
    self::ITALY_EDITION_1983 => '/propriumdesanctis_IT_1983/lectionary/',
    ```

    Editio typica missals have none — `CalendarHandler.php:1631` notes "Latin edition Roman Missals do not have lectionary data" — and there is no Ambrosian lectionary at all (`CalendarHandler.php:1056`).

`ls src/Handlers/` has no lectionary handler, and `openapi.json` contains no path or schema whose name matches `lectionar`. The only readings a client can obtain today are the ones already baked into a computed `/calendar` response, which is the wrong granularity for editing and gives no access to locales the caller did not request.

## Why this blocks work

LiturgicalCalendarFrontend#503 is rebuilding the missals editor as a sanctorale editor whose whole premise is showing an event's structure, its names, and its readings together. The first two are readable today; **the third is not readable at all**, so even a strictly read-only version of that screen cannot be built.

## Proposal

Mirror `DecreesHandler`, which has already solved this exact shape for the sibling resource:

- `DecreesHandler.php:249-251` reads `lectionary/decrees/{locale}.json`
- `DecreesHandler.php:324-333` aggregates readings for an event **across every lectionary locale file**

Suggested surface:

```text
GET /lectionary/sanctorale/{event_key}   → {locale: {first_reading, responsorial_psalm,
                                                     gospel_acclamation, second_reading?, gospel}}
GET /lectionary/sanctorale               → the full map, or a paginated index
```

Requirements the shape must satisfy:

- **Resolve both tiers, and say which one answered.** For an `event_key` defined in a national missal, the per-missal file and the rite-level file may both carry an entry. A client editing readings must know which file it is looking at.
- **Distinguish absent from empty.** The empty-string placeholder is an established convention here — `propriumdesanctis_US_2011/lectionary/en_US.json` has `StIsidore` with all four fields `""`. "No readings curated yet" and "no entry in this locale" are different states and both are legitimate.
- **Degrade cleanly for rites with no lectionary.** Ambrosian has no lectionary directory; the answer should be a well-formed "not available for this rite", not a 500 or an empty object indistinguishable from "nothing curated".

## Context

One of three routes the sanctorale editor needs, and the one that blocks even the read-only phase. See also the `StIsidore` collision (two saints sharing one lectionary entry) filed separately. Refs LiturgicalCalendarFrontend#503.

JohnRDOrazio:
Two read routes the sanctorale editor needs, designed as one surface rather than two bolted-on endpoints.

Unblocks **LiturgicalCalendarFrontend#503** (the sanctorale editor: an event's structure, its names and its readings shown together — the last two were not readable at all), and precedes **#943**, the corresponding writes.

Closes #941
Closes #942

## #941 — one locale's names is a lossy answer

`GET /missals/{missal_id}` folds a single negotiated locale's name into each sanctorale row: one name per row, no note of which locale it came from, and no way to see that a key is translated in twelve locales and left empty in two.

**Chosen surface: the sidecar `GET /missals/{missal_id}/i18n`**, not a `?locales=all` variant of the row response — the reason the issue gave, and it holds up: it leaves the row shape untouched so no existing consumer has to learn a second spelling of `name`, and it is the natural target for #943, which writes *locale files*, not rows. The current single-locale row response stays the default, which is right for calendar consumers.

```json
{
  "missal_id": "EDITIO_TYPICA_2008",
  "locales": ["de","en","es","fr","hr","hu","id","it","la","nl","pl","pt","sk","vi"],
  "event_keys": ["StPioPietrelcina", "OurLadyOfGuadalupe", "JuanDiego"],
  "i18n":     { "hu": { "StPioPietrelcina": "Pietralcinai Szent Pio, pap", "OurLadyOfGuadalupe": "", "JuanDiego": "" }, "…": {} },
  "coverage": { "StPioPietrelcina": { "translated": ["en","es","fr","hr","hu","id","it","la","nl","pl","sk","vi"], "empty": ["de","pt"], "missing": [] } }
}
```

`i18n` is each locale file **verbatim**, which is what makes absent-vs-empty survive the round trip: a key **absent** from a locale's map has no entry in that file, a key mapped to the **empty string** has an entry whose translation is not written yet. `coverage` states that three-way split per `event_key` so a client need not derive it from a fourteen-locale map — that derivation being exactly the work the issue says clients should not have to do.

### Schema drift, also fixed

`openapi.json` documented this route family's `200` for `/missals/{missal_id}` as `LitCalMissalsPath.json#/definitions/Missal` — the *metadata* shape of a `/missals` index row (`missal_id, name, region, year_published, year_limits, locales, api_path`). The route returns an array of sanctorale rows. It now points at a new `LitCalMissalSanctoralePath.json`, whose item is `PropriumDeSanctis.json#/definitions/PropriumDeSanctis` (the response *is* the source rows with the negotiated `name` folded in — `name` was already optional there), and a test validates every missal's real response against it so the two cannot drift apart again.

## #942 — the sanctorale lectionary had no read route at all

```text
GET /lectionary/{rite}/sanctorale              → index: every event_key, and which sources carry it
GET /lectionary/{rite}/sanctorale/{event_key}  → every locale's readings, per source
```

Built on `DecreesHandler`'s across-every-locale aggregation (`DecreesHandler.php:324-333`) rather than re-derived.

**Which tier answered.** The readings live in two tiers — the rite-wide corpus (`rite/roman/lectionary/sanctorum/{locale}.json`, 6 locales, 211 keys) and one folder per national Missal (`US_2011`, `IT_1983`) — and an `event_key` may be carried by both. Every entry names its `tier` and `source_id` instead of being merged into one answer, so a client editing readings knows which file it is looking at. The missal tier is enumerated from `RomanMissal`'s explicit lectionary-path map, **not** by globbing, so it carries no dependency on the `{dir}/{dir}.json` folder-naming convention #940 is fixing.

**Absent vs. empty.** Per source: `locales` (every locale the source declares) splits into `locales_with_entry` / `locales_without_entry`, and `locales_with_empty_entry` — a subset of the first — flags the all-empty-string placeholder. `entries` carries the source entries verbatim, empty strings included. Real data exercises both: some `event_key` in the rite corpus is present in five locale files and absent from the sixth, while many are present everywhere and wholly empty.

**A rite with no lectionary.** Ambrosian answers `200`:

```json
{ "rite": "ambrosian", "section": "sanctorale", "lectionary_available": false,
  "sources": [], "litcal_lectionary": [],
  "message": "No sanctorale lectionary data is defined for the ambrosian rite." }
```

…and the same for the item route with `"readings": []`. An unknown `event_key` under a rite that *does* have a lectionary is a `404`, so "this rite has none" and "nobody curated this event" never collide — neither a 500 nor an empty object indistinguishable from "nothing curated".

`lectionary` joins `calendar`/`events`/`data` as a rite-carrying route (`Router::extractRiteSegment()`, `Router::canonicalRiteUrl()`), so the bare spelling is deprecated and advertises the canonical explicit one; `OpenApiCanonicalFormTest::RITE_ROUTES` is widened to cover it, and the new paths pass its Link-header and deprecation invariants.

## Tests

`phpunit_tests/Handlers/LectionaryHandlerTest.php` (14) and `phpunit_tests/Handlers/MissalsHandlerI18nTest.php` (8), plus three `Router::extractRiteSegment()` cases.

Fixtures are **discovered** by reading the source files with `json_decode` — an oracle independent of the handler's own code path — rather than naming an `event_key`. Two reasons: a hard-coded key is a hostage to the rename landing in parallel under #939, and a discovered fixture proves the property holds of the data as it is rather than of the one row somebody remembered. The tests fail loudly (`assertNotNull` / `fail`) if the corpus ever stops containing an instance of a distinction, rather than passing vacuously.

One state has no instance in the committed corpus: every missal locale file currently carries an entry for every row, so `coverage.missing` is empty everywhere. It is exercised anyway, against a throwaway source tree assembled in `sys_get_temp_dir()` with `Router::$apiFilePath` re-pointed at it — nothing under `jsondata/` is touched. An untested branch of a three-way split is a coin toss, and this is the branch the split exists for.

New OUTPUT schemas, all declaring `SchemaRole::OUTPUT` via `LitSchema::role()`: `LitCalLectionaryPath.json`, `LitCalMissalTranslationsPath.json`, `LitCalMissalSanctoralePath.json`. Every handler response is validated against them in-process.

## Verification

| command | result |
|---------------------------|---------------------------------------------------------|
| `composer lint`           | exit 0 |
| `composer analyse`        | `[OK] No errors` (PHPStan level 10) |
| `composer parallel-lint`  | 811 files, no syntax error |
| `composer lint:jsondata`  | `OK — 23 file(s) match their canonical encoding` |
| `composer lint:openapi`   | `Your API description is valid.` |
| `composer test:quick`     | `Tests: 4317, Assertions: 188309, Warnings: 1, Skipped: 11` — 0 failures |

The single warning is `DiskSourceDataWriterTest`'s `mkdir(): File exists`; both files involved are byte-identical to `origin/development` and untouched here.

## Summary by CodeRabbit

* **New Features**
  * Added lectionary endpoints for browsing sanctorale sources and retrieving event readings by rite and event key.
  * Added missal translation endpoints with locale-specific entries and translated, empty and missing coverage details.
  * Added Roman and Ambrosian rite paths, including clear responses when a lectionary is unavailable.

* **Documentation**
  * Expanded OpenAPI documentation and response schemas for lectionary and missal endpoints.
  * Updated missal responses to return sanctorale data with the selected locale.

### DIFF
--- a/jsondata/schemas/LitCalLectionaryPath.json
+++ b/jsondata/schemas/LitCalLectionaryPath.json
@@ -1,0 +1,165 @@
+{
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "title": "LitCalLectionaryPath",
+    "description": "A response of the `/lectionary` path. The section index and a single event's readings are different shapes, so the root is a choice between the two; their required properties are disjoint, so exactly one branch ever matches. This schema describes OUTPUT: the readings it carries are the source entries verbatim, which is why they are validated against CommonDef's SourceReadings rather than against Readings.",
+    "oneOf": [
+        { "$ref": "#/definitions/LectionarySectionIndex" },
+        { "$ref": "#/definitions/LectionaryEventReadings" }
+    ],
+    "definitions": {
+        "Tier": {
+            "type": "string",
+            "title": "Lectionary Tier",
+            "description": "which layer of the corpus a lectionary source belongs to: `rite` for the rite-wide sanctorale lectionary, `missal` for a lectionary that belongs to one Roman Missal edition",
+            "enum": [ "rite", "missal" ]
+        },
+        "SourceId": {
+            "title": "Lectionary Source ID",
+            "description": "identifies the source within its tier: a rite identifier for the `rite` tier, a Missal ID for the `missal` tier",
+            "oneOf": [
+                { "$ref": "./CommonDef.json#/definitions/Rite" },
+                { "$ref": "./CommonDef.json#/definitions/MissalID" }
+            ]
+        },
+        "Locales": {
+            "type": "array",
+            "description": "locales, identified by the basename of the locale file",
+            "items": { "$ref": "./CommonDef.json#/definitions/Locale" },
+            "uniqueItems": true
+        },
+        "SourceRef": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "tier": { "$ref": "#/definitions/Tier" },
+                "source_id": { "$ref": "#/definitions/SourceId" }
+            },
+            "required": [ "tier", "source_id" ]
+        },
+        "SourceSummary": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "tier": { "$ref": "#/definitions/Tier" },
+                "source_id": { "$ref": "#/definitions/SourceId" },
+                "locales": { "$ref": "#/definitions/Locales" },
+                "event_key_count": {
+                    "type": "integer",
+                    "minimum": 0,
+                    "description": "how many distinct event_key values this source carries across all of its locales"
+                }
+            },
+            "required": [ "tier", "source_id", "locales", "event_key_count" ]
+        },
+        "LectionarySectionIndex": {
+            "type": "object",
+            "title": "Lectionary Section Index",
+            "description": "the response of `GET /lectionary/{rite}/sanctorale`",
+            "additionalProperties": false,
+            "properties": {
+                "rite": { "$ref": "./CommonDef.json#/definitions/Rite" },
+                "section": {
+                    "type": "string",
+                    "enum": [ "sanctorale" ]
+                },
+                "lectionary_available": {
+                    "type": "boolean",
+                    "description": "false when the rite has no lectionary data at all, which is a different state from a lectionary that exists but curates nothing for a given event"
+                },
+                "sources": {
+                    "type": "array",
+                    "items": { "$ref": "#/definitions/SourceSummary" }
+                },
+                "litcal_lectionary": {
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "additionalProperties": false,
+                        "properties": {
+                            "event_key": { "$ref": "./CommonDef.json#/definitions/EventKey" },
+                            "sources": {
+                                "type": "array",
+                                "description": "every source that carries an entry for this event_key, in resolution order",
+                                "items": { "$ref": "#/definitions/SourceRef" },
+                                "minItems": 1
+                            },
+                            "api_path": { "type": "string" }
+                        },
+                        "required": [ "event_key", "sources", "api_path" ]
+                    }
+                },
+                "message": {
+                    "type": "string",
+                    "description": "present only when `lectionary_available` is false, saying so in words"
+                }
+            },
+            "required": [ "rite", "section", "lectionary_available", "sources", "litcal_lectionary" ]
+        },
+        "LectionaryEventReadings": {
+            "type": "object",
+            "title": "Lectionary Readings for one Event",
+            "description": "the response of `GET /lectionary/{rite}/sanctorale/{event_key}`",
+            "additionalProperties": false,
+            "properties": {
+                "rite": { "$ref": "./CommonDef.json#/definitions/Rite" },
+                "section": {
+                    "type": "string",
+                    "enum": [ "sanctorale" ]
+                },
+                "event_key": { "$ref": "./CommonDef.json#/definitions/EventKey" },
+                "lectionary_available": {
+                    "type": "boolean",
+                    "description": "false when the rite has no lectionary data at all; an unknown event_key under a rite that does have one is a 404 instead, so the two states never collide"
+                },
+                "readings": {
+                    "type": "array",
+                    "description": "one entry per source that carries this event_key, in resolution order; a source that carries nothing for it is omitted rather than listed empty",
+                    "items": {
+                        "type": "object",
+                        "additionalProperties": false,
+                        "properties": {
+                            "tier": { "$ref": "#/definitions/Tier" },
+                            "source_id": { "$ref": "#/definitions/SourceId" },
+                            "locales": {
+                                "$ref": "#/definitions/Locales",
+                                "description": "every locale this source declares, i.e. every locale file present in its folder"
+                            },
+                            "locales_with_entry": {
+                                "$ref": "#/definitions/Locales",
+                                "description": "the declared locales whose file has an entry for this event_key; exactly the keys of `entries`"
+                            },
+                            "locales_without_entry": {
+                                "$ref": "#/definitions/Locales",
+                                "description": "the declared locales whose file has no entry at all for this event_key"
+                            },
+                            "locales_with_empty_entry": {
+                                "$ref": "#/definitions/Locales",
+                                "description": "a subset of `locales_with_entry`: locales whose entry is the all-empty-string placeholder, meaning the structure is in place but the readings are not written yet"
+                            },
+                            "entries": {
+                                "type": "object",
+                                "description": "the source entries verbatim, keyed by locale — empty strings included, so nothing is flattened away",
+                                "propertyNames": { "$ref": "./CommonDef.json#/definitions/Locale" },
+                                "additionalProperties": { "$ref": "./CommonDef.json#/definitions/SourceReadings" }
+                            }
+                        },
+                        "required": [
+                            "tier",
+                            "source_id",
+                            "locales",
+                            "locales_with_entry",
+                            "locales_without_entry",
+                            "locales_with_empty_entry",
+                            "entries"
+                        ]
+                    }
+                },
+                "message": {
+                    "type": "string",
+                    "description": "present only when `lectionary_available` is false, saying so in words"
+                }
+            },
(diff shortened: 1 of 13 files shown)


JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
@coderabbitai review

JohnRDOrazio:
Valid — fixed in 5cbf0053. One correction to the framing, since it changes where the blame sits.

**Nothing was removed.** `origin/development` also declares only `200` and `404` on this operation:

```
$ git show origin/development:jsondata/schemas/openapi.json | jq '.paths["/missals/{missal_id}"].get.responses | keys'
["200","404"]
```

So this is a pre-existing gap, not a regression introduced here. The finding stands on its merits though — the
`503` is genuinely reachable on this operation, from two sites:

- `MissalsHandler::handle():143` — index build failure, wrapped as `ServiceUnavailableException` to surface a
  503 rather than a generic 500, and it runs for every missals operation.
- `MissalsHandler::handleGetRequest():184` — null index, before the path-parameter branch.

And the asymmetry was real: `/missals` and the new `/missals/{missal_id}/i18n` both declared it while
`/missals/{missal_id}` did not, so the three disagreed about one failure mode.

Fixed here rather than deferred because it is the **same class of drift this PR already corrects on this exact
operation** — the `200` pointed at `LitCalMissalsPath.json#/definitions/Missal`, the index-metadata row, rather
than the sanctorale rows the route actually returns. Leaving the `503` wrong while fixing the `200` would have
been an odd place to stop.

Now consistent:

```
/missals                  -> ["200","404","503"]
/missals/{missal_id}      -> ["200","404","503"]
/missals/{missal_id}/i18n -> ["200","404","503"]
```

Edited programmatically and asserted to round-trip byte-exactly through
`json.dumps(indent=2, ensure_ascii=False)`; zero `\uXXXX` escapes. `composer lint:jsondata` OK (23 files),
`composer lint:openapi` exit 0 ("Woohoo! Your API description is valid."), `composer lint` clean,
`composer analyse` `[OK] No errors`, and `phpunit_tests/Schemas` plus both new handler suites
`OK (667 tests, 3501 assertions)`.

PR #947, opened 2026-08-31
fix(missals): split the StIsidore event_key, restore the missal naming convention, and gate both in CI

JohnRDOrazio:
Two source-data consistency fixes in the missals tree, plus the CI guard that keeps
either from recurring silently.

Closes #939
Closes #940

## #939 — `StIsidore` was one `event_key` for two different saints

`propriumdesanctis_1970` means **Isidore of Seville** (4 April, "Bishop and Doctor of the
Church"); `propriumdesanctis_US_2011` meant **Isidore the Farmer** (15 May) under the very
same key.

The issue framed this as a lectionary collision. It is that, but it is also worse than
that: missals are delta layers merged by `event_key`, and
`LiturgicalEventCollection::addLiturgicalEvent()` is keyed on that string alone, so the US
row did not sit beside Seville's — it **overwrote** it. The US national calendar carried no
4 April Isidore at all. And because the sanctorale lectionary is keyed on `event_key` too,
with the per-missal tier loading over the rite-level one, the US missal's all-empty
placeholder erased Seville's readings along with him.

Verified in-process against `/calendar/nation/US/2025` (4 April 2025 is a free Lenten
weekday, so nothing else masks the loss):

```text
before: StIsidore        2025-05-15  [ US ] Saint Isidore                          first_reading=""
after:  StIsidore        2025-04-04  Saint Isidore, Bishop and Doctor of the Church first_reading="2 Corinthians 4:1-2,5-7"
        StIsidoreFarmer  2025-05-15  [ US ] Saint Isidore                          first_reading=""
```

The US row is now `StIsidoreFarmer`, renamed in the structure file and in all three
sidecars. `propriumdesanctis_1970/**` and `rite/roman/lectionary/sanctorum/*.json` are
untouched — every `StIsidore` there legitimately means Seville.

**The two pre-rename checks the issue asked for:**

- *Does anything reference `StIsidore` expecting the US meaning?* No national or diocesan
  calendar definition, no decree, no test source fixture. The only place carrying the US
  sense was the generated golden-master fixture `nation-US-2023`, regenerated here. (The
  Ambrosian tree already disambiguated its own `StIsidoreOfSeville`.)
- *Does an `event_key` uniqueness check belong in CI?* Yes — added, see below.

⚠️ **`StIsidore` on 15 May is a renamed `event_key` in the `US` calendar response.** A
client that pins that key sees it move to `StIsidoreFarmer`, and sees a new `StIsidore`
appear on 4 April in years where it is not suppressed.

[...]

Issue #922, opened 2026-08-30
Routes and WebSocket suites give no signal when a stale container holds port 8000: 131 phantom failures


JohnRDOrazio:
## Summary

The `Routes/*` and `WebSocket/*` suites talk to a live server over `127.0.0.1:8000` but never check that
the server answering is the code under test — or a working server at all. When a stale container holds
that port, **131 tests fail** with no indication that the cause is environmental, and the failures look
exactly like a regression in whatever branch is checked out.

## What was observed

`ss -ltn` shows `127.0.0.1:8000` LISTENing, and every path 404s:

```text
/                    404
/calendars           404
/health              404
```

But something is genuinely serving — the 404 carries a PHP signature:

```text
HTTP/1.1 404 Not Found
X-Powered-By: PHP/8.4.20
Content-Type: text/html; charset=UTF-8
```

**`PHP/8.4.20`, while the local CLI is `PHP 8.4.23`.** So the responder is not a local `php -S` (none
appears in `ps`, and `stop-server.sh` reports only a stale `server.pid`); it is a containerised PHP of a
different patch version, still holding the forwarded port.

Meanwhile the Docker daemon is unreachable from this WSL distro:

```text
/var/run/docker.sock                       does not exist
docker version  -> failed to connect to the docker API at unix:///var/run/docker.sock
/usr/bin/docker (CLI 29.7.2)               present
/mnt/wsl/docker-desktop                    present, dated at this boot
```

## Likely cause

A Docker Desktop update or restart left the WSL integration inconsistent: the socket bind-mount into this
distro is gone, while the container (or at least its port proxy) survives and keeps `127.0.0.1:8000`. The
result is the worst of both worlds — the port is occupied so a local `start-server.sh` cannot bind it, and
`docker compose restart litcal-api` cannot be run to fix the container because the socket is missing.

This is an environment fault, not a code fault. **The issue being filed is that the test suite cannot tell
the difference**, and neither can the developer reading its output.

## Why it is worth a guard

Across six concurrent branches this produced an identical 131-test failure set every time. Establishing
that the failures were environmental required running the same suites against a clean `development`
checkout at the same moment and diffing the failing-test names — the set difference was empty. That is the
right technique, but it is a lot of ceremony to rediscover, and the raw counts are actively misleading:
they also fluctuate run to run on an unchanged tree (21-34 errors / 110-164 failures were seen), because
several of these tests share one Postgres and one port.

## Suggested fix

[...]

PR #950, opened 2026-08-31
feat(missals): sanctorale entry writes through the SourceDataWriter seam

### LINKED ISSUE #943: Add sanctorale write routes (PUT/PATCH /missals/{missal_id}) through the SourceDataWriter seam

JohnRDOrazio:
The sanctorale was the last major body of curated source data with no write path — `/missals` and `/missals/{missal_id}` were `GET`-only, which is why `assets/js/missals-editor.js`'s 405 branch is the only one that ever ran (LiturgicalCalendarFrontend#503).

Closes #943. **Builds on #948** (the sanctorale read routes, now merged) and should land after it — see *Branch base* at the bottom.

## The surface

```text
PUT    /missals/{missal_id}/{event_key}   create an entry
PATCH  /missals/{missal_id}/{event_key}   partial update
DELETE /missals/{missal_id}/{event_key}   remove it, and collect its sidecars
```

**The entry, not the Missal, carries the key in the URL, and that is load-bearing rather than cosmetic.** The issue asks for the decrees guard — *"FINDING 3: reject event_key changes — orphans i18n/lectionary entries permanently"* (`DecreesHandler.php:762`) — and that guard needs a channel to fire on. With the key only in the body there is no way to express a rename and therefore no way to refuse one; with the key in the path, a payload `event_key` that disagrees is a 422 that names the remedy, exactly as `requireValidatedPayload()` does for `decree_id`. `i18n` can never collide with an `event_key` (every key starts upper-case per `CommonDef.json#/definitions/EventKey`), and the sub-resource is excluded explicitly anyway rather than relying on that.

`/missals/{missal_id}` consequently stays read-only, where the Router previously routed PUT/PATCH/DELETE into a `Not yet implemented` stub. A whole-file replace — which is what the frontend currently sends — can express a rename and cannot express which entry it meant, so it is the one shape that must not exist. **The frontend needs a corresponding change**; it will keep seeing a 405 until then, now with `Allow: GET, POST`.

**DELETE is included, and the open question in the issue has a concrete answer.** An `event_key` cannot be renamed, so correcting one means delete-and-recreate — which is precisely how `StIsidore` was split in #939. Without a DELETE that garbage-collects, that correction leaves the name and the readings behind under a key nothing declares, which is the half-finished rename `composer lint:missals` invariant 3 fails on.

## The invariants, and where each one comes from

**Fan-out on create.** The key is added to **every** locale file of the Missal's `i18n/` folder and of its readings tier, with an empty value where the payload supplies none. This is not a convention invented here: all 187 keys of `propriumdesanctis_1970` appear in all fourteen of its locale files — 187 of them empty in `de.json` — and every US_2011 entry appears in both lectionary files, all empty. It is what keeps a locale file a complete, diffable inventory of what still needs translating, and it is what lets `GET /missals/{missal_id}/i18n` distinguish *present-but-untranslated* from *absent* at all. An entry that already exists is **never** overwritten with a placeholder: the rite-level corpus is shared, so creating `StPeterClaver` in US_2011 must not blank the readings the 2002 editio typica already has.

Only files whose content actually changes are staged, so a PATCH that touches no sidecar proposes no sidecar rather than fourteen identical ones for a reviewer to read past.

**`event_key` identity.** The rule is `scripts/lint-missals.php` invariant 2 (from #939), **extracted to `SanctoraleKeyIdentity` rather than restated**, and asked prospectively instead of retrospectively:

> Within a rite, an `event_key` declared by more than one sanctorale Missal must carry the same month/day in every one of them.

So a key another Missal declares on a *different* date is refused (409), and one it declares on the *same* date is allowed — re-declaring across delta layers to change a grade for a national calendar is normal and correct. Checked on PATCH too: moving a date can break the agreement as easily as adding a row. The declarations are read through the same aggregate-aware loader the rest of the handler uses, so a key the submitter created in another Missal a moment ago — nowhere on disk, in queue mode — still collides.

**Delete-time GC.** `removeKeyFromLocaleFiles()` is lifted from `DecreesHandler.php:856`. The Missal's own `i18n/` folder is always collected. The readings folder depends on the tier: a national edition's `lectionary/` is its own, but the rite-level `sanctorum` corpus is shared by every Roman Missal, so a key another Missal still declares keeps its readings. That is `DecreesHandler`'s `$stillReferenced` check applied to the tier that is actually shared, and the response reports which happened via `readings_retained`.

**Readings tier.** The Missal's own `lectionary/` folder when it has one (the national editions), the rite-level `sanctorum` corpus otherwise (the editiones typicae) — the same two tiers `GET /lectionary/{rite}/sanctorale` reports, so readings are written back to the tier they were read from.

**Locale scope.** A sidecar map may only name locales the target folder already has a file for. Adding a locale changes the set `/missals` advertises for a Missal (`RomanMissal::produceMetadata()` globs that folder), so it is a separate act rather than a side effect of writing one saint — and it means every path this handler writes derives from a filename that already exists, never from a request string.

## Separate authorizability: decided, not deferred

The issue asks whether the structure and the i18n/lectionary sidecars should be separately authorizable — "adding a saint and translating an existing saint's name are different acts with plausibly different reviewers". **They are not split here, and this is a decision rather than an omission.**

The reason is that `commitStagedFiles()` takes one `ChangeResource` and produces one batch, and the batch is the mechanism that keeps a row and its fan-out from drifting apart. Splitting the authorization means splitting the batch, at which point a create can be approved while its fourteen name entries are refused — the entry exists in the structure file and in no locale file, which is precisely the inconsistency this issue exists to prevent. The invariant and the authorization boundary have to be the same boundary.

What a real split would take, if it is ever wanted: a new OpenFGA object type (`missal_translation`, say) with its own relations and a model migration; new `AccessRequestRepository` object-type validation and access-request plumbing; admin UI for granting it; and — the substantive part — a way to express "this batch needs two approvals, and neither half may publish alone", which the change-request schema has no notion of today. That is a feature in its own right, not a flag. A *partial* split, where the two are authorized separately but still committed together, would give reviewers the appearance of a boundary with none of the enforcement, which is worse than the single clear one.

`ChangeResource::missal()` mirrors `OpenFgaAuthorizationMiddleware::forMissals()` exactly — editio typica → `general_roman_calendar:{missal_id}`, national edition → `national_calendar:roman/{nation}` — so the object a caller is authorized on is the object the proposal is filed against. The three editiones typicae carrying sanctorale data are already in `AccessRequestRepository::GRC_OBJECT_IDS`, so **no OpenFGA model migration is needed**, and the Router's existing `count($requestPathParts) >= 1` guard already covers the entry path.

## Two findings worth calling out

**`JsonFormatter` is the wrong encoder for this family.** It collapses a simple string array onto one line — `[ "white" ]` — and the sanctorale files spell `common` and `color` expanded. Encoding through it would rewrite every row: adding one entry to `propriumdesanctis_1970` would produce a diff touching all 187, and a reviewer approving that change request could not see what changed. Plain pretty-printing round-trips all five structure files, both lectionary folders and the `sanctorum` corpus **byte-for-byte**, which `testEncodingAnUntouchedSanctoraleFileReproducesItByteForByte()` now pins so it cannot quietly stop being true. (Eight `i18n` files are the exception, in whitespace only — seven carry CRLF, one is indented with two spaces. A write that adds a key to one normalizes it; a file that is not touched is not restaged.)

**`SourceDataSchemaResolver` would have validated the temporale as a sanctorale.** `{missal_folder}/{missal_folder}.json` widens to `missals/[^/]+/[^/]+\.json`, which also matches `missals/propriumdetempore/propriumdetempore.json`. Adding the missal families without the temporale row ahead of them would have turned every temporale change request into a wrong-red — a temporale row has no `month`/`day` at all. `TEMPORALE_FILE` is listed first, and `testEveryCommittedSourceFileTheResolverClaimsAlreadyValidates` now covers the four newly-claimed families against the real corpus.

Also: `.gitignore`'s decrees lockfile glob still pointed at the pre-rite-partition path (`jsondata/sourcedata/decrees/*.lock`), so it has matched nothing since the tree was partitioned. Corrected alongside the new missals entry.

## Tests

`MissalsHandlerWriteTest` (24) and `MissalsQueueModeTest` (4), plus `SanctoraleKeyIdentityTest` (6) and additions to `ChangeResourceTest` and `SourceDataSchemaResolverTest`.

Every write lands in a throwaway shadow of `jsondata/` with `Router::$apiFilePath` repointed at it (`ShadowProjectRootTrait`); nothing under `jsondata/` is touched. **Each guard is exercised by constructing the failing case**, and each asserts two things: that the request is refused, and that an md5 fingerprint of the whole source tree is unchanged afterwards — a guard that threw *after* writing half the fan-out would pass the first assertion alone.

Fixtures are discovered from the corpus (`keyDeclaredOnlyBy()`) rather than naming an `event_key`, and fail loudly if the corpus stops containing an instance of the distinction, so they are not hostage to a rename.

| guard | shown firing by |
|---|---|
| fan-out reaches every locale, empty included | `testCreateFansTheKeyIntoEveryI18nLocaleFileIncludingTheEmptyOnes` — asserts all 14 files individually, not a count |
| readings fan-out, both tiers | `…EveryReadingsLocaleFileOfTheRiteTier` (6 files), `…ANationalEditionWritesItsOwnLectionaryTier` (2 files, and the rite corpus untouched) |
| `event_key` mutation refused | `testPatchRefusesAPayloadEventKeyThatDisagreesWithTheUrl` |
| date collision refused | `testCreateIsRefusedWhenAnotherMissalDeclaresTheKeyOnADifferentDate` |
| …and the legitimate re-declaration allowed | `testCreateIsAllowedWhenAnotherMissalDeclaresTheKeyOnTheSameDate` |
| a PATCH cannot move a shared key off the agreed date | `testPatchIsRefusedWhenItMovesASharedKeyOffTheAgreedDate` |
| DELETE collects every sidecar | `testDeleteGarbageCollectsTheKeyFromEveryI18nAndLectionarySidecar` |
| …but keeps readings another Missal still needs | `testDeleteRetainsRiteLevelReadingsForAKeyAnotherMissalStillDeclares` |
| queue mode: `submitted`, one batch, nothing on disk | `testCreatingASanctoraleEntryIsSubmittedAndTouchesNoFile` — 1 structure + 14 i18n + 6 readings rows in one batch |
| deleting an entry is not deleting the Missal | `testDeletingAnEntryIsNotFlaggedAsAResourceDeletion` |

## Verification

| command | result |
|--------------------------|--------------------------------------------------------------|
| `composer lint`          | exit 0 |
| `composer analyse`       | `[OK] No errors` (PHPStan level 10) |
| `composer parallel-lint` | 816 files, no syntax error |
| `composer lint:jsondata` | `OK — 23 file(s) match their canonical encoding` |
| `composer lint:openapi`  | `Your API description is valid.` |
| `composer test:quick`    | `Tests: 4361, Assertions: 188586, Warnings: 1, Skipped: 11` — 0 failures |

The single warning is `DiskSourceDataWriterTest`'s `mkdir(): File exists`; both files involved are untouched here and it is present on the branch point.

`openapi.json` was edited programmatically and round-trips byte-exactly through `json.dumps(indent=2, ensure_ascii=False)`; the diff is 354 lines of pure insertion with no reformatting. Every new operation declares the failures it can actually throw, `503` included.

`composer lint:missals` **cannot run on this branch**: `scripts/lint-missals.php` needs `JsonData::missalsFolderFor()`, which arrives with #947 through `development`. Its three invariants were checked with an equivalent standalone pass instead, which reports exactly two failures — `StIsidore`'s date disagreement and the Ambrosian `propriumdesanctis.json` naming — i.e. the two conditions #939/#940 fixed on `development` and which this branch's base predates. **No `jsondata/sourcedata` file is touched by this PR**, so neither is introduced here, and both disappear on merge.

## Branch base

This branch was cut from #948's head (`5cbf0053`), which predates the merges of #944 and #947. Two consequences, both resolved by merging `development`:

- `phpunit_tests/Support/ShadowProjectRootTrait.php` appears as an added file. It is **byte-identical** to the blob on `development` (`42ff43f3`), so the merge is trivial and it simply vanishes from the diff.
- A fast-forward onto `development` was attempted first and refused by tooling policy in this environment, so it is left for the merge.

Worth doing after this lands: refactor `scripts/lint-missals.php`'s invariant 2 onto `SanctoraleKeyIdentity`, so the corpus lint and the write guard share the one definition in code as well as in intent. It could not be done here because the script is not in this branch's base.

## Summary by CodeRabbit

* **New Features**
  * Added authenticated API support to create, update and delete individual Roman Missal sanctorale entries.
  * Synchronised changes across language data and readings for national and General Roman calendar resources.
  * Added queue-based submission support for changes requiring review.
* **Bug Fixes**
  * Prevented invalid edits, conflicting dates, event-key renaming and unsupported missal-level write requests.
  * Improved deletion handling while preserving shared readings still in use.
* **Documentation**
  * Published request schemas and OpenAPI documentation for the new operations.


JohnRDOrazio:
Both valid. Fixed in a4a18441.

## The failing test

`testAuthenticatedPutAtItemLevelReachesUnimplementedStub` asserted the response body contains
`"Not yet implemented"`. This PR **retired** that stub rather than filling it in, so the string is
correctly gone — the status is still 405, only the reason changed.

The reason it moved: #943's central requirement is *reject `event_key` mutation*, and an `event_key`
carried only in the request body makes a rename **inexpressible**, therefore unrefusable. The key
became a path segment, and writes moved to `/missals/{missal_id}/{event_key}`. `/missals/{missal_id}`
is consequently read-only, and a plain `405` there is now the right answer.

Replaced with two tests:

- `testAuthenticatedPutAtMissalLevelIsMethodNotAllowed` — 405, and asserts the body does **not**
  mention the stub, so a lingering mention would flag the old route surviving.
- `testAuthenticatedPutAtEntryLevelReachesTheWriteHandler` — asserts **not** 405 (the regression that
  would leave this PR shipped-but-dead: route unregistered or method disallowed), then 400 for the
  empty payload, since `{}` cannot pass schema validation.

## The docblock

Correct, and worth more than its "nitpick" label. `patterns()` still read:

> Order is defensive rather than load-bearing: no two templates here admit the same path…

This PR made that false. `{missal_folder}/{missal_folder}.json` widens to `missals/[^/]+/[^/]+\.json`,
which also matches `missals/propriumdetempore/propriumdetempore.json` — the overlap the temporale row
was moved above `MISSAL_FILE` to resolve. A maintainer trusting the old wording would have felt free to
sort the table and silently restore the bug, and the symptom would have been a wrong-red on every
temporale change request.

The docblock now states the dependency, says "do not sort this table", and scopes the remaining
defensive claim to the rows where it still holds.

I also checked the ordering is actually enforced rather than only documented — swapping the two rows:

```
1) SourceDataSchemaResolverTest::testEveryStageablePathFamilyResolvesToItsSchema@temporale
   Tests: 29, Failures: 1
```

and `OK (29 tests, 30 assertions)` once restored.

## Note on the two local failures

They are environmental, not the change. The responses carry `"file": "/var/www/html/src/..."` — a
container path — while this branch lives under `/home/johnrdorazio/...`, and a direct entry-level `PUT`
against that port answers `401` rather than `405`, i.e. routing this branch does not have. `:8000` is
serving a different revision. CI builds the server from the branch, so these pass there.

Worth recording: #922's preflight cannot catch this case, and said so when it landed — it proves "a
Liturgical Calendar API", not "the code under test". A stale container *of this project* passes it. This
is that limitation observed live, and an argument for the API exposing a build identifier.

`composer lint` clean · `composer analyse` `[OK] No errors` · `composer lint:jsondata` OK ·
`composer lint:openapi` valid · `composer lint:missals` OK (8 folders, 485 rows, 65 sidecars).

